PR #5, opened 2019-08-28
Make writeConnectionSecretToRef optional

negz:

Fixes https://github.com/crossplaneio/crossplane/issues/719

The resource claim reconciler (and API definitions) consider writeConnectionSecretToRef to be optional, but the managed resource reconciler fails if it is not specified. This change aligns the codebase on the reference being optional. Managed resources that do not specify a secret reference will provision successfully without publishing their connection details to a secret.

### Checklist

I have:

### DIFF
--- a/pkg/resource/api.go
+++ b/pkg/resource/api.go
@@ -88,7 +88,7 @@ func NewAPIManagedConnectionPropagator(c client.Client, t runtime.ObjectTyper) *
 
 // PropagateConnection details from the supplied resource to the supplied claim.
 func (a *APIManagedConnectionPropagator) PropagateConnection(ctx context.Context, cm Claim, mg Managed) error {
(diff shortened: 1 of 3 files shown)


negz:
> we also could just leave it as is, making our opinion on best practice more apparent to the user by requiring the no-op publisher be passed in if they do not desire to write connection secrets for a resource.

Whether or not to use the no-op publisher is a controller-level decision, so it only makes sense to use the no-op publisher if your resource never supports publishing connection secrets at all. This change relates more to the case in which a managed resource supports writing a connection secret, but the managed resource author omits `writeConnectionSecretRefTo`.

negz:
Ping @hasheddan @muvaf does this PR make sense given the context I've added to https://github.com/crossplaneio/crossplane-runtime/issues/9? If so, I'd appreciate a review so I can merge.

PR #12, opened 2019-09-05
Don't update managed resources if we determine they're up to date

### LINKED ISSUE #7: Determine whether resource update is necessary during observation.

Controllers using the the Generic Managed Resource Reconciler (GMRR) must make two cloud provider API calls during the most common reconcile path, when they only require one API call. 

The GMRR requires controller authors to satisfy the `resource.ExternalClient` interface, which looks like:

```go
type ExternalClient interface {
	Observe(ctx context.Context, mg Managed) (ExternalObservation, error)
	Create(ctx context.Context, mg Managed) (ExternalCreation, error)
	Update(ctx context.Context, mg Managed) (ExternalUpdate, error)
	Delete(ctx context.Context, mg Managed) error
}
```

The GMRR first calls `Observe` to report whether the external resource exists, and if so sync the managed resource's status. It then either:

* Calls `Delete` if the external resource exists but the managed resource was deleted.
* Calls `Create` if the external resource does not exist.
* Calls  `Update` if neither of the above cases apply.

Currently `Update` is called regardless of whether the external resource is out of sync with the managed resource. The `Update` implementation must either perform a no-op update of the external resource (typically at the expense of an API call), or get a fresh view of the external resource in order to determine whether it requires updating.

The vast majority of reconcile invocations over a managed resource's lifecycle will hit the `Update` branch, and not require an update. This will typically involve one of the two following cloud provider API call patterns.

If the `Update` implementation chooses to perform no-op updates:
1. Get the external resource from the cloud provider API during `Observe`.
1. Update the external resource in the cloud provider API during `Update`.

If the `Update` implementation chooses to determine whether an update is required:
1. Get the external resource from the cloud provider API during `Observe`.
1. Get the external resource from the cloud provider API during `Update`.

In either case this is one more API call than is necessary; the `Observe` method has all the information necessary (i.e. a copy of the managed resource and a copy of the external resource) to determine whether the external resource must be updated.

The GMRR could make `Observe` rather than `Update` responsible for determining whether an update was necessary by adding a `ResourceInSync` field to `resource.ExternalObservation`:

```go
type ExternalObservation struct {
	ResourceExists    bool
	ResourceInSync    bool
	ConnectionDetails ConnectionDetails
}
```

This would allow the GRMM logic to return early if no create, update, or delete was necessary, having made only one cloud provider API call during `Observe`.

negz:

This allows us to determine whether an external resource needs updating inside the Observe method, which must already get a fresh copy of the external resource from the cloud provider API in order to determine whether it exists and update the managed resource's status.

Note that I made the field `ResourceUpToDate` rather than `ResourceNeedsUpdate`, so `Update` will be called by default unless `Observe` determines it's not needed. This maintains compatibility with existing `ExternalClient` implementations, and generally seems like the safer default.

Fixes #7 

### Checklist

I have:

### DIFF
--- a/pkg/resource/managed_reconciler.go
+++ b/pkg/resource/managed_reconciler.go
@@ -114,16 +114,17 @@ type ExternalClient interface {
 	Observe(ctx context.Context, mg Managed) (ExternalObservation, error)
 
 	// Create an external resource per the specifications of the supplied
-	// Managed resource.
+	// Managed resource. Called when Observe reports that the associated
+	// external resource does not exist.
 	Create(ctx context.Context, mg Managed) (ExternalCreation, error)
 
 	// Update the external resource represented by the supplied Managed
-	// resource, if necessary. Update implementations must handle the case in
-	// which no update is necessary.
+	// resource, if necessary. Called unless Observe reports that the
+	// associated external resource is up to date.
 	Update(ctx context.Context, mg Managed) (ExternalUpdate, error)
 
 	// Delete the external resource upon deletion of its associated Managed
-	// resource.
+	// resource. Called when the managed resource has been deleted.
 	Delete(ctx context.Context, mg Managed) error
 }
 
@@ -186,6 +187,7 @@ func (c *NopClient) Delete(ctx context.Context, mg Managed) error { return nil }
 // resource.
 type ExternalObservation struct {
 	ResourceExists    bool
+	ResourceUpToDate  bool
 	ConnectionDetails ConnectionDetails
 }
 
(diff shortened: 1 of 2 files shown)


muvaf:
At the time we call Update, we know that the status is in sync with provider because Observe has already been called. I think we can have an interface with one function _IsSameAs_ that Managed resources implement. It will do the check solely between the values of fields on the CR, ideally compare status with spec. If it returns true, we’d just not call Update. That requires to have status with full information that we can compare against spec though.

PR #18, opened 2019-09-13
Allow both resource claims and managed resources to pass the class kind predicate

negz:

It turns out both our watches use the same predicate and apply it to the watched kind, not the kind that actually gets enqueued.

### Checklist

I have:


negz:
I've also run `make tests`

negz:
Confirmed this works by building stack-gcp against this commit and creating a `RedisCluster`, `RedisClusterClass`, and `CloudSQLInstanceClass`. The binding happens as soon as the managed resource becomes available, as expected.

jbw976:
Thank you for finding this issue and getting it fixed so quickly @negz!! 🙇

PR #24, opened 2019-09-14
Fix support for static provisioning

negz:

This PR fixes support for managed resources that do not reference a class. Controllers built against crossplane-runtime will currently panic if they encounter a managed resource without a class reference. Any dynamically provisioned managed resource will have a class reference, but this breaks the static provisioning workflow in which a managed resource is explicitly provisioned, then later claimed.

The PR also fixes watch predicates for the static binding flow We currently support dynamic provisioning in the resource claim reconciler by using a watch predicate that allows either managed resources that directly reference a non-portable resource class of a given kind, or resource claims that reference a non-portable resource class of a given kind indirectly via a portable resource class.

To support static provisioning (i.e. explicitly claiming an existing managed resource) we must also allow resource claims that explicitly reference a managed resource. Writing one predicate to do all of this was getting cumbersome, so I have refactored the predicate interface a little.

This PR will require additional PRs to update Crossplane core, and each stack.

### Checklist

I have:

### DIFF
--- a/pkg/resource/predicates.go
+++ b/pkg/resource/predicates.go
@@ -67,6 +67,10 @@ func HasDirectClassReferenceKind(k NonPortableClassKind) PredicateFn {
 			return false
 		}
 
+		if r.GetNonPortableClassReference() == nil {
+			return false
+		}
+
 		return r.GetNonPortableClassReference().GroupVersionKind() == schema.GroupVersionKind(k)
 	}
 }
--- a/pkg/resource/predicates.go
+++ b/pkg/resource/predicates.go
@@ -126,9 +126,9 @@ func HasIndirectClassReferenceKind(c client.Client, oc runtime.ObjectCreater, k 
 	}
 }
 
-// NoPortableClassReference accepts ResourceClaims that do not reference a
+// HasNoPortableClassReference accepts ResourceClaims that do not reference a
 // specific portable class
-func NoPortableClassReference() PredicateFn {
+func HasNoPortableClassReference() PredicateFn {
 	return func(obj runtime.Object) bool {
 		cr, ok := obj.(PortableClassReferencer)
 		if !ok {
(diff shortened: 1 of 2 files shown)


negz:
I've tested that both the static and dynamic provisioning flows work by using a build of https://github.com/crossplaneio/stack-gcp/pull/24 to statically and then dynamically provision a `CloudMemorystoreInstance` and bind it to a `RedisCluster`.

pkg/resource/predicates.go
```diff
@@ -41,20 +41,33 @@ func NewPredicates(fn PredicateFn) predicate.Funcs {
...
 
-// TODO(negz): Use separate predicates for resource claims and managed
-// resources once controller-runtime supports this.
-// https://github.com/kubernetes-sigs/controller-runtime/issues/572
-
-// HasClassReferenceKinds accepts objects that reference the supplied
-// non-portable class kind, either directly or indirectly via a portable class
-// kind.
-func HasClassReferenceKinds(c client.Client, oc runtime.ObjectCreater, k ClassKinds) PredicateFn {
+// AnyOf accepts objects that pass any of the supplied the supplied predicate
```
hasheddan: This is really cool! Wonder if any upstream k8s projects (likely controller-runtime) would want to add this as part of the `NewControllerManagedBy` chain.

jbw976: nit: stutter?
> the supplied the supplied predicate

pkg/resource/predicates.go
```diff
@@ -41,20 +41,33 @@ func NewPredicates(fn PredicateFn) predicate.Funcs {
...
+		if !ok {
+			return false
 		}
 
-		return HasIndirectClassReferenceKind(c, oc, k)(obj)
+		if r.GetResourceReference() == nil {
+			return false
+		}
+
+		return r.GetResourceReference().GroupVersionKind() == schema.GroupVersionKind(k)
```
jbw976: i was wondering about this the other day, do we need to do any `ToLower()` normalization or `EqualFold()` comparison in these predicates?

negz: The type metadata supplied to the predicate will always match the actual case of our API types (e.g. `RDSInstanceClass`, not `rdsinstanceclass`), so the question is whether we want to require our users to match the actual case. I'm ambivalent about that so I'll avoid changing it for the time being. Everything will work fine as long as the correct case is used. It seems typical to do this in Kubernetes, i.e. I usually see `kind: Deployment`, not `kind: deployment`.

pkg/resource/predicates.go
negz: I believe they'd need to also be interested in taking on the `NewPredicates` pattern for it to work. I'm on the fence as to whether it's worth contributing; it's a fairly simple pattern.

PR #36, opened 2019-09-27
portable class: embed all necessary fields

### LINKED ISSUE #26: Make portable classes subtypes of runtimev1alpha1.PortableClass

Currently every portable class is a distinct type that embeds `runtimev1alpha1.PortableClass`, with the exact same shape. For example:

```go
type PortableClass struct {
	NonPortableClassReference *corev1.ObjectReference `json:"classRef,omitempty"`
}

type RedisClusterClass struct {
	metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty"`

	runtimev1alpha1.PortableClass `json:",inline"`
}
```

We could probably write a little less code by making them a type (in the Go sense) of PortableClass, e.g.:

```go
type PortableClass struct {
	metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty"`

	NonPortableClassReference *corev1.ObjectReference `json:"classRef,omitempty"`
}

type RedisClusterClass PortableClass
```

hasheddan:

Embeds all subfields in `PortableClass`. Crossplane will need to bump `crossplane-runtime` dependency version and remove subfields from the portable class types to complete change.

Fixes #26 

### Checklist

I have:


hasheddan:
@negz this has not been tested, but I will do so now

hasheddan:
@negz setting the underlying type of portable classes to `PortableClass` does not allow it to satisfy the `PortableClass` interface because types do not inherit the methods of their underlying types. However, we can still make the `PortableClass` type the only thing that is embedded in individual portable class types. I have implemented here for your reference: https://github.com/hasheddan/crossplane/commit/b05fd346b17045e8bd47183044108c4566d1e61f

negz:
@hasheddan Thanks for trying that out! It's possible that using a [type alias](https://go.googlesource.com/proposal/+/master/design/18130-type-alias.md) would work, but I think that would be an abuse of that feature.

On reflection I'm slightly inclined to abandon this effort. When I see that a struct embeds `TypeMeta` and `ObjectMeta` it's immediately obvious to me that it intends to be a Kubernetes resource. It didn't occur to me that this might make the code a little less obvious when I wrote #26, but seeing how the actual code reads in the context of https://github.com/hasheddan/crossplane/commit/b05fd346b17045e8bd47183044108c4566d1e61f makes me think the current pattern is better than the slight optimisation of not having to include type and object metadata. What do you think?

hasheddan:
Agreed! Was worth a shot :)

PR #42, opened 2019-10-08
Constantly propagate connection secrets

negz:

Fixes https://github.com/crossplaneio/crossplane-runtime/issues/35.

This commit allows constant propagation of connection secrets from managed resources to their bound resource claims. It does this by updating the existing APIManagedConnectionPropagator to add 'propagation annotations', which can be used by a new 'secret propagating reconciler' to watch both secrets for constant propagation. The predicates and enqueue handler required to implement this are included.

I've tested that this works by adding a secret propagation controller using this reconciler to the CloudMemorystoreInstance controller (because I had a GCP testing setup at hand). I've confirmed that:

1. The propagation annotations look as expected.
1. Fields added to the managed resource secret are instantly propagated to the claim secret.
1. Fields removed from the managed resource secret are instantly removed from the claim secret.
1. Fields changed in the managed resource secret are instantly propagated to the claim secret.
1. Fields added to the claim secret (that are not in the managed resource secret) are instantly removed.

I still need to:

- [x] Implement the AllOf predicate, which I only realised I needed while writing this up.
- [x] Confirm it works in practice (probably using EKS).
- [ ] Document the existence of this controller in the Services Developer Guide

Resources that never update their connection secret after creation time can omit this controller. Resources (like EKS) whose connection secret could change over time would instantiate it as follows:

```go
type EKSClusterSecretController struct{}

func (c *EKSClusterSecretController) SetupWithManager(mgr ctrl.Manager) error {
	p := resource.NewPredicates(resource.AnyOf(
		resource.AllOf(resource.IsControlledByKind(v1alpha1.KubernetesClusterKind), resource.IsPropagated()),
		resource.AllOf(resource.IsControlledByKind(v1alpha2.EKSClusterKind), resource.IsPropagator()),
	))

	return ctrl.NewControllerManagedBy(mgr).
		Named(strings.ToLower(fmt.Sprintf("connectionsecret.%s.%s", v1alpha2.EKSClusterKind, v1alpha2.Group))).
		Watches(&source.Kind{Type: &corev1.Secret{}}, &resource.EnqueueRequestForPropagator{}).
		For(&v1.Secret{}).
		WithEventFilter(p).
		Complete(resource.NewSecretPropagatingReconciler(mgr))
}


```

### Checklist

I have:

### DIFF
--- a/pkg/resource/api.go
+++ b/pkg/resource/api.go
@@ -37,6 +37,7 @@ const (
 	errUpdateClaim          = "cannot update resource claim"
 	errGetSecret            = "cannot get managed resource's connection secret"
 	errSecretConflict       = "cannot establish control of existing connection secret"
+	errUpdateSecret         = "cannot update connection secret"
 	errCreateOrUpdateSecret = "cannot create or update connection secret"
 	errUpdateManaged        = "cannot update managed resource"
 	errUpdateManagedStatus  = "cannot update managed resource status"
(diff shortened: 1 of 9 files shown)


negz:
> The changeset is a bit long, so it's hard to review it in a ton of detail

I'll shamefully admit that I had this all broken down into bite sized commits to aid in review, but then I made enough changes in the last commit that I got lazy and squashed them all. :(

negz:
> This will also be a corev1.Secret, right?

Right - copy and paste error.

I've updated the description to describe how I've tested this. I'm going to merge and then update the docs, given they live in a separate repo.

PR #43, opened 2019-10-08
namespace-name scheme for managed resources

muvaf:

This PR changes the existing ObjectMetaConfigurator so that it will set the `GenerateName` field of the managed resource to `<claim ns>-<claim name>-`, which will have api-server name the managed resource as `<claim ns>-<claim name>-<5 char rand string>`. Addresses the new naming scheme in https://github.com/crossplaneio/crossplane/blob/c8ae09bc1c3befbc7ec7b96a37c3f729975d5b92/design/one-pager-managed-resource-api-design.md#custom-resource-instance-name

All claim reconcilers who use this configurator will be creating resource with the new naming scheme but existing CR instances shouldn't be affected.

Tested manually. Links will be updated once one-pager is merged.

### Checklist

I have:

### DIFF
--- a/pkg/resource/api.go
+++ b/pkg/resource/api.go
@@ -60,16 +60,18 @@ func NewAPIManagedCreator(c client.Client, t runtime.ObjectTyper) *APIManagedCre
 func (a *APIManagedCreator) Create(ctx context.Context, cm Claim, cs NonPortableClass, mg Managed) error {
 	cmr := meta.ReferenceTo(cm, MustGetKind(cm, a.typer))
 	csr := meta.ReferenceTo(cs, MustGetKind(cs, a.typer))
-	mgr := meta.ReferenceTo(mg, MustGetKind(mg, a.typer))
 
 	mg.SetClaimReference(cmr)
 	mg.SetNonPortableClassReference(csr)
 	if err := a.client.Create(ctx, mg); err != nil {
 		return errors.Wrap(err, errCreateManaged)
 	}
-
-	meta.AddFinalizer(cm, claimFinalizerName)
+	// Since we use GenerateName feature of ObjectMeta, final name of the
+	// resource is calculated during the creation of the resource. So, we
+	// can generate a complete reference only after the creation.
+	mgr := meta.ReferenceTo(mg, MustGetKind(mg, a.typer))
 	cm.SetResourceReference(mgr)
+	meta.AddFinalizer(cm, claimFinalizerName)
 
 	return errors.Wrap(a.client.Update(ctx, cm), errUpdateClaim)
 }
--- a/pkg/resource/configurator.go
+++ b/pkg/resource/configurator.go
@@ -19,8 +19,6 @@ package resource
 import (
 	"context"
 	"fmt"
-	"reflect"
-	"strings"
 
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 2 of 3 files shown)


pkg/resource/configurator.go
```diff
@@ -75,3 +75,28 @@ func kindish(obj runtime.Object) string {
 	}
 	return strings.ToLower(reflect.TypeOf(obj).Elem().Name())
 }
+
+type ObjectMetaConfiguratorWithNewNamingFormat struct {
```
muvaf: I can definitely use some help with the name of that struct..

pkg/resource/configurator.go
```diff
@@ -75,3 +75,28 @@ func kindish(obj runtime.Object) string {
...
+
+// NewObjectMetaConfigurator returns a new ObjectMetaConfigurator.
+func NewObjectMetaConfiguratorWithNewNamingFormat(t runtime.ObjectTyper) *ObjectMetaConfiguratorWithNewNamingFormat {
+	return &ObjectMetaConfiguratorWithNewNamingFormat{typer: t}
+}
+
+// Configure the supplied Managed resource's object metadata.
+func (c *ObjectMetaConfiguratorWithNewNamingFormat) Configure(_ context.Context, cm Claim, cs NonPortableClass, mg Managed) error {
+	mg.SetNamespace(cs.GetNamespace())
+	if mg.GetName() == "" {
```
negz: Do we need this check? I recall that configurators are applied in order, so if a resource claim implementation wanted to use this configurator but also specify a bespoke name it could do so by applying this configurator first, then overriding its behaviour in a subsequent configurator.

pkg/resource/configurator.go
muvaf: I say we just change the existing configurator then. In both cases, the default would be the new scheme and people who want to use the old one can supply their own configurator.

pkg/resource/configurator.go
muvaf: I was under impression that already existing resource might come through this configurator and we won't want to set `GenerateName` for an existing resource, but I see that's not the case. It's called only if the resource doesn't exist. Removed the check.

pkg/resource/api.go
```diff
@@ -60,16 +60,17 @@ func NewAPIManagedCreator(c client.Client, t runtime.ObjectTyper) *APIManagedCre
...
-	mgr := meta.ReferenceTo(mg, MustGetKind(mg, a.typer))
 
 	mg.SetClaimReference(cmr)
 	mg.SetNonPortableClassReference(csr)
 	if err := a.client.Create(ctx, mg); err != nil {
 		return errors.Wrap(err, errCreateManaged)
 	}
 
-	meta.AddFinalizer(cm, claimFinalizerName)
+	// Final name/namespace values of the managed resource are determined _after_ creation.
```
negz: It might be helpful to explain why this is the case (because we now use generatename, I assume). It wasn't immediately obvious to me.

pkg/resource/configurator.go
```diff
@@ -56,7 +54,7 @@ func NewObjectMetaConfigurator(t runtime.ObjectTyper) *ObjectMetaConfigurator {
 // Configure the supplied Managed resource's object metadata.
 func (c *ObjectMetaConfigurator) Configure(_ context.Context, cm Claim, cs NonPortableClass, mg Managed) error {
```
negz: Should this configurator set the managed resource's external name to the value of the claim's `ExternalNameAnnotationKey`, if it exists at dynamic provisioning time?

muvaf: Another PR is coming up for use of `ExternalNameAnnotationKey`. Actually, I'll go ahead and remove that constant from this PR.

But the idea is that if get an annotation on the claim, we propagate it down to managed resource's annotation, then managed reconciler and/or external client will make use of that annotation. So, the name of the managed resource is not affected by this annotation on the claim.

negz: Ah I see. I agree that makes more sense than setting the actual name of the managed resource, given the annotation is about external names.

PR #44, opened 2019-10-08
Add UnavailableWithMessage condition helper

turkenh:
Signed-off-by: Hasan Turken <turkenh@gmail.com>

Adds a helper function getting a Unavailable status condition with a message.
To be initially consumed by https://github.com/crossplaneio/stack-gcp/pull/44

### Checklist

I have:

### DIFF
--- a/apis/core/v1alpha1/condition.go
+++ b/apis/core/v1alpha1/condition.go
@@ -207,6 +207,21 @@ func Unavailable() Condition {
 	}
 }
 
+// UnavailableWithMessage returns a condition that indicates the managed
+// resource is not currently available for use and contains the message returned
+// by the API. Unavailable should be set only when Crossplane expects the managed
+// resource to be available but knows it is not, for example because its API
+// reports it is unhealthy.
+func UnavailableWithMessage(msg string) Condition {
+	return Condition{
+		Type:               TypeReady,
+		Status:             corev1.ConditionFalse,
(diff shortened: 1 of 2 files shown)


apis/core/v1alpha1/condition.go
```diff
@@ -207,6 +207,21 @@ func Unavailable() Condition {
 	}
 }
 
+// UnavailableWithMessage returns a condition that indicates the managed
+// resource is not currently available for use and contains the message returned
+// by the API. Unavailable should be set only when Crossplane expects the managed
+// resource to be available but knows it is not, for example because its API
+// reports it is unhealthy.
+func UnavailableWithMessage(msg string) Condition {
```
negz:
```go
func (c Condition) WithMessage(msg string) Condition {
    return Condition{
	Type:               c.Type,
	Status:             c.Status,
	LastTransitionTime: c.LastTransitionTime,
	Reason:             c.Reason,
	Message: 	    msg,
    }
}
```

What do you think about implementing this same functionality by adding something like the above to our existing `Condition` type? This would allow any current Condition to be reported with a message, for example:

```go
r.SetConditions(runtimev1alpha1.Unavailable().WithMessage("resource is in maintenance"))
```

If we did this, a `WithError` variant may also make sense for convenience.

turkenh: Yes, this sounds much better. Done!

Issue #27, opened 2019-08-29
Crossplane should refuse to operate on existing external resources


negz:
### What problem are you facing?

As more and more Crossplane managed resources support updating (in addition to creating and deleting) their external resources, we're likely to run into cases where Crossplane will "implicitly adopt" and start managing existing external resources. This will become more prevalent as more Crossplane managed resources allow configuration of their underlying external resource names per https://github.com/crossplaneio/crossplane/issues/624.

For example, assume a critical 1TB CloudSQL database named `prod` exists in a particular GCP account. Further assume Crossplane is configured to use this GCP account, and a resource claim for a 10GB CloudSQL database with external name `prod` was created. Today there is no logic that would prevent Crossplane considering the existing database to be under Crossplane management, and trying to resize it down to 10GB.

### How could Crossplane help solve your problem?

Crossplane will need to be able to distinguish external resources it created from existing external resources. This may be difficult, given that not all cloud providers and external resources support storing metadata.

Relates to crossplaneio/crossplane-runtime#22.

negz:
I'm less convinced we need to be concerned about this - at least until we get some data from users that it's a problem. For now I'm going to close this issue.

ugurcancaykara:
Hi, can I get an update about this ? 
Use case scenario:
- Provisioned two DynamoDB table (Table resource) in two aws region
- Provisioned a DynamoDB GlobalTable for them
I was planning to configure autoscaling when it comes PROVISIONED billing mode however it's not currently supported.
Since crossplane recovers the changes I make manually that's not an option but a solution like this would be helpful. I would provision these resources as code and put the update: Observe policy to manifest file and I could handle the rest manually.

metrictwo:
+1 for an update, please. This feature is table-stakes given the risk of accidentally modifying existing production resources.

jeanduplessis:
@ugurcancaykara @metrictwo would either of these solutions work for you:
https://github.com/crossplane/crossplane/pull/3531/files
https://github.com/crossplane/crossplane/pull/3822/files

metrictwo:
@jeanduplessis Thanks for the response. Neither of those quite work for our use case. What we'd ideally like is a way to configure crossplane such that when a crossplane resource is created that represents an existing infra resource, it will import fields, but refuse to reconcile the resource altogether until/unless it's been indicated by some field or annotation that it's okay to do so. We want crossplane to be free to manage resources that it created, though.

bobh66:
[...]

PR #45, opened 2019-10-09
Use managed resource name as external name 

muvaf:

This PR introduces `Initializer` hook to generic managed reconciler. Existing `Establisher`s are refactored to be `Initalizer`s. 

`Initialize` is called before any `ExternalClient` calls to make sure everything is ready for queries.

The first use is calculating the external name for the managed resource. The default initializer that all managed reconciler will use is now `ManagedNameAsExternalName`, which means, if not already given, `crossplane.io/external-name` annotation will be populated by the name of the managed resource.

Additionally;
- During `Bind` operation, we propagate back the external name to the claim's annotation.
- During managed resource creation we propagate down the external name annotation from claim to the managed resource.

Fixes https://github.com/crossplaneio/crossplane/issues/624

Implements https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-managed-resource-api-design.md#external-resource-name
### Checklist

I have:

### DIFF
--- a/pkg/meta/meta_test.go
+++ b/pkg/meta/meta_test.go
@@ -19,6 +19,8 @@ package meta
 import (
 	"testing"
 
+	"github.com/crossplaneio/crossplane-runtime/apis/core/v1alpha1"
+
 	"github.com/google/go-cmp/cmp"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
@@ -750,6 +752,58 @@ func TestWasCreated(t *testing.T) {
 		t.Run(name, func(t *testing.T) {
 			got := WasCreated(tc.o)
 			if diff := cmp.Diff(tc.want, got); diff != "" {
+				t.Errorf("WasCreated(...): -want, +got:\n%s", diff)
+			}
+		})
+	}
+}
+
+func TestGetExternalName(t *testing.T) {
+	testName := "my-external-name"
+
+	cases := map[string]struct {
+		o    metav1.Object
+		want string
+	}{
+		"ExternalNameExists": {
+			o:    &corev1.Pod{ObjectMeta: metav1.ObjectMeta{Annotations: map[string]string{v1alpha1.ExternalNameAnnotationKey: testName}}},
+			want: testName,
+		},
+		"NoExternalName": {
+			o:    &corev1.Pod{},
+			want: "",
+		},
+	}
+
+	for name, tc := range cases {
+		t.Run(name, func(t *testing.T) {
+			got := GetExternalName(tc.o)
+			if diff := cmp.Diff(tc.want, got); diff != "" {
+				t.Errorf("WasCreated(...): -want, +got:\n%s", diff)
+			}
+		})
+	}
+}
+
+func TestSetExternalName(t *testing.T) {
+	testName := "my-external-name"
+
+	cases := map[string]struct {
+		o    metav1.Object
+		name string
+		want metav1.Object
+	}{
+		"SetsTheCorrectKey": {
+			o:    &corev1.Pod{},
+			name: testName,
+			want: &corev1.Pod{ObjectMeta: metav1.ObjectMeta{Annotations: map[string]string{v1alpha1.ExternalNameAnnotationKey: testName}}},
+		},
+	}
+
+	for name, tc := range cases {
+		t.Run(name, func(t *testing.T) {
+			SetExternalName(tc.o, tc.name)
+			if diff := cmp.Diff(tc.want, tc.o); diff != "" {
 				t.Errorf("WasCreated(...): -want, +got:\n%s", diff)
 			}
 		})
--- a/pkg/meta/meta_test.go
+++ b/pkg/meta/meta_test.go
@@ -506,6 +506,59 @@ func TestRemoveFinalizer(t *testing.T) {
 	}
 }
 
+func TestFinalizerExists(t *testing.T) {
+	finalizer := "fin"
+	funalizer := "fun"
+
+	type args struct {
+		o         metav1.Object
+		finalizer string
+	}
+
+	cases := map[string]struct {
+		args args
+		want bool
+	}{
+		"NoExistingFinalizers": {
+			args: args{
+				o:         &corev1.Pod{},
+				finalizer: finalizer,
+			},
+			want: false,
+		},
+		"FinalizerExists": {
+			args: args{
+				o: &corev1.Pod{
+					ObjectMeta: metav1.ObjectMeta{
+						Finalizers: []string{finalizer},
+					},
+				},
+				finalizer: finalizer,
+			},
+			want: true,
+		},
+		"AnotherFinalizerExists": {
+			args: args{
+				o: &corev1.Pod{
+					ObjectMeta: metav1.ObjectMeta{
+						Finalizers: []string{funalizer},
+					},
+				},
+				finalizer: finalizer,
+			},
+			want: false,
+		},
+	}
+
+	for name, tc := range cases {
+		t.Run(name, func(t *testing.T) {
+			if diff := cmp.Diff(tc.want, FinalizerExists(tc.args.o, tc.args.finalizer)); diff != "" {
+				t.Errorf("tc.args.o.GetFinalizers(...): -want, +got:\n%s", diff)
+			}
+		})
+	}
+}
+
 func TestAddLabels(t *testing.T) {
 	key, value := "key", "value"
 	existingKey, existingValue := "ekey", "evalue"
--- a/pkg/meta/meta_test.go
+++ b/pkg/meta/meta_test.go
@@ -19,8 +19,6 @@ package meta
 import (
 	"testing"
 
-	"github.com/crossplaneio/crossplane-runtime/apis/core/v1alpha1"
-
 	"github.com/google/go-cmp/cmp"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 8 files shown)


muvaf:
I am thinking of adding `ExternalNameGetter` interface to `Managed` interface so that we can have the following in all managed resources:
```
func (m *ManagedType) GetExternalName() string {
    if m.Annotations == nil {
        return ""
    }
    return m.Annotations[v1alpha1.ExternalNameAnnotationKey]
}
```
Instead of using `cr.Annotations[v1alpha1.ExternalNameAnnotationKey]` [everywhere](https://github.com/crossplaneio/stack-gcp/pull/46/files#diff-086576e912db107bfde4c2d36c67477cR124), we can just call that function.

negz:
>I am thinking of adding ExternalNameGetter interface to Managed interface.

The `Managed` interface is already pretty large - I'd prefer not to let it grow too much more unless it really needs to. If we really need such functionality I think it would be better to add it as a function in `crossplane-runtime/pkg/meta`. e.g.:

```go
type Annotated interface {
    GetAnnotations() map[string]string
}

func ExternalName(a Annotated) string {
    # No need to check for nil - a zero value map is safe for reads, and will return its zero value, i.e. ""
    return a.GetAnnotations()[v1alpha1.ExternalNameAnnotationKey]
}

File: pkg/resource/api.go [hunk 1]
```diff
@@ -154,6 +154,10 @@ func NewAPIManagedBinder(c client.Client) *APIManagedBinder {
 // Bind the supplied resource to the supplied claim.
 func (a *APIManagedBinder) Bind(ctx context.Context, cm Claim, mg Managed) error {
 	cm.SetBindingPhase(v1alpha1.BindingPhaseBound)
+	// Propagate back the final name of the external resource to the claim.
+	if meta.GetExternalName(mg) != "" {
```
negz: How likely is it that the external name annotation will be unset at binding time? Should we ensure the managed resource does not become bindable until its external name is set?

pkg/resource/api.go
```diff
@@ -154,6 +154,10 @@ func NewAPIManagedBinder(c client.Client) *APIManagedBinder {
...
+		meta.SetExternalName(cm, meta.GetExternalName(mg))
```
negz: I believe setting the claim's external name will require us to call `client.Update(cm)`. It looks like we currently only call `client.Status().Update(cm)`. My preference would be to call `client.Update` from inside the `APIManagedBinder` and `APIManagedStatusBinder`.

pkg/resource/managed_reconciler.go
```diff
@@ -224,13 +238,15 @@ type mrManaged struct {
...
 	ManagedFinalizer
+	ManagedInitializer
 }
 
 func defaultMRManaged(m manager.Manager) mrManaged {
 	return mrManaged{
 		ManagedConnectionPublisher: NewAPISecretPublisher(m.GetClient(), m.GetScheme()),
 		ManagedEstablisher:         NewAPIManagedFinalizerAdder(m.GetClient()),
 		ManagedFinalizer:           NewAPIManagedFinalizerRemover(m.GetClient()),
+		ManagedInitializer:         NewManagedNameAsExternalName(m.GetClient()),
```
negz: Could we merge the initializer and establisher interfaces into one? I prefer the name Initializer. We could either support chaining establishers, or just have one establisher that sets the external name annotation and adds a finalizer in order to reduce the amount of `Update` calls we must make.

pkg/resource/managed_reconciler.go
```diff
@@ -341,6 +365,14 @@ func (r *ManagedReconciler) Reconcile(req reconcile.Request) (reconcile.Result,
 		return reconcile.Result{RequeueAfter: r.shortWait}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
 	}
 
+	if err := r.managed.Initialize(ctx, managed); err != nil {
```
hasheddan: Trying to work through the case where someone modifies the name of the external resource directly in the provider UI or something after a claim is already bound to a managed resource. If we only propagate the external name to the claim at "bind time" does that mean that the external name will not be propagated after the first time it is set?

pkg/resource/managed_reconciler.go
muvaf: That was my first thought as well - move `Establish` up before the first `Observe` call and implement a new establisher.

At the time, I believe we put the `Establish` call in `!observation.ResourceExists` because we didn't want to set the finalizer of a resource that we didn't create in the first place and we have to do the naming assignment before the first `Observe` call. That's why I didn't go with that idea.

However, I can see that our thinking around managing existing resources started to change (see https://github.com/crossplaneio/crossplane-runtime/issues/27). Seems to me, our direction for existing resource will not have much to do with the finalizers after all. So, I'm happy to merge them into one.

Naming still has to be done before the `Observe` call, so, I'll refactor the existing establisher to be an initializer and remove the `ManagedEstablisher` completely.

pkg/resource/managed_reconciler.go
muvaf: The definition of the external name is basically what provider uses as identifier. TBH, I can't see a scenario where the identifier can be changed unless you delete and re-create the resource.

pkg/resource/api.go
muvaf: Yes, I realized that it didn't propagate back while testing new CloudSQL controller. Thanks for spotting!

pkg/resource/api.go
muvaf: > How likely is it that the external name annotation will be unset at binding time?

Only if developer of a resource, which acquires its name from the provider, forgets to add it as annotation & does not use the annotation in the provider queries. Or there is a bug and external name is set and then unset for some reason. I don't know, not very likely given that in all queries they'd depend on this annotation.

> Should we ensure the managed resource does not become bindable until its external name is set?

It's merely informational on `Claim` level after its creation, so, not sure if we want it to block the binding process. But one can also say, we'd have everyone onboard to use it if we don't bind it otherwise. Although, this would break all controllers who do not use the annotation until they do.

pkg/resource/managed_reconciler.go
```diff
@@ -222,15 +222,18 @@ type ManagedReconciler struct {
...
 	ManagedFinalizer
+	ManagedInitializer
 }
 
 func defaultMRManaged(m manager.Manager) mrManaged {
 	return mrManaged{
 		ManagedConnectionPublisher: NewAPISecretPublisher(m.GetClient(), m.GetScheme()),
-		ManagedEstablisher:         NewAPIManagedFinalizerAdder(m.GetClient()),
 		ManagedFinalizer:           NewAPIManagedFinalizerRemover(m.GetClient()),
+		ManagedInitializer: InitializerChain{
```
muvaf: The developers who work on the resources that acquires their name after the creation will have to provide their own `Initializer`s. With this pattern, instead of supplying an empty chain, they should supply `NewAPIManagedFinalizerAdder(m.GetClient())`. Not sure how we'd enforce that.

In fact, I started to think why allow people not to use `NewAPIManagedFinalizerAdder(m.GetClient())`? It could be a part of managed reconciler and all resource would be processed by it. I couldn't come up with a case where we don't want to set the finalizer of the resource, given that external policies(or something possibly other than finalizers) will be how we handle the management of the resource #27

hasheddan: > Not sure how we'd enforce that.

We could just inject it in the `InitializerChain` within the `WithManagedInitializer()` func. However, that would effectively render it impossible not to use `NewAPIManagedFinalizerAdder(m.GetClient())`.

muvaf: Then if they did use `NewAPIManagedFinalizerAdder(m.GetClient())` in `WithManagedInitializers` call, there would be duplicate of it.

But if we inject it and don't export `APIManagedFinalizerAdder` and `APIManagedFinalizerRemover` and friends at all, they'd not be able to add it. I think I'll go with that.

hasheddan: Yeah if we are going use it no matter what then exposing it to the user is a bad UX.

muvaf: Well, `WithManagedInitializers` doesn't have a way to access the client that's needed to call `NewAPIManagedFinalizerAdder`. I'd like to get more feedback on whether it's worth it to enforce `NewAPIManagedFinalizerAdder` even though developer forgets it for the resources that won't use `NewManagedNameAsExternalName`(not too few) before making not-so-small/hacky changes to make it work. @negz

pkg/resource/api.go
negz: > Although, this would break all controllers who do not use the annotation until they do.

That's a good point - we have plenty of controllers that don't yet use the managed resource reconciler (and thus don't honor this external name annotation) and probably won't for some time. Given that all resources _do_ use this resource claim reconciler I agree we shouldn't assume the annotation is present.

pkg/meta/meta_test.go
```diff
@@ -506,6 +506,59 @@ func TestRemoveFinalizer(t *testing.T) {
 	}
 }
 
+func TestFinalizerExists(t *testing.T) {
+	finalizer := "fin"
+	funalizer := "fun"
```
negz: I approve of this test value. 👍

pkg/resource/managed_reconciler.go
negz: Hrm this is a tough one. In practice pretty much every managed resource is going to want to add a finalizer.

I'm inclined to keep this as you currently have it implemented. If your controller falls into the most common case of using its managed resource name you can stick with the default. If you need to override the default behaviour then you need to feed in the FinalizerAdder as well as your external name strategy.

muvaf:
@negz I think this is ready for merge. Just opened https://github.com/crossplaneio/crossplane-runtime/issues/51 to track the finalizer UX thing we discussed.

PR #46, opened 2019-10-10
Implementating ReferenceResolver in managed reconciler

soorena776:
Signed-off-by: soorena776 <javad@upbound.io>

Fixes crossplaneio/crossplane#887

### Checklist

I have:

### DIFF
--- a/apis/core/v1alpha1/condition.go
+++ b/apis/core/v1alpha1/condition.go
@@ -34,6 +34,9 @@ const (
 	// TypeSynced managed resources are believed to be in sync with the
 	// Kubernetes resources that manage their lifecycle.
 	TypeSynced ConditionType = "Synced"
+
+	// TypeReferencesResolved managed resources' references are resolved
+	TypeReferencesResolved = "ReferencesResolved"
 )
 
 // A ConditionReason represents the reason a resource is in a condition.
@@ -51,6 +54,13 @@ const (
 const (
 	ReasonReconcileSuccess ConditionReason = "Successfully reconciled managed resource"
 	ReasonReconcileError   ConditionReason = "Encountered an error during managed resource reconciliation"
+)
+
+// Reason references for a resource are or are not resolved
+const (
+	ReasonReferenceResolveSuccess  ConditionReason = "Successfully resolved managed resource references to other resources"
+	ReasonResolveReferencesBlocked ConditionReason = "One or more of referenced resources do not exist, or are not yet Ready"
+	ReasonReferenceResolveError    ConditionReason = "Encountered an error while resolving managed resource references to other resources"
 )
 
 // A Condition that may apply to a managed resource.
@@ -111,6 +121,25 @@ func NewConditionedStatus(c ...Condition) *ConditionedStatus {
 	s := &ConditionedStatus{}
 	s.SetConditions(c...)
 	return s
+}
+
+// GetCondition returns the condition for the given ConditionType if exists,
+// otherwise returns nil
+func (s *ConditionedStatus) GetCondition(ct ConditionType) Condition {
+	for _, c := range s.Conditions {
+		if c.Type == ct {
+			return c
+		}
+	}
+
+	return Condition{Type: ct, Status: corev1.ConditionUnknown}
+}
+
+// IsConditionTrue returns a boolean indicating whether the given condition
+// exists and is true
+func (s *ConditionedStatus) IsConditionTrue(ct ConditionType) bool {
+	cond := s.GetCondition(ct)
+	return cond.Status != corev1.ConditionTrue
 }
 
 // SetConditions sets the supplied conditions, replacing any existing conditions
@@ -236,6 +265,31 @@ func ReconcileError(err error) Condition {
 		Status:             corev1.ConditionFalse,
 		LastTransitionTime: metav1.Now(),
 		Reason:             ReasonReconcileError,
+		Message:            err.Error(),
+	}
+}
+
+// ReferenceResolutionSuccess returns a condition indicating that Crossplane
+// successfully resolved the references used in the managed resource
+func ReferenceResolutionSuccess() Condition {
+	return Condition{
+		Type:               TypeReferencesResolved,
+		Status:             corev1.ConditionTrue,
+		LastTransitionTime: metav1.Now(),
+		Reason:             ReasonReferenceResolveSuccess,
+	}
+}
+
+// ReferenceResolutionBlocked returns a condition indicating that Crossplane is
+// unable to resolve the references used in the managed resource. This could
+// mean that one or more of referred resources do not yet exist, or are not yet
+// Ready
+func ReferenceResolutionBlocked(err error) Condition {
+	return Condition{
+		Type:               TypeReferencesResolved,
+		Status:             corev1.ConditionFalse,
+		LastTransitionTime: metav1.Now(),
+		Reason:             ReasonResolveReferencesBlocked,
 		Message:            err.Error(),
 	}
 }
--- a/apis/core/v1alpha1/condition.go
+++ b/apis/core/v1alpha1/condition.go
@@ -60,7 +60,6 @@ const (
 const (
 	ReasonReferenceResolveSuccess  ConditionReason = "Successfully resolved managed resource references to other resources"
 	ReasonResolveReferencesBlocked ConditionReason = "One or more of referenced resources do not exist, or are not yet Ready"
-	ReasonReferenceResolveError    ConditionReason = "Encountered an error while resolving managed resource references to other resources"
 )
 
 // A Condition that may apply to a managed resource.
--- a/apis/core/v1alpha1/condition.go
+++ b/apis/core/v1alpha1/condition.go
@@ -134,13 +134,6 @@ func (s *ConditionedStatus) GetCondition(ct ConditionType) Condition {
 	return Condition{Type: ct, Status: corev1.ConditionUnknown}
 }
 
-// IsConditionTrue returns a boolean indicating whether the given condition
-// exists and is true
-func (s *ConditionedStatus) IsConditionTrue(ct ConditionType) bool {
-	cond := s.GetCondition(ct)
-	return cond.Status != corev1.ConditionTrue
-}
-
 // SetConditions sets the supplied conditions, replacing any existing conditions
 // of the same type. This is a no-op if all supplied conditions are identical,
 // ignoring the last transition time, to those already set.
--- a/pkg/resource/claim_reconciler.go
+++ b/pkg/resource/claim_reconciler.go
@@ -183,7 +183,6 @@ type crManaged struct {
 	ManagedConnectionPropagator
 	ManagedBinder
 	ManagedFinalizer
-	ManagedReferenceResolver
 }
 
 func defaultCRManaged(m manager.Manager) crManaged {
@@ -193,7 +192,6 @@ func defaultCRManaged(m manager.Manager) crManaged {
 		ManagedConnectionPropagator: NewAPIManagedConnectionPropagator(m.GetClient(), m.GetScheme()),
 		ManagedBinder:               NewAPIManagedBinder(m.GetClient()),
 		ManagedFinalizer:            NewAPIManagedUnbinder(m.GetClient()),
-		ManagedReferenceResolver:    NewReferenceResolver(m.GetClient()),
 	}
 }
 
@@ -370,21 +368,6 @@ func (r *ClaimReconciler) Reconcile(req reconcile.Request) (reconcile.Result, er
 			// class is (re)created.
 			claim.SetConditions(v1alpha1.Creating(), v1alpha1.ReconcileError(err))
 			return reconcile.Result{RequeueAfter: aShortWait}, errors.Wrap(r.client.Status().Update(ctx, claim), errUpdateClaimStatus)
-		}
-
-		if !IsConditionTrue(claim.GetCondition(v1alpha1.TypeReferencesResolved)) {
-			if err := r.managed.ResolveReferences(ctx, class); err != nil {
-				condition := v1alpha1.ReconcileError(err)
-				if IsReferencesAccessError(err) {
-					condition = v1alpha1.ReferenceResolutionBlocked(err)
-				}
-
-				claim.SetConditions(condition)
-				return reconcile.Result{RequeueAfter: aShortWait}, errors.Wrap(r.client.Status().Update(ctx, claim), errUpdateClaimStatus)
-			}
-
-			// Add ReferenceResolutionSuccess to claim conditions
-			claim.SetConditions(v1alpha1.ReferenceResolutionSuccess())
 		}
 
 		if err := r.managed.Configure(ctx, claim, class, managed); err != nil {
--- a/pkg/resource/claim_reconciler.go
+++ b/pkg/resource/claim_reconciler.go
@@ -183,6 +183,7 @@ type crManaged struct {
 	ManagedConnectionPropagator
 	ManagedBinder
 	ManagedFinalizer
+	ManagedReferenceResolver
 }
 
 func defaultCRManaged(m manager.Manager) crManaged {
@@ -192,6 +193,7 @@ func defaultCRManaged(m manager.Manager) crManaged {
 		ManagedConnectionPropagator: NewAPIManagedConnectionPropagator(m.GetClient(), m.GetScheme()),
 		ManagedBinder:               NewAPIManagedBinder(m.GetClient()),
 		ManagedFinalizer:            NewAPIManagedUnbinder(m.GetClient()),
+		ManagedReferenceResolver:    NewReferenceResolver(m.GetClient()),
 	}
 }
 
@@ -368,6 +370,21 @@ func (r *ClaimReconciler) Reconcile(req reconcile.Request) (reconcile.Result, er
 			// class is (re)created.
 			claim.SetConditions(v1alpha1.Creating(), v1alpha1.ReconcileError(err))
 			return reconcile.Result{RequeueAfter: aShortWait}, errors.Wrap(r.client.Status().Update(ctx, claim), errUpdateClaimStatus)
+		}
+
+		if !IsConditionTrue(claim.GetCondition(v1alpha1.TypeReferencesResolved)) {
+			if err := r.managed.ResolveReferences(ctx, class); err != nil {
+				condition := v1alpha1.ReconcileError(err)
+				if IsReferencesAccessError(err) {
+					condition = v1alpha1.ReferenceResolutionBlocked(err)
+				}
+
+				claim.SetConditions(condition)
+				return reconcile.Result{RequeueAfter: aShortWait}, errors.Wrap(r.client.Status().Update(ctx, claim), errUpdateClaimStatus)
+			}
+
+			// Add ReferenceResolutionSuccess to claim conditions
+			claim.SetConditions(v1alpha1.ReferenceResolutionSuccess())
 		}
 
 		if err := r.managed.Configure(ctx, claim, class, managed); err != nil {
--- a/pkg/resource/managed_reconciler.go
+++ b/pkg/resource/managed_reconciler.go
@@ -245,7 +245,7 @@ func defaultMRManaged(m manager.Manager) mrManaged {
 			NewManagedNameAsExternalName(m.GetClient()),
 			NewAPIManagedFinalizerAdder(m.GetClient()),
 		},
-		ManagedReferenceResolver: NewReferenceResolver(m.GetClient()),
+		ManagedReferenceResolver: NewAPIManagedReferenceResolver(m.GetClient()),
 	}
 }
 
--- a/pkg/resource/managed_reconciler.go
+++ b/pkg/resource/managed_reconciler.go
@@ -431,14 +431,12 @@ func (r *ManagedReconciler) Reconcile(req reconcile.Request) (reconcile.Result, 
 	if !observation.ResourceExists {
 		if !IsConditionTrue(managed.GetCondition(v1alpha1.TypeReferencesResolved)) {
 			if err := r.managed.ResolveReferences(ctx, managed); err != nil {
-				// update the status according to the type of the err
-				switch err.(type) {
-				case NotReadyError:
-					managed.SetConditions(v1alpha1.ReferenceResolutionBlocked(err))
-				default:
-					managed.SetConditions(v1alpha1.ReconcileError(err))
+				condition := v1alpha1.ReconcileError(err)
+				if IsReferencesAccessError(err) {
+					condition = v1alpha1.ReferenceResolutionBlocked(err)
 				}
 
+				managed.SetConditions(condition)
 				return reconcile.Result{RequeueAfter: r.longWait}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
 			}
 
--- a/pkg/resource/managed_reconciler.go
+++ b/pkg/resource/managed_reconciler.go
@@ -90,11 +90,11 @@ func (m ManagedInitializerFn) Initialize(ctx context.Context, mg Managed) error 
 }
 
 // ManagedReferenceResolverFn is the pluggable struct to produce objects with ManagedReferenceResolver interface.
-type ManagedReferenceResolverFn func(context.Context, Managed) error
+type ManagedReferenceResolverFn func(context.Context, CanReference) error
 
 // ResolveReferences calls ManagedReferenceResolverFn function
-func (m ManagedReferenceResolverFn) ResolveReferences(ctx context.Context, mg Managed) error {
-	return m(ctx, mg)
+func (m ManagedReferenceResolverFn) ResolveReferences(ctx context.Context, res CanReference) error {
+	return m(ctx, res)
 }
 
 // An ExternalConnecter produces a new ExternalClient given the supplied
--- a/pkg/resource/managed_reconciler.go
+++ b/pkg/resource/managed_reconciler.go
@@ -372,6 +372,21 @@ func (r *ManagedReconciler) Reconcile(req reconcile.Request) (reconcile.Result, 
 		return reconcile.Result{RequeueAfter: r.shortWait}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
 	}
 
+	if !IsConditionTrue(managed.GetCondition(v1alpha1.TypeReferencesResolved)) {
+		if err := r.managed.ResolveReferences(ctx, managed); err != nil {
+			condition := v1alpha1.ReconcileError(err)
+			if IsReferencesAccessError(err) {
+				condition = v1alpha1.ReferenceResolutionBlocked(err)
+			}
+
+			managed.SetConditions(condition)
+			return reconcile.Result{RequeueAfter: r.longWait}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
+		}
+
+		// Add ReferenceResolutionSuccess to the conditions
+		managed.SetConditions(v1alpha1.ReferenceResolutionSuccess())
+	}
+
 	observation, err := external.Observe(ctx, managed)
 	if err != nil {
 		// We'll usually hit this case if our Provider credentials are invalid
@@ -429,21 +444,6 @@ func (r *ManagedReconciler) Reconcile(req reconcile.Request) (reconcile.Result, 
 	}
 
 	if !observation.ResourceExists {
-		if !IsConditionTrue(managed.GetCondition(v1alpha1.TypeReferencesResolved)) {
-			if err := r.managed.ResolveReferences(ctx, managed); err != nil {
-				condition := v1alpha1.ReconcileError(err)
-				if IsReferencesAccessError(err) {
-					condition = v1alpha1.ReferenceResolutionBlocked(err)
-				}
-
-				managed.SetConditions(condition)
-				return reconcile.Result{RequeueAfter: r.longWait}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
-			}
-
-			// Add ReferenceResolutionSuccess to the conditions
-			managed.SetConditions(v1alpha1.ReferenceResolutionSuccess())
-		}
-
 		creation, err := external.Create(ctx, managed)
 		if err != nil {
 			// We'll hit this condition if we can't create our external
--- a/pkg/resource/managed_reconciler.go
+++ b/pkg/resource/managed_reconciler.go
@@ -32,8 +32,10 @@ import (
 )
 
 const (
-	managedControllerName   = "managedresource.crossplane.io"
-	managedFinalizerName    = "finalizer." + managedControllerName
+	managedControllerName               = "managedresource.crossplane.io"
+	managedFinalizerName                = "finalizer." + managedControllerName
+	managedResourceStructTagPackageName = "resource"
+
 	managedReconcileTimeout = 1 * time.Minute
 
 	defaultManagedShortWait = 30 * time.Second
@@ -84,6 +86,14 @@ type ManagedInitializerFn func(ctx context.Context, mg Managed) error
 
 // Initialize calls ManagedInitializerFn function.
 func (m ManagedInitializerFn) Initialize(ctx context.Context, mg Managed) error {
+	return m(ctx, mg)
+}
+
+// ManagedReferenceResolverFn is the pluggable struct to produce objects with ManagedReferenceResolver interface.
+type ManagedReferenceResolverFn func(context.Context, Managed) error
+
+// ResolveReferences calls ManagedReferenceResolverFn function
+func (m ManagedReferenceResolverFn) ResolveReferences(ctx context.Context, mg Managed) error {
 	return m(ctx, mg)
 }
 
@@ -224,6 +234,7 @@ type mrManaged struct {
 	ManagedConnectionPublisher
 	ManagedFinalizer
 	ManagedInitializer
+	ManagedReferenceResolver
 }
 
 func defaultMRManaged(m manager.Manager) mrManaged {
@@ -234,6 +245,7 @@ func defaultMRManaged(m manager.Manager) mrManaged {
 			NewManagedNameAsExternalName(m.GetClient()),
 			NewAPIManagedFinalizerAdder(m.GetClient()),
 		},
+		ManagedReferenceResolver: NewReferenceResolver(m.GetClient()),
 	}
 }
 
@@ -417,6 +429,23 @@ func (r *ManagedReconciler) Reconcile(req reconcile.Request) (reconcile.Result, 
 	}
 
 	if !observation.ResourceExists {
+		if !IsConditionTrue(managed.GetCondition(v1alpha1.TypeReferencesResolved)) {
+			if err := r.managed.ResolveReferences(ctx, managed); err != nil {
+				// update the status according to the type of the err
+				switch err.(type) {
+				case NotReadyError:
+					managed.SetConditions(v1alpha1.ReferenceResolutionBlocked(err))
+				default:
+					managed.SetConditions(v1alpha1.ReconcileError(err))
+				}
+
+				return reconcile.Result{RequeueAfter: r.longWait}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
+			}
+
+			// Add ReferenceResolutionSuccess to the conditions
+			managed.SetConditions(v1alpha1.ReferenceResolutionSuccess())
+		}
+
 		creation, err := external.Create(ctx, managed)
 		if err != nil {
 			// We'll hit this condition if we can't create our external
--- a/pkg/resource/reference_resolver.go
+++ b/pkg/resource/reference_resolver.go
@@ -44,28 +44,60 @@ type fieldHasTagPair struct {
 	hasTheTag  bool
 }
 
-// NotReadyError is a custom error interface, that indicates the resource is not ready
-type NotReadyError interface {
-	error
+// ReferenceStatusType is an enum type for the possible values for a Reference Status
+type ReferenceStatusType int
 
-	resources() []string
+const (
+	// ReferenceStatusUnknown is the default value
+	ReferenceStatusUnknown ReferenceStatusType = iota
+	// ReferenceNotFound shows that the reference is not found
+	ReferenceNotFound
+	// ReferenceNotReady shows that the reference is not ready
+	ReferenceNotReady
+	// ReferenceReady shows that the reference is ready
+	ReferenceReady
+)
+
+func (t ReferenceStatusType) String() string {
+	return []string{"Unknown", "NotFound", "NotReady", "Ready"}[t]
 }
 
-type resourceNotReadyErr struct {
-	items []string
+// ReferenceStatus has the name and status of a reference
+type ReferenceStatus struct {
+	Name   string
+	Status ReferenceStatusType
 }
 
-// NewNotReadyErr returns a new NotReadyError
-func NewNotReadyErr(items []string) NotReadyError {
-	return &resourceNotReadyErr{items}
+func (r ReferenceStatus) String() string {
+	return fmt.Sprintf("%s:%s,", r.Name, r.Status)
 }
 
-func (r *resourceNotReadyErr) Error() string {
-	return fmt.Sprintf("These resources are not ready to be referenced: %s", r.items)
+// referencesAccessErr is used to indicate that one or more references can not
+// be accessed
+type referencesAccessErr struct {
+	statuses []ReferenceStatus
 }
 
-func (r *resourceNotReadyErr) resources() []string {
-	return r.items
+// newReferenceAccessErr returns a referencesAccessErr if any of the given
+// references are not ready
+func newReferenceAccessErr(statuses []ReferenceStatus) error {
+	for _, st := range statuses {
+		if st.Status != ReferenceReady {
+			return &referencesAccessErr{statuses}
+		}
+	}
+
+	return nil
+}
+
+func (r *referencesAccessErr) Error() string {
+	return fmt.Sprintf("Some of the referenced resources cannot be accessed. {%s}", r.statuses)
+}
+
+// IsReferencesAccessError returns true if the given error is of type referencesAccessErr
+func IsReferencesAccessError(err error) bool {
+	_, result := err.(*referencesAccessErr)
+	return result
 }
 
 // AttributeReferencer is an interface for referencing and resolving
@@ -74,13 +106,9 @@ func (r *resourceNotReadyErr) resources() []string {
 // for more information
 type AttributeReferencer interface {
 
-	// ValidateReady validates that the referenced managed resource type has the
-	// 'Ready' condition type and it is 'True'. If the returned error is nil,
-	// the resource is interpretted to be `Ready`. Otherwise, if the error type
-	// is `NotReadyError`, then the reference resolution status is updated as
-	// `ReferenceResolutionBlocked`, otherwise it updates the status as
-	// `ReconcileError`
-	ValidateReady(context.Context, Managed, client.Reader) error
+	// GetStatus looks up the referenced objects in K8S api and returns a list
+	// of ReferenceStatus
+	GetStatus(context.Context, Managed, client.Reader) ([]ReferenceStatus, error)
 
 	// Build retrieves referenced resource, as well as other non-managed
 	// resources (like a `Provider`), and builds the referenced attribute
@@ -129,10 +157,18 @@ func (r *APIManagedReferenceResolver) ResolveReferences(ctx context.Context, mg 
 	}()
 
 	// make sure that all the references are ready
+	allStatuses := []ReferenceStatus{}
 	for _, referencer := range referencers {
-		if err := referencer.ValidateReady(ctx, mg, r.reader); err != nil {
+		statuses, err := referencer.GetStatus(ctx, mg, r.reader)
+		if err != nil {
 			return err
 		}
+
+		allStatuses = append(allStatuses, statuses...)
+	}
+
+	if err := newReferenceAccessErr(allStatuses); err != nil {
+		return err
 	}
 
 	// build and assign the attributes
--- a/pkg/resource/reference_resolver.go
+++ b/pkg/resource/reference_resolver.go
@@ -1,0 +1,231 @@
+/*
+Copyright 2019 The Crossplane Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package resource
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+
+	"github.com/pkg/errors"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	attributeReferencerTagName = "attributereferencer"
+	errPanicedResolving        = "paniced while resolving references: %v"
+)
+
+// Error strings
+const (
+	errTaggedFieldlNotImplemented = "ManagedReferenceResolver: the field has the %v tag, but has not implemented AttributeReferencer interface"
+	errBuildAttribute             = "ManagedReferenceResolver: could not build the attribute"
+	errAssignAttribute            = "ManagedReferenceResolver: could not assign the attribute"
+)
+
+// fieldHasTagPair is used in findAttributeReferencerFields
+type fieldHasTagPair struct {
+	fieldValue reflect.Value
+	hasTheTag  bool
+}
+
+// NotReadyError is a custom error interface, that indicates the resource is not ready
+type NotReadyError interface {
+	error
+
+	resources() []string
+}
+
+type resourceNotReadyErr struct {
+	items []string
+}
+
+// NewNotReadyErr returns a new NotReadyError
+func NewNotReadyErr(items []string) NotReadyError {
+	return &resourceNotReadyErr{items}
+}
+
+func (r *resourceNotReadyErr) Error() string {
+	return fmt.Sprintf("These resources are not ready to be referenced: %s", r.items)
+}
+
+func (r *resourceNotReadyErr) resources() []string {
+	return r.items
+}
+
+// AttributeReferencer is an interface for referencing and resolving
+// cross-resource attribute references. See
+// https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-cross-resource-referencing.md
+// for more information
+type AttributeReferencer interface {
+
+	// ValidateReady validates that the referenced managed resource type has the
+	// 'Ready' condition type and it is 'True'. If the returned error is nil,
+	// the resource is interpretted to be `Ready`. Otherwise, if the error type
+	// is `NotReadyError`, then the reference resolution status is updated as
+	// `ReferenceResolutionBlocked`, otherwise it updates the status as
+	// `ReconcileError`
+	ValidateReady(context.Context, Managed, client.Reader) error
+
+	// Build retrieves referenced resource, as well as other non-managed
+	// resources (like a `Provider`), and builds the referenced attribute
+	Build(context.Context, Managed, client.Reader) (string, error)
+
+	// Assign accepts a managed resource object, and assigns the given value to the
+	// corresponding property
+	Assign(Managed, string) error
+}
+
+// A ManagedReferenceResolver resolves the references to other managed
+// resources, by looking them up in the Kubernetes API server. The references
+// are the fields in the managed resource that implement AttributeReferencer
+// interface and have
+// `attributeReferencerTagName:"managedResourceStructTagPackageName"` tag
+type ManagedReferenceResolver interface {
+	ResolveReferences(context.Context, Managed) error
+}
+
+// APIManagedReferenceResolver resolves implements ManagedReferenceResolver interface
+type APIManagedReferenceResolver struct {
+	reader client.Reader
+}
+
+// NewReferenceResolver returns a new APIManagedReferenceResolver
+func NewReferenceResolver(c client.Reader) *APIManagedReferenceResolver {
+	return &APIManagedReferenceResolver{c}
+}
+
+// ResolveReferences resolves references made to other managed resources
+func (r *APIManagedReferenceResolver) ResolveReferences(ctx context.Context, mg Managed) (err error) {
+	// retrieve all the referencer fields from the managed resource
+	referencers, err := findAttributeReferencerFields(mg, false)
+	if err != nil {
+		// if there is an error it should immediately panic, since this means an
+		// attribute is tagged but doesn't implement AttributeReferencer
+		panic(err)
+	}
+
+	// this recovers from potential panics during execution of
+	// AttributeReferencer methods
+	defer func() {
+		if r := recover(); r != nil {
+			err = errors.Errorf(errPanicedResolving, r)
+		}
+	}()
+
+	// make sure that all the references are ready
+	for _, referencer := range referencers {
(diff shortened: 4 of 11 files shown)


pkg/resource/reference_resolver.go
```diff
@@ -0,0 +1,140 @@
...
+// ReferenceResolver is an interface for resolving cross resource attribute references.
+// It should be implemented by strcuts that are used for custom resources feilds.
+// See #874 for more information
+type ReferenceResolver interface {
+	// Resolve retrieve and builds the referenced attribute
+	Resolve(reader client.Reader) (value interface{}, err error)
+
+	// Assign accepts a resource object, and assigns the given value to the corresponding property
+	Assign(resource interface{}, value interface{}) error
+}
```
soorena776: For instance, the VPCID referencer will be implemented as:

```go
type VPCIDResolver struct {
	VPC *LocalObjectReference
}

func (v *VPCIDResolver) Resolve(reader client.Reader) (interface{}, error) {

	var vpc network.VPC
	if err := reader.Get(ctx, v.VPC, &vpc); err != nil {
		return nil, err
	}

	return vpc.Id, nil
}

func (v *VPCIDResolver) Assign(resource interface{}, value interface{}) error {
	subnet, isExpectedType := resource.(*network.Subnet)
	if !isExpectedType {
		return fmt.Errorf("unexpected resource type")
	}

	subnet.Spec.VPCID = value.(string)
	return nil
}
```

and then will be used in [Subnet type](https://github.com/crossplaneio/stack-aws/blob/master/apis/network/v1alpha2/subnet_types.go#L28) like:
```go
VPCIDRef *VPCIDResolver `json:"vpcIdRef,omitempty" managed_resource:"reference_resolver"`
```

soorena776: the `managed_resource:"reference_resolver"` is intended to help readability of the code, but visually linking it to managed reconciler's reference resolution mechanism.

negz: I'm still concerned that this results in a an API like:

```yaml
vpcIDRef:  # This looks like it will be an object reference
  vpc:  # ...but it is not
    name: myVPC
```

pkg/resource/reference_resolver.go
```diff
@@ -0,0 +1,140 @@
...
+
+// ReferenceResolver is an interface for resolving cross resource attribute references.
+// It should be implemented by strcuts that are used for custom resources feilds.
+// See #874 for more information
+type ReferenceResolver interface {
+	// Resolve retrieve and builds the referenced attribute
+	Resolve(reader client.Reader) (value interface{}, err error)
+
+	// Assign accepts a resource object, and assigns the given value to the corresponding property
+	Assign(resource interface{}, value interface{}) error
```
negz: Could we use `mg Managed` rather than `resource interface{}`?

pkg/resource/reference_resolver.go
```diff
@@ -0,0 +1,140 @@
...
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// ReferenceResolver is an interface for resolving cross resource attribute references.
+// It should be implemented by strcuts that are used for custom resources feilds.
+// See #874 for more information
+type ReferenceResolver interface {
+	// Resolve retrieve and builds the referenced attribute
+	Resolve(reader client.Reader) (value interface{}, err error)
```
negz: Could `value` be `string` rather than `interface{}`? I prefer to avoid the empty interface where possible, and I can't think of many (any?) cases where a reference would be resolved to an integer, byte array, struct, etc.

pkg/resource/reference_resolver.go
```diff
@@ -0,0 +1,140 @@
...
+	// ignore nil fields
+	if v.Kind() == reflect.Ptr && v.IsNil() {
+		return nil, nil
+	}
+
+	if objHasTheRightTag {
+		if ar, implements := obj.(ReferenceResolver); implements {
+			return []ReferenceResolver{ar}, nil
+		} else if v.Kind() == reflect.Struct {
+			return nil, fmt.Errorf("the field has the 'reference_resolver' tag, but has not implemented ReferenceResolver interface")
```
negz: I'd prefer to stick to our convention of using `pkg/errors` - i.e. `errors.Errorf` here.

pkg/resource/reference_resolver.go
```diff
@@ -0,0 +1,140 @@
...
+	for _, resolver := range resolvers {
+		r := reflect.ValueOf(resolver).Elem()
+		for i := 0; i < r.NumField() && r.Field(i).CanInterface(); i++ {
+			if ref, ok := r.Field(i).Interface().(corev1.LocalObjectReference); ok {
+				result = append(result, ref)
+			}
+
+			if ref, ok := r.Field(i).Interface().(*corev1.LocalObjectReference); ok {
+				result = append(result, *ref)
+			}
```
negz: Is the intent here to support both required and optional references (i.e. object reference and pointer to object reference)? Could we simplify this by requiring that all references are either required or optional?

pkg/resource/reference_resolver.go
```diff
@@ -0,0 +1,140 @@
...
+	case reflect.Struct:
+		for i := 0; i < v.NumField() && v.Field(i).CanInterface(); i++ {
+			val, ok := reflect.TypeOf(obj).Field(i).Tag.Lookup("managed_resource")
+			hasTheRightTag := ok && val == "reference_resolver"
+
+			if resolvers, err := findRefResolverFields(v.Field(i).Interface(), hasTheRightTag); err == nil {
+				result = append(result, resolvers...)
+			} else {
+				return nil, err
+			}
```
negz:
```go
resolvers, err := findRefResolverFields(v.Field(i).Interface(), hasTheRightTag);
if err != nil {
	return nil, err
}
result = append(result, resolvers...)
```

Would the above logic be more idiomatic?

Disclaimer: I have an unusually strong aversion to `else` and often go to awkward lengths to avoid using it.

https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow
https://blog.timoxley.com/post/47041269194/avoid-else-return-early

pkg/resource/reference_resolver.go
```diff
@@ -0,0 +1,140 @@
...
+type ReferenceResolver interface {
+	// Resolve retrieve and builds the referenced attribute
+	Resolve(reader client.Reader) (value interface{}, err error)
+
+	// Assign accepts a resource object, and assigns the given value to the corresponding property
+	Assign(resource interface{}, value interface{}) error
+}
+
+// ResolveReferencers resolves references made to other resources
+func ResolveReferencers(ctx context.Context, mgd Managed, c client.Reader) (bool, error) {
```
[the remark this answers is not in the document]
soorena776: Agreed, references seem to be a better wording.

pkg/resource/reference_resolver.go
soorena776: I can change it to string for now. But I want to be vary of limitations that this will imply in the future.

pkg/resource/reference_resolver.go
soorena776: I think that's a good idea.

pkg/resource/reference_resolver.go
soorena776: I don't have a strong opinion except that I am trying to keep `err` internal to the `if` block. I'll implement your suggestion.

pkg/resource/reference_resolver.go
```diff
@@ -0,0 +1,140 @@
...
+		} else if v.Kind() == reflect.Struct {
+			return nil, fmt.Errorf("the field has the 'reference_resolver' tag, but has not implemented ReferenceResolver interface")
+		}
+	}
+
+	switch v.Kind() {
+	case reflect.Struct:
+		for i := 0; i < v.NumField() && v.Field(i).CanInterface(); i++ {
+			val, ok := reflect.TypeOf(obj).Field(i).Tag.Lookup("managed_resource")
+			hasTheRightTag := ok && val == "reference_resolver"
```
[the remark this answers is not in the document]
soorena776: These strings technically are called [`packages`](https://github.com/golang/go/commit/25733a94fde4f4af4b6ac5ba01b7212a3ef0f013). 
And since there related to `resource` package, I gave them the more specific name `managed_resource`. I prefer to keep it this way as `crossplane` is less informative when the code itself is in Crossplane project.

pkg/resource/reference_resolver.go
soorena776: @negz for single value object resolvers like `VPCIDResolver`, we could embed the reference:
```
type VPCIDResolver struct {
	*LocalObjectReference
}
```
But for composite attributes, I don't see any other way around this.

apis/core/v1alpha1/condition.go
```diff
@@ -53,6 +56,13 @@ const (
 	ReasonReconcileError   ConditionReason = "Encountered an error during managed resource reconciliation"
 )
 
+// Reason references for a resource are or are not resolved
+const (
+	ReasonReferenceResolveSuccess  ConditionReason = "Successfully resolved managed resource references to other resources"
+	ReasonResolveReferencesBlocked ConditionReason = "One or more of referred resources do not exists, or are not yet Ready"
+	ReasonReferenceResolveError    ConditionReason = "Encountered an error during resolving managed resource references to other resources"
+)
```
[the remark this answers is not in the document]
soorena776: Are you suggesting to use the existing `TypeReady`, instead of creating creating the new `TypeReferencesResolved` ConditionType? If so I'd slightly prefer to create a new type to avoid modifying the existing code. Also, a separate condition would tell the history of reference resolution in resource rather than re-writing it with the most up to date `Ready` condition.

pkg/resource/reference_resolver.go
negz: https://github.com/crossplaneio/crossplane-runtime/issues/49

> for single value object resolvers like VPCIDResolver, we could embed the reference:

I'd prefer this over the current pattern. I think implementing our own type would be better, for the reasons outlined in the above issue. Using our own type would let us make `name` a required field.

> But for composite attributes, I don't see any other way around this.

I'm still not convinced this use case exists. Did we have any examples of composite attributes in which the additional attributes could not be derived from a single attribute? i.e. Deriving the project ID from the network name.

soorena776: `Subnetwork` in GCP is the only case that I can think of. If we can retrieve the ProjectId from the network, we are good Otherwise we will need some api object from which we can retrieve the project id

pkg/resource/managed_reconciler.go
```diff
@@ -406,6 +410,23 @@ func (r *ManagedReconciler) Reconcile(req reconcile.Request) (reconcile.Result,
...
 		}
 
+		if !managed.IsConditionTrue(v1alpha1.TypeReferencesResolved) {
+			if err := r.managed.ResolveReferences(ctx, managed); err != nil {
+				// update the status according to the type of the err
+				switch err.(type) {
+				case *resourceNotReadyErr:
+					managed.SetConditions(v1alpha1.ResolveReferencesBlocked(err))
+				default:
+					managed.SetConditions(v1alpha1.ResolveReferencesError(err))
```
negz: On reading this block I'd prefer to avoid introducing a distinct `ResolveReferencesError`.

Every other case in which we encounter an unexpected error during reconciliation is reflected using `ReconcileError`, making this the only case we do something different.

pkg/resource/reference_resolver.go
```diff
@@ -0,0 +1,222 @@
...
+// AttributeReferencer is an interface for referencing and resolving
+// cross-resource attribute references. See
+// https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-cross-resource-referencing.md
+// for more information
+type AttributeReferencer interface {
+	// Build retrieves and builds the referenced attribute
+	Build(getter LocalObjectGetter) (value string, err error)
+
+	// Assign accepts a resource object, and assigns the given value to the corresponding property
+	Assign(mgd Managed, value string) error
```
negz: Nitpick: `Managed` parameters are named `mg`, not `mgd` almost everywhere else in the codebase.

pkg/resource/reference_resolver.go
```diff
@@ -0,0 +1,222 @@
...
+}
+
+// A ManagedReferenceResolver resolves the references to other managed resources
+type ManagedReferenceResolver interface {
+	ResolveReferences(context.Context, Managed) error
+}
+
+// ReferenceResolver resolves implements ManagedReferenceResolver interface
```
negz: Nitpick: I know this is the only real implementation of this interface, but I suggest adding a little context as to how it satisfies the interface. i.e. "A ReferenceResolver resolves references to other managed resources by looking them up in the Kubernetes API server". There's an established pattern of these default implementations of interfaces being named `APIFoo` (.e.g `APIManagedReferenceResolver`).

pkg/resource/reference_resolver.go
```diff
@@ -0,0 +1,222 @@
...
+
+// NewReferenceResolver returns a new ReferenceResolver
+func NewReferenceResolver(c client.Reader) *ReferenceResolver {
+	return &ReferenceResolver{c}
+}
+
+// ResolveReferences resolves references made to other managed resources
+func (r *ReferenceResolver) ResolveReferences(ctx context.Context, mgd Managed) (err error) {
+	// this recovers from potential panics during execution of
+	// AttributeReferencer methods
```
negz: Under what circumstances would these methods panic? Could we design this such that there was no risk of panicing, or is this risk always inherent when using reflection?

pkg/resource/reference_resolver.go
```diff
@@ -0,0 +1,222 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// getterFunc returns a function to retrieve local references
+func (r *ReferenceResolver) getterFunc(ctx context.Context, mgd Managed) LocalObjectGetter {
```
negz: Nitpick: I see function shortened to `Fn` more commonly than `Func` in Go. The only citation I can provide for this off the top of my head is  https://godoc.org/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#CreateOrUpdate

pkg/resource/reference_resolver.go
```diff
@@ -0,0 +1,222 @@
...
+				Namespace: mgd.GetNamespace(),
+				Name:      ref.Name,
+			},
+			obj)
+
+		if err != nil {
+			return err
+		}
+
+		return nil
```
negz: This function could omit the error handling case at the end by writing:

```go
return r.reader.Get(ctx,
	types.NamespacedName{
		Namespace: mgd.GetNamespace(),
		Name:      ref.Name,
		},
	obj)
```

How would you feel about inlining this in `ResolveReferences`? given how simple this method is I suspect that would be easier to read.

pkg/resource/reference_resolver.go
```diff
@@ -0,0 +1,222 @@
...
+	for _, referencer := range referencers {
+		val, err := referencer.Build(getter)
+		if err != nil {
+			return err
+		}
+
+		err = referencer.Assign(mgd, val)
+		if err != nil {
+			return err
+		}
```
negz: https://godoc.org/github.com/pkg/errors

While I don't think we necessarily need to wrap _every_ error, I think it could be useful to wrap the errors here, and possibly in a few other conditions introduced in this PR, to provide a little more context. e.g. Here I think an error string like "could not assign reference: reason" vs "could not build reference: reason" would be illustrative.

https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully explains the rationale in more detail  under "annotating errors".

pkg/resource/reference_resolver.go
```diff
@@ -0,0 +1,222 @@
...
+}
+
+// findRefResolverFields recursively finds all non-nil fields in a struct and its sub-structs
+// that implement AttributeReferencer and have `managed_resource:"reference_resolver"` tag
+func findRefResolverFields(obj interface{}, objHasTheRightTag bool) ([]AttributeReferencer, error) { // nolint:gocyclo
+
+	v := reflect.ValueOf(obj)
+	result := []AttributeReferencer{}
+
+	// ignore nil fields
```
negz: Nitpick: This is a case where I feel the comment explains something that is quite obvious from how the code reads. Even as someone unfamiliar with the `reflect` package this code reads like "is the kind of this thing a pointer, and is it nil? If so return early".

pkg/resource/reference_resolver.go
```diff
@@ -0,0 +1,222 @@
...
+	if v.Kind() == reflect.Ptr && v.IsNil() {
+		return nil, nil
+	}
+
+	if objHasTheRightTag {
+		if ar, implements := obj.(AttributeReferencer); implements {
+			return []AttributeReferencer{ar}, nil
+		} else if v.Kind() == reflect.Struct {
+			return nil, errors.Errorf("the field has the 'reference_resolver' tag, but has not implemented AttributeReferencer interface")
```
negz: This seems like a condition that we'd ideally test for when implementing the type, rather than at runtime. I typically feel like it's appropriate to panic rather than return an error in these cases where a binary is fundamentally broken due to programmer error.

pkg/resource/reference_resolver.go
```diff
@@ -0,0 +1,222 @@
...
+
+	apierrs "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/crossplaneio/crossplane-runtime/apis/core/v1alpha1"
+)
+
+const (
+	attributeReferencerTagName = "attribute_referencer"
```
negz: Would `attributereferencer` be more idiomatic - similar to `omitempty`?

pkg/resource/reference_resolver.go
```diff
@@ -0,0 +1,222 @@
...
+			return err
+		}
+
+		return nil
+	}
+}
+
+// findRefResolverFields recursively finds all non-nil fields in a struct and its sub-structs
+// that implement AttributeReferencer and have `managed_resource:"reference_resolver"` tag
+func findRefResolverFields(obj interface{}, objHasTheRightTag bool) ([]AttributeReferencer, error) { // nolint:gocyclo
```
negz: I think it's a good convention to provide explanatory context when we disable a linter warning, even if it's just something like "this function is easier to understand in its current form than it would be to split it up into several less complex functions".

pkg/resource/reference_resolver.go
```diff
@@ -0,0 +1,222 @@
...
+	referencers, err := findRefResolverFields(mgd, false)
+	if err != nil {
+		return err
+	}
+
+	getter := r.getterFunc(ctx, mgd)
+	err = validateReferencesAreReady(getter, referencers)
+	if err != nil {
+		return err
+	}
```
negz: Nitpick: this could be `if err := validateReferencesAreReady(...); err != nil`

pkg/resource/reference_resolver.go
```diff
@@ -0,0 +1,222 @@
...
+
+	for _, referencer := range referencers {
+		val, err := referencer.Build(getter)
+		if err != nil {
+			return err
+		}
+
+		err = referencer.Assign(mgd, val)
```
negz: Nitpick: could be `if err := referencer.Assign(...); err != nil`

pkg/resource/reference_resolver.go
negz: My understanding is that the `reference_resolver` is intended to increase readability of our API types, rather than being necessary to implement this functionality. If that is the case, and if this reference resolution code would be simpler without the tag (relying only on checking whether types satisfied the `AttributeReferencer` interface), I think I'd prefer we omit the tag.

pkg/resource/reference_resolver.go
```diff
@@ -0,0 +1,222 @@
...
+import (
+	"context"
+	"fmt"
+	"reflect"
+
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	apierrs "k8s.io/apimachinery/pkg/api/errors"
```
negz: The standing convention in Crossplane is to import this package as `kerrors`.

pkg/resource/reference_resolver.go
```diff
@@ -0,0 +1,222 @@
...
+
+	return result
+}
+
+// validateReferencesAreReady checks if all the given LocalObjectReferences
+// exist and have a `TypeReady` condition which is set to `ConditionTrue`
+func validateReferencesAreReady(getter LocalObjectGetter, referencers []AttributeReferencer) error {
+	for _, referencer := range referencers {
+		for _, reference := range getAllObjectReferences(referencer) {
+			var obj runtime.Object
```
negz: Are you sure this works?

If I follow correctly `getter` is effectively always `r.reader.Get`. That method needs to know the type metadata (i.e. the group, version and kind) of the resource it is asked to get - you can't look an object up in the Kubernetes API by namespace and name alone. `Get` determines the type metadata by using reflection to determine the name of the type contained within the `runtime.Object` interface and looking that type up in its scheme (which is pretty much a big map of type names to pointers to instances of those types). Therefore I wouldn't expect passing a nil interface to `Get` to work.

apis/core/v1alpha1/condition.go
```diff
@@ -106,6 +116,25 @@ func NewConditionedStatus(c ...Condition) *ConditionedStatus {
 	return s
 }
 
+// GetCondition returns the condition for the given ConditionType if exists,
+// otherwise returns nil
+func (s *ConditionedStatus) GetCondition(ct ConditionType) *Condition {
+	for i := 0; i < len(s.Conditions); i++ {
```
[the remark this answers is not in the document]
soorena776: I prefer more convenient `range` style as well, but you can't use range for sending references. 

https://github.com/kyoh86/scopelint

apis/core/v1alpha1/condition.go
```diff
@@ -232,3 +261,44 @@ func ReconcileError(err error) Condition {
 		Message:            err.Error(),
 	}
 }
+
+// ResolveReferencesSuccess returns a condition indicating that Crossplane
```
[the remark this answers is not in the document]
soorena776: I think your suggestion make more sense. applied

pkg/resource/managed_reconciler.go
```diff
@@ -406,6 +410,23 @@ func (r *ManagedReconciler) Reconcile(req reconcile.Request) (reconcile.Result,
 			return reconcile.Result{RequeueAfter: r.shortWait}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
 		}
 
+		if !managed.IsConditionTrue(v1alpha1.TypeReferencesResolved) {
+			if err := r.managed.ResolveReferences(ctx, managed); err != nil {
+				// update the status according to the type of the err
+				switch err.(type) {
```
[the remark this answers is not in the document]
soorena776: I thought about this, but avoided thinking it might be too much. I implement the corresponding error interface.

pkg/resource/reference_resolver.go
negz: > If we can retrieve the ProjectId from the network, we are good 

I'm 99.% sure we can. We'd follow the network's provider ref to its provider resource and thus the project used by that provider.

pkg/resource/reference_resolver.go
soorena776: In general there is no any guarantee that the implementation of AttributeReferencer methods won't panic during runtime since it literally executes `go` code. More specifically, I was thinking a scenario where the `Build` method is accessing an array in the referenced resources status, where the array is empty or nil.

So I think this risk is always inherent. Not necessarily because of reflection, but because of running random go code inside those two methods.

pkg/resource/reference_resolver.go
soorena776: good catch!

I thought about inlining, but initially we were thinking about adding some validation logic (e.g. same group and version for the referenced object) and I thought making it a standalone function make sense. Do you think we should keep it as is for such validations in the future?

pkg/resource/reference_resolver.go
soorena776: I very much agree with this, and I'll add context. I have worked with the `pkg/errors` you linked, and one issue is that to `Errorf` resets the error stack where it is called. This is fine for 
 capturing external errors for the first time, but using it later will rewrite the stack. The reason I didn't use this was because I didn't know where we stand regarding this.

It's worth exploring some options and conventions here. Go 1.13 also has added the notion of native [error wrappting](https://golang.org/doc/go1.13#error_wrapping)

pkg/resource/reference_resolver.go
soorena776: The tag doesn't adding any significant complexity, other than just figuring that a field is indeed intended to be a `AttributeReferencer`. Beside increasing readability it is used to manifest the intention of the developer.

Without this tag, it is impossible for the `ReferenceResolver` to decide whether  it should error when a field has some but not all of `AttributeReferencer` methods. Similarly

pkg/resource/reference_resolver.go
soorena776: I meant this to be temporary as a quick workaround for `make reviewable`to pass. I'll simplify it or add more context

pkg/resource/reference_resolver.go
soorena776: You are right, it seems that the GKV is needed to retrieve an object. Do you think the following solution would make sense or would it be too noisy?

Add another method called `RefereneKind` to the `AttributeReference`, to make it look like:
```go
type AttributeReferencer interface {

        // ReferenceKind returns an object that represents an instance of the referenced object
        ReferenceKind() runtime.Object

	// Build retrieves and builds the referenced attribute
	Build(getter LocalObjectGetter) (value string, err error)

	// Assign accepts a resource object, and assigns the given value to the corresponding property
	Assign(mg Managed, value string) error
}
```

and then populating ReferenceKind like 
```go
func (s *Subnet)ReferneceKind() runtime.Object{
    return &VPC{}
}
```

soorena776: It looks weird that this level of details are needed to retrieve an object. According to [the docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/) name and namespace should suffice to uniquely identify an object.

> Namespaces provide a scope for names. Names of resources need to be unique within a namespace, but not across namespaces. Namespaces can not be nested inside one another and each Kubernetes resource can only be in one namespace.

apis/core/v1alpha1/condition.go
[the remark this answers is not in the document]
negz: Oh right I missed that here. Good point!

pkg/resource/reference_resolver.go
soorena776: It seems that a general [Unstructured](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/client.go#L166) object also could be used. Digging more to see this still would need GKV.

pkg/resource/reference_resolver.go
negz: > Go 1.13 also has added the notion of native error wrappting

I think I personally prefer the `pkg/errors` API, but admittedly haven't spent enough time understanding the new support. My preference would be to stick with the established convention of using `pkg/errors` for the time being.

pkg/resource/reference_resolver.go
negz: I'd prefer to inline for now. I'm less concerned about applying validation to the references when they're hard coded in API types (and can thus be manually validated for sanity during the PR process) than I would be if we allowed the reference kind to be specified as part of the resource YAML.

pkg/resource/reference_resolver.go
soorena776: Another option would be to add the a key to the `resource:"attributereferencer kind=vpc.network.aws.crossplane.io/v1alpha2"`

negz: > Maybe there is another api to retrieve an object given the name and namespace?

Based on my understanding of how the actual underlying REST API works, I'm fairly sure there is not. I'm pretty sure the docs you referenced are glossing over the fact that they mean "the name of a particular resource kind is unique within a namespace"; for example it's common to have a `Deployment` and a `Service` both named `myapp` in `mynamespace`. This is because the REST API is like:
```
https://my.api.server/apis/${group}/${version}/namespaces/${namespace}/${kind}/${name}
```

https://itnext.io/deep-dive-into-how-kubernetes-rest-api-works-517c86f1640b
https://kubernetes.io/docs/reference/using-api/api-concepts/

> It seems that a general Unstructured object also could be used. Digging more to see this still would need GKV.

Yeah, we use that extensively in the workload code, but you're correct that you still need to supply it with a GVK for it to work. Such an unstructured object also would not satisfy the `Managed` interface.

> Add another method called RefereneKind to the AttributeReference

I'd prefer if we could find something simpler. I'm playing with some ideas that are taking a long time to come to fruition so I'm going to post this response as is and reply ASAP if something better comes to mind.

negz: What do you think of something like the below pattern?

```go
// This is the interfaces that resolvable references must implement.
type ResolvableReference interface {
	Resolve(context.Context, client.Reader, Managed) error
}

// NetworkReference is an example of a type that would be embedded in a managed
// resource spec.
type NetworkReference struct {
	Name string `json:"name"`
}

func (r *NetworkReference) Resolve(ctx context.Context, c client.Reader, mg Managed) error {
	subnet, ok := obj.(*v1alpha2.Subnetwork)
	if !ok {
		return errors.New("asked to resolve an object that was not a subnet")
	}

	// Let the explicit network win if it's already set.
	if subnet.Spec.Network != "" {
		return nil
	}

	n := types.NamespacedName{Namespace: mg.GetNamespace(), Name: r.Name}
	network := &v1alpha2.Network{}
	if err := c.Get(ctx, n, network); err != nil {
		return errors.Wrap(err, "could not get network")
	}

	p := &gcpv1alpha2.Provider{}
	if err := c.Get(ctx, meta.NamespacedNameOf(network.Spec.ProviderReference), p); err != nil {
		return errors.Wrap(err, "could not get provider")
	}

	subnet.Spec.Network = fmt.Sprintf("project/%s/global/networks/%s", p.Spec.ProjectID, network.Spec.Name)

	return nil
}
```

We'd maintain your current pattern of being able to discover and resolve for all fields that supported this interface without having to update each controller's code, but a little more of the implementation is shifted into the methods that each resolvable API type must satisfy.

pkg/resource/reference_resolver.go
negz: > We'd follow the network's provider ref to its provider resource and thus the project used by that provider.

It's worth noting that for this case whatever is doing the resolving would need a `client.Reader` in order to get and inspect the providerRef similar to the example I posted at https://github.com/crossplaneio/crossplane-runtime/pull/46#discussion_r334722588. I didn't quite grok how automatic the current implementation is (in that the resolvers in the current implementation never actually talk to Kubernetes) when writing this comment initially.

My preference is still for the simpler API - i.e. `name` only. In the case of the subnetwork->network relationship it would be redundant to require subnet authors to specify which provider the network was in given that:

1. We can discover that.
1. There's only one valid provider - the one the network references.

pkg/resource/reference_resolver.go
negz: > Another option would be to add a key to the resource:"attributereferencer kind=vpc.network.aws.crossplane.io/v1alpha2" tag

This is kind of intriguing, but I think it falls apart for resources like subnet that need to get multiple other resources (network, provider) to resolve a field.

soorena776: I don't understand how this is resolving the problem we are discussing here:
> How could we retrieve the api object without having its GVK, or how could we pass the GVK in a succinct way?

If you are proposing to use this scheme instead of what we have already discussed:

- ```go
  func (r *NetworkReference) Resolve(ctx context.Context, c client.Reader, mg Managed) error {
   ```
  The reason that I am splitting `Resolve` methods into `Build` and `Assign` methods is intentional, and is meant to simplify the logic by breaking the work into smaller pieces. Not only the developer should build the attribute, they should also assign it to the target variable. I don't want this to be assumed entirely in the `Resolve` method.

- ```go
	// Let the explicit network win if it's already set.
	if subnet.Spec.Network != "" {
		return nil
	}
   ```
  This validation should not be the business of resolving references. Even if we decided it should be, it should not be left to the developers to opt in or out, it should be enforced for all `AttributeReferencer`types.

- ```go
	n := types.NamespacedName{Namespace: mg.GetNamespace(), Name: r.Name}
	network := &v1alpha2.Network{}
	if err := c.Get(ctx, n, network); err != nil {
		return errors.Wrap(err, "could not get network")
	}
   ```
  I have been trying to factor out the common parts out of these individual interface implementations. For instance using a `getterFn` instead of passing a `client.Reader` is an effort for this goal.

soorena776: I think the question here is whether we should somehow pass the GKV (I suggested two options of an additional interface method, or a tag key), or just forget about the `Blocked` status a resource and always show an error if resolution for any reason is not successful

soorena776: > This is kind of intriguing, but I think it falls apart for resources like subnet that need to get multiple other resources (network, provider) to resolve a field.

Didn't we decide that we are only referring to a single resource? I thought a subnetwork could get both `network` and `project` from the network.

negz: > I thought a subnetwork could get both network and project from the network.

It can, but doing so requires a Kubernetes client. See my comment at https://github.com/crossplaneio/crossplane-runtime/pull/46#discussion_r334724772.

negz: > How could we retrieve the api object without having its GVK

As far as I am aware this is impossible.

>  or how could we pass the GVK in a succinct way?

It's not directly solving that issue, but it is avoiding it by avoiding having the `*ReferenceResolver` need to know about the GVK in the first place, by moving the client logic into the resolver itself.

Part of the rationale here is that I think we'll need to do something like this anyway for cases like subnetwork -> network. As demonstrated by my snippet above the project ID is a transitive property of the network - you need to use a Kubernetes client to get the network's provider, then get the provider's project ID.

I understand that this could potentially be done automatically under your original proposal in which this type looks like:

```
spec:
  networkRef:
    networkName: foo
    providerName: bar
```

However this feels like a poor user experience to me given that:

1. There's only ever one valid value for `providerName` - the provider the referenced `network` references in its `providerRef`.
2. It strays from API conventions.

I'm totally on board for minimising the amount of logic we need to implement in each API type, but if we can't find a solid abstraction I don't think it's too bad to have API type implementors need to provide a little more logic in order to resolve their types.

soorena776: I think we are talking about two different things here.

- _Interface methods implementation:_
  I am fully onboard with exposing the type of the referenced object in the interface methods (see https://github.com/crossplaneio/crossplane-runtime/pull/46#discussion_r333297962). Your version of the single method `ReferenceResolver` is slightly different than the signature I have shown in the linked comment, but I'm not opposed to passing `client.Client` instead of `LocalObjectGetter` to be able to retrieve other objects (like `providerRef`).

- _Determining the state of the referred objects in `validateReferencesAreReady`_
  Beside the need for reference type in the interface methods, we also need to determine whether the `AttributeReferencer` fields that we have extracted by reflection, are in the `Ready` state, and in order to do this we need to retrieve them and as you pointed out, we need the GVK of the referenced objects.

I'm happy to jump into a meeting to address the confusion.

negz: > Interface methods implementation:

Got it. I'm happy if we land on something closer to your proposal in https://github.com/crossplaneio/crossplane-runtime/pull/46#discussion_r333297962. I can see the value in distinct build and assign stages, and have no preference as to whether we pass `client.Reader` or `LocalObjectGetter` as long as we can handle the transitive property case (e.g. the network -> project ID case).

> Determining the state of the referred objects in validateReferencesAreReady

If I follow the code correctly this currently works by:

1. Finding all referencers via their struct tag and method set.
1. Passing them to `validateReferencesAreReady`, which gets each referenced object and returns `newResourceNotReadyErr` if any are not `Ready`. This error is passed back up to the managed resource reconciler and reflected in the resources status condition.
1. Iterating through the referencers, calling their build and assign methods.

I think we could get similar functionality with something like the interface I proposed by:

1. Finding all referencers via their struct tag and method set.
1. Calling their `Resolve` methods.
1. Allowing the `Resolve` method to inspect the readiness of the dependent resource and return `newResourceNotReadyErr` where appropriate. This error would be passed back up to the managed resource reconciler.

Under such a pattern we'd still be checking for the readiness of our dependent objects and reflecting this via a status condition (rather than just proceeding and potentially returning an error), but it again shifts more of the implementation detail into a contract placed on the implementors of the `Resolve` methods.

Does that make sense?

soorena776: @negz thanks for the clarification, your point is clear now.

My problem with this approach is that _all_ interface implementations have to deal with this error handling scenario, and implement this in the interface methods. This goes against the `managed_reconciler` principal, where all the common code is collected in one place. It also contradicts with your goals of code generation, as common pattern is spread across type definitions.

Additionally, this would also require one to always *remember* this error validation, and write unit-tests for every implementation of this interface, as it won't be possible to put any enforcements. 

I still think passing the referenced type by a tag or another interface method (like `ReferenceKind` suggested above) benefits us here. The other alternative I think would be to get rid of determining `Ready` state for each api type altogether, and directly call its interface resolve methods.

soorena776: @negz 
Another workaround would be to add another method to the interface, that actually determines the readiness of the object or objects in question. Something like:
```go
type AttributeReferencer interface {

        // ValidateReady validates if the referenced object is ready.
        ValidateReady(client.Client) error

	// Build retrieves and builds the referenced attribute
	Build(getter LocalObjectGetter) (value string, err error)

	// Assign accepts a resource object, and assigns the given value to the corresponding property
	Assign(mg Managed, value string) error
}
```

`ValidateReady` makes sure each api type implements the check for readiness, and separates the readiness concern from resolution concern.

negz: > This goes against the managed_reconciler principal, where all the common code is collected in one place.

I don't think it's as black and white as this makes it seem. Our typical pattern is to provide a scaffold (the managed reconciler) that manipulates an interface (the external connector and client). We frequently put constraints or expectations on how authors must implement those interfaces, rather than prioritising the abstraction of as much common code as possible out into the scaffold.

The more of the work you move into the scaffold that is shared between types, the less room there is for variance inside the interface the scaffold manipulates. This is a tradeoff; more room for variance means more flexibility, but it also means more work that must be done by the implementer, and thus more room for them to make mistakes.

> I still think passing the referenced type by a tag or another interface method (like ReferenceKind suggested above) benefits us here.

Perhaps it would help me get behind this idea if I understood how it would work for the GCP subnetwork edge case we've been using to test these proposals.

soorena776: > The more of the work you move into the scaffold that is shared between types, the less room there is for variance inside the interface the scaffold manipulates. This is a tradeoff;...
I agree, there is a trade-off here.

My other concern is the error validation enforcement. I think https://github.com/crossplaneio/crossplane-runtime/pull/46#discussion_r335140028 addresses the versatility between types, and yet enforces the error validation. How do you think about this?

pkg/resource/reference_resolver.go
soorena776: I agree. I'm planning to only use a name, by embedding a `LocalObjectReference` in the type

```go
type VPCIDResolver struct {
	*LocalObjectReference
}
```

This would resolve in a yaml like
```yaml
spec:
  ...
  vpcIdResolver:
    name: <vpcname>
  ...
```

apis/core/v1alpha1/condition.go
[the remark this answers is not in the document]
soorena776: I can see your point, and I agree that if we use `TypeReady`, `ReferencesResolved` status can be deduced by,  say `Ready`. However having a separate condition enables us to skip running `ResolveReferences` in each reconcile step, when `ReferencesResolved` status is overwritten by a later `ReconcileError` status.

pkg/resource/reference_resolver.go
soorena776: > Perhaps it would help me get behind this idea if I understood how it would work for the GCP subnetwork edge case we've been using to test these proposals.

For `Subnetwork`, we can skip checking the status of `provider` since it is not a managed resource. Hence, just use the single `LocalObjectReference` approach with a `resource:"attributereferencer kind=vpc.network.aws.crossplane.io/v1alpha2"` tag, and only validate the readiness of the `vpc`.

apis/core/v1alpha1/condition.go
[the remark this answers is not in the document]
soorena776: switched back to using `range`, per your other comment to avoid pointer conditions.

pkg/resource/reference_resolver.go
negz: Let me see if I understand correctly. For the GCP subnetwork:

* `ValidateReady` would use the supplied `client.Client` to validate that the GCP network was in state `Ready`, returning an error if it was not. It would not validate the Provider (which doesn't have a concept of readiness). 

* `Build` would use the supplied `LocalObjectGetter` to get the `Network` and its `Provider`, then return a string formed from the two, e.g. `/projects/id/global/networks/name`.

* `Assign` would set the `network` field of the supplied `Managed` to the value returned by `Build`.

Am I interpreting this correctly? If so it seems like a good direction. My only comments would be:

* Do we need the `kind` struct tag in this design? It's not immediately obvious to me why.
* There's a lot of similarity between `LocalObjectGetter` and `client.Client` (or `client.Reader`). Could `ValidateReady` and `Build` take the same client-ish argument?

soorena776: Yes, this is exactly what I meant!

> Do we need the kind struct tag in this design? It's not immediately obvious to me why.

No we don't. Sorry for the confusion, the following comment meant for the case we are going the path using tags.

> There's a lot of similarity between LocalObjectGetter and client.Client (or client.Reader). Could ValidateReady and Build take the same client-ish argument?

Yes, I was exactly thinking the same, both signatures need to be equal.

The rational behind passing `LocalObjectGetter` is two things:
- validate only local references are being made
- hide details like `namespace` and `context.Context`

However, looks like in the new design we do need to read objects possibly from other namespaces (`provider` in `Subnetwork` case). To solve this problem, I am going to pass a `client.Reader` and a `context.Context`, but in reflection step require that _all_  fields in the implementer type are `LocalObjectReference`.

negz: This sounds good to me! Thanks for bearing with me on this design.

muvaf: > * `ValidateReady` would use the supplied `client.Client` to validate that the GCP network was in state `Ready`, returning an error if it was not. It would not validate the Provider (which doesn't have a concept of readiness). 
> * `Build` would use the supplied `LocalObjectGetter` to get the `Network` and its `Provider`, then return a string formed from the two, e.g. `/projects/id/global/networks/name`.
> * `Assign` would set the `network` field of the supplied `Managed` to the value returned by `Build`.

If I have 5 reference fields, I'd have to implement 15 different methods, all very similar to each other. My two cents, I feel like one `Resolve` method to do all this for one field is simpler, which would result in 5 methods. Maybe, for simple fields that gets a value from the referred resource and assigns it to another field in the referrer, we might be able to have a default implementation like `SimpleFieldResolver` that developer can just opt-in to use for specific fields. Not 100% sure if that's technically feasible though.

soorena776: @muvaf The subject of why breaking down the `Resolve` method to smaller ones, came up earlier in this thread https://github.com/crossplaneio/crossplane-runtime/pull/46#discussion_r335146479
> breaking it down to smaller methods so the user is enforced to roughly follow the pattern of `{ValidateReady,Build,Assign}`

I understand that it implementing the `AttributeReferencer` interface would be noisy, but I think this scaffolding is easier to enforce at the time of development, and more useful in the managed resource (things like determining the condition state of the resolution).


As for creating `SimpleFieldResolver`, I agree. One example would be `VPCIdResolver` in AWS types that would be roughly the same (except the assign implementation), that I am considering to use, so all types that need to refer to a vpcId use that type and avoid defining their own.

pkg/resource/managed_reconciler.go
```diff
@@ -417,6 +421,23 @@ func (r *ManagedReconciler) Reconcile(req reconcile.Request) (reconcile.Result,
 	}
 
 	if !observation.ResourceExists {
+		if !IsConditionTrue(managed.GetCondition(v1alpha1.TypeReferencesResolved)) {
+			if err := r.managed.ResolveReferences(ctx, managed); err != nil {
+				// update the status according to the type of the err
+				switch err.(type) {
+				case NotReadyError:
```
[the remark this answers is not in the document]
soorena776: Good to know. Are you suggesting to use a different name?

apis/core/v1alpha1/condition.go
```diff
@@ -53,6 +56,13 @@ const (
 	ReasonReconcileError   ConditionReason = "Encountered an error during managed resource reconciliation"
 )
 
+// Reason references for a resource are or are not resolved
+const (
+	ReasonReferenceResolveSuccess  ConditionReason = "Successfully resolved managed resource references to other resources"
```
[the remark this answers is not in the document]
soorena776: @muvaf thanks for the review!

I think my main goal here was to make the cross referencing as _managed_ as possible, by laying out the overall required steps, and detecting the references automatically, rather than leaving it to the developer to write the logic and then just simply execute it by putting it in `SetupWithManager `. Taking this slightly more managed approach, has the following advantages:

- More readability, and visibility through using struct field tags
- Enforcing how resolution steps should look like
- Early panic if a referencer is not implemented correctly

That said, I agree that there is a trade-off here. It would be possible to go the `Initializer` route and avoid the reflection logic, but then the mentioned goodies won't be there. Given that I am already invested in this implementation, I prefer keeping the existing approach.

soorena776:
@muvaf thanks for the review!

I think my main goal here was to make the cross referencing as _managed_ as possible, by laying out the overall required steps, and detecting the references automatically, rather than leaving it to the developer to write the logic and then just simply execute it by putting it in `SetupWithManager `. Taking this slightly more managed approach, has the following advantages:

- More readability, and visibility through using struct field tags
- Enforcing how resolution steps should look like
- Early panic if a referencer is not implemented correctly

That said, I agree that there is a trade-off here. It would be possible to go the `Initializer` route and avoid the reflection logic, but then the mentioned goodies won't be there. Given that I am already invested in this implementation, I prefer keeping the existing approach.

pkg/resource/managed_reconciler.go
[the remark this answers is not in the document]
muvaf: >When one commands ResolveReferences the expectation is that the command is either executed successfully or has some errors

I think this is right. Personally, I'd go for `GetUnresolvedReferences` since IMO we don't actually resolve but wait for the resources to be ready. But that's highly opinionated, I can see your case fitting as well.

> I think the rational behind using an error is similar to Get case in most cloud provider go APIs

Actually, I'd compare `Find` or `List` instead of `Get`. In `Get`, you assume the resource exists, so, getting non-existent resource is a problem. But it should be fine to return empty array to `List` or `Find`.

> I like the idea of returning an array of unresolved references as the error message though. Do you think that would be useful?

Yes, I'd love to know that detail if my resource gets stuck at that stage for a long time for some reason.

pkg/resource/managed_reconciler.go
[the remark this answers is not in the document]
muvaf: It's actually because of too much similarity with `error` type. If `NotReadyError` interface has a method other than `Error() string`, that'd be sufficiently different I think. It'd still satisfy `Error` interface and others wouldn't satisfy `NotReadyError` since they don't know about that other method; maybe a method to get list of unresolved dependencies? https://github.com/crossplaneio/crossplane-runtime/pull/46#discussion_r335495121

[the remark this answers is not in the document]
soorena776: @muvaf I have changed the `NotReadyError` in the later commits, let me know what you think

pkg/resource/managed_reconciler.go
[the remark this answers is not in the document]
soorena776: I have change the `NotReadyError`, so that it accepts a list of items as an input, and uses that in `Error` function.
https://github.com/crossplaneio/crossplane-runtime/pull/46/files#diff-45d446315c96683e664c463ff0674883R59-R61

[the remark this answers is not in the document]
soorena776: > Yes, I'd love to know that detail if my resource gets stuck at that stage for a long time for some reason.

@muvaf I changed the code to return a detailed status for the references. See the last commit

pkg/resource/reference_resolver.go
negz: Per discussion this morning I'd prefer us to remove this. It may be a good direction, but it would apply to all the places we execute implementations of interfaces that are supplied in stacks so I think we should defer for now and add this in general later if we think it's useful.

pkg/resource/claim_reconciler.go
```diff
@@ -370,6 +372,21 @@ func (r *ClaimReconciler) Reconcile(req reconcile.Request) (reconcile.Result, er
 			return reconcile.Result{RequeueAfter: aShortWait}, errors.Wrap(r.client.Status().Update(ctx, claim), errUpdateClaimStatus)
 		}
 
+		if !IsConditionTrue(claim.GetCondition(v1alpha1.TypeReferencesResolved)) {
```
hasheddan: Just wanted to make a note here because I think we discussed removing this since we do it in the managed reconciler.

pkg/resource/claim_reconciler.go
```diff
@@ -370,6 +372,21 @@ func (r *ClaimReconciler) Reconcile(req reconcile.Request) (reconcile.Result, er
...
+				}
+
+				claim.SetConditions(condition)
+				return reconcile.Result{RequeueAfter: aShortWait}, errors.Wrap(r.client.Status().Update(ctx, claim), errUpdateClaimStatus)
+			}
+
+			// Add ReferenceResolutionSuccess to claim conditions
+			claim.SetConditions(v1alpha1.ReferenceResolutionSuccess())
+		}
+
```
negz: Per discussion this morning I'd like to avoid resolving resource class references if possible. I believe it should be unnecessary, as the managed resource reconciler will take over once the claim reconciler has dynamically provisioned a managed resource.

pkg/resource/managed_reconciler.go
```diff
@@ -360,6 +372,21 @@ func (r *ManagedReconciler) Reconcile(req reconcile.Request) (reconcile.Result,
 		return reconcile.Result{RequeueAfter: r.shortWait}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
 	}
 
+	if !IsConditionTrue(managed.GetCondition(v1alpha1.TypeReferencesResolved)) {
```
negz: Per our discussion this morning one down side of gating reference resolution on a condition being set is that we'll only ever process the resolution once. In the current implementation if someone edits a managed resource and changes a reference that corresponds to a mutable field, we'll ignore it.

I was thinking we could avoid this by updating the `AttributeReferencer` implementations to be no-ops if the fields they would assign were already set. This would avoid the need to check the `TypeReferencesResolved` condition, but would have the same issue in that it wouldn't support updating references after creation time.

Perhaps we should just run reference resolution every time we reconcile the managed resource? Definitely happy to defer this to a future PR in the interest of getting this iteration merged.

pkg/resource/reference_resolver.go
```diff
@@ -0,0 +1,282 @@
...
+
+	if err := newReferenceAccessErr(allStatuses); err != nil {
+		return err
+	}
+
+	// build and assign the attributes
+	for _, referencer := range referencers {
+		val, err := referencer.Build(ctx, res, r.client)
+		if err != nil {
+			return errors.WithMessage(err, errBuildAttribute)
```
negz: https://godoc.org/github.com/pkg/errors#Wrap

Nitpick - the prevalent convention is to use `Wrap` rather than `WithMessage`. It seems they're identical except that `Wrap` includes a stack trace and `WithMessage` does not.

pkg/resource/resource.go
```diff
@@ -138,3 +138,8 @@ func IsBindable(b Bindable) bool {
 func IsBound(b Bindable) bool {
 	return b.GetBindingPhase() == v1alpha1.BindingPhaseBound
 }
+
+// IsConditionTrue returns if condition status is true
+func IsConditionTrue(c v1alpha1.Condition) bool {
+	return c.Status == corev1.ConditionTrue
+}
```
negz: We should add a test for this. Here's one I prepared earlier:

```go
func TestIsConditionTrue(t *testing.T) {
	cases := map[string]struct {
		c    v1alpha1.Condition
		want bool
	}{
		"IsTrue": {
			c:    v1alpha1.Condition{Status: corev1.ConditionTrue},
			want: true,
		},
		"IsFalse": {
			c:    v1alpha1.Condition{Status: corev1.ConditionFalse},
			want: false,
		},
		"IsUnknown": {
			c:    v1alpha1.Condition{Status: corev1.ConditionUnknown},
			want: false,
		},
		"IsUnset": {
			c:    v1alpha1.Condition{},
			want: false,
		},
	}

	for name, tc := range cases {
		t.Run(name, func(t *testing.T) {
			got := IsConditionTrue(tc.c)
			if diff := cmp.Diff(tc.want, got); diff != "" {
				t.Errorf("IsConditionTrue(...): -want, +got:\n%s", diff)
			}
		})
	}
}
```

apis/core/v1alpha1/condition.go
```diff
@@ -113,6 +122,25 @@ func NewConditionedStatus(c ...Condition) *ConditionedStatus {
...
+// otherwise returns nil
+func (s *ConditionedStatus) GetCondition(ct ConditionType) Condition {
+	for _, c := range s.Conditions {
+		if c.Type == ct {
+			return c
+		}
+	}
+
+	return Condition{Type: ct, Status: corev1.ConditionUnknown}
+}
```
negz: We should add a test for this. Here's one I prepared earlier:

```go
func TestGetCondition(t *testing.T) {
	cases := map[string]struct {
		cs   *ConditionedStatus
		t    ConditionType
		want Condition
	}{
		"ConditionExists": {
			cs:   NewConditionedStatus(Available()),
			t:    TypeReady,
			want: Available(),
		},
		"ConditionDoesNotExist": {
			cs: NewConditionedStatus(Available()),
			t:  TypeSynced,
			want: Condition{
				Type:   TypeSynced,
				Status: corev1.ConditionUnknown,
			},
		},
	}

	for name, tc := range cases {
		t.Run(name, func(t *testing.T) {
			got := tc.cs.GetCondition(tc.t)
			if diff := cmp.Diff(tc.want, got); diff != "" {
				t.Errorf("tc.cs.GetConditions(...): -want, +got:\n%s", diff)
			}
		})
	}
}
```

pkg/resource/reference_resolver.go
```diff
@@ -0,0 +1,282 @@
...
+func IsReferencesAccessError(err error) bool {
+	_, result := err.(*referencesAccessErr)
+	return result
+}
+
+// CanReference is a type that is used as ReferenceResolver input
+type CanReference interface {
+	runtime.Object
+	metav1.Object
+}
```
negz: I mentioned this morning that I'd prefer we return to using `Managed` if we don't need to resolve references for resource classes. This was because I was under the mistaken impression that `CanReference` was a union of the `Managed` and `Class` interfaces. I see that it is in fact a much smaller interface, which I like. I think this interface makes sense even if we get rid of the class-level reference resolution.

pkg/resource/reference_resolver.go
```diff
@@ -0,0 +1,282 @@
...
+
+	return nil
+}
+
+func (r *referencesAccessErr) Error() string {
+	return fmt.Sprintf("%s", r.statuses)
+}
+
+// IsReferencesAccessError returns true if the given error is of type referencesAccessErr
+func IsReferencesAccessError(err error) bool {
```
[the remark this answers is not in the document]
soorena776: Good catch! Initially the plan was to use this in `ValidateReady` methods, and that's why it was exported. Making it private.

pkg/resource/managed_reconciler.go
soorena776: I agree with the direction, but I think this should be deferred to another pr. I opened https://github.com/crossplaneio/crossplane-runtime/issues/53 to track this.

pkg/resource/reference_resolver.go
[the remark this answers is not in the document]
soorena776: Actually, we are going to use this in custom reconciler functions, so this needs to be exported. I update the documentation

PR #47, opened 2019-10-10
Add ZapLogger with option to disable stacktraces

turkenh:
Signed-off-by: Hasan Turken <turkenh@gmail.com>

This PR is to show how stacktraces could be disabled with current revision of controller runtime. If we update controller runtime, we need to implement like in https://github.com/crossplaneio/stack-gcp/pull/49

Related to: Related https://github.com/crossplaneio/crossplane/issues/529
### Checklist

I have:
- [ ] Run `make reviewable` to ensure this PR is ready for review.
- [ ] Ensured this PR contains a neat, self documenting set of commits.

### DIFF
--- a/pkg/logging/logging.go
+++ b/pkg/logging/logging.go
@@ -19,7 +19,12 @@ limitations under the License.
 package logging
 
 import (
+	"os"
+
 	"github.com/go-logr/logr"
+	"github.com/go-logr/zapr"
+	"go.uber.org/zap"
+	runtimezap "sigs.k8s.io/controller-runtime/pkg/log/zap"
 	runtimelog "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/logging/logging.go
```diff
@@ -34,8 +39,27 @@ var (
...
+	ZapLogger = zapLogger
 )
 
+// ZapLogger is a Logger implementation from controller runtime.
+// If development is true, a Zap development config will be used
+// (stacktraces on warnings, no sampling), otherwise a Zap production
+// config will be used (stacktraces on errors, sampling).
+// If disableStacktrace is true, stacktraces enabled on fatals
+// independent of config.
+func zapLogger(development bool, disableStacktrace bool) logr.Logger {
```
jbw976: as discussed in https://github.com/crossplaneio/stack-gcp/pull/49, I like the idea of having a common reusable way to get a logger that only does stack traces for fatal level.  Can we continue pursuing that on this PR?  If we finish and merge this one, then all consumer repos (main crossplane, GCP/AWS/Azure stacks) can just call this one single implementation.

pkg/logging/logging.go
```diff
@@ -34,8 +39,27 @@ var (
...
+	zl := runtimezap.RawLoggerTo(os.Stderr, development)
+
+	if disableStacktrace {
```
jbw976: do you think this should use the new controller-runtime `v0.2.2` way of initializing the logger? It is probably not worth updating the entire controller-runtime to just get the new logger initialization, but it could also be nice to get a new controller-runtime update anyways.

What is your opinion @turkenh?

turkenh: As you noted, updating controller-runtime is not necessary for this specific issue but I agree that it make sense to bump controller-runtime to `v0.2.2` which is latest `v0.2`, currently it is on `v0.2.0-beta.5` which does not sound that stable :)
Still I'll go and ask on slack dev channel to make sure everyone is ok with this.

negz:
I'm afraid I don't agree with @jbw976 regarding having crossplane-runtime provide helper code to produce correctly loggers. I'd rather keep any such code (which I imagine should be fairly succinct?) limited to `main.go`. As of  https://github.com/crossplaneio/crossplane-runtime/pull/108 crossplane-runtime is aware only of the interfaces it should use for logging; not what concrete implementation should be used.

How should we proceed with this PR? I personally don't feel anything is currently broken, so my vote would be to close it.

PR #48, opened 2019-10-10
Introduce label selector based class scheduling

negz:

This enables label based class matching per #926. Changes will also be required to crossplane, and all stacks.

### Checklist

I have:


apis/core/v1alpha1/resource.go
```diff
@@ -82,13 +117,12 @@ type ResourceClaimStatus struct {
 
 // A ResourceSpec defines the desired state of a managed resource.
 type ResourceSpec struct {
-	// WriteConnectionSecretToReference specifies the name of a Secret, in the
-	// same namespace as this managed resource, to which any connection details
+	// WriteConnectionSecretToReference specifies the namespace and name of a Secret  to which any connection details
```
hasheddan: nit: I think this comment may have escaped being wrapped

pkg/resource/claim_defaulting_reconciler.go
```diff
@@ -0,0 +1,122 @@
...
+		// succeeded. It's possible another controller can successfully set a
+		// class even though we can't, so it would be confusing to mark this
+		// claim as failing to be reconciled. Instead we return an error - we'll
+		// be requeued but abort immediately if the claim was defaulted.
+		return reconcile.Result{}, errors.Wrap(err, errListClasses)
+	}
+
+	defaults := []unstructured.Unstructured{}
+	for _, c := range classes.Items {
+		if c.GetAnnotations()[v1alpha1.AnnotationDefaultClassKey] == v1alpha1.AnnotationDefaultClassValue {
```
hasheddan: I know specifying default as an annotation rather than a label was an intentional decision. Is the motivation behind that because that we don't want conflicts with labels used for scheduling?

pkg/resource/claim_scheduling_reconciler.go
```diff
@@ -0,0 +1,157 @@
...
+		return reconcile.Result{}, errors.Wrap(err, errListClasses)
+	}
+
+	if len(classes.Items) == 0 {
+		// None of our classes matched the class selector.
+		// There's nothing for us to do.
+		return reconcile.Result{Requeue: false}, nil
+	}
+
+	random := rand.New(rand.NewSource(time.Now().UnixNano()))
```
hasheddan: I like that we don't just use the first one we find, reminds me of [this property](https://nathanleclaire.com/blog/2014/04/27/a-surprising-feature-of-golang-that-colored-me-impressed/) of go.

pkg/resource/configurator.go
```diff
@@ -52,16 +52,21 @@ func NewObjectMetaConfigurator(t runtime.ObjectTyper) *ObjectMetaConfigurator {
...
 	mg.SetGenerateName(fmt.Sprintf("%s-%s-", cm.GetNamespace(), cm.GetName()))
 	if meta.GetExternalName(cm) != "" {
 		meta.SetExternalName(mg, meta.GetExternalName(cm))
 	}
-	// TODO(negz): Don't set this potentially cross-namespace owner reference.
-	// We probably want to use the resource's reclaim policy, not Kubernetes
-	// garbage collection, to determine whether to delete a managed resource
-	// when its claim is deleted per https://github.com/crossplaneio/crossplane/issues/550
+
+	// TODO(negz): Avoid setting this owner reference? Kubernetes specifies that
```
hasheddan: Do you know what happens if we try and set this owner reference?

pkg/resource/claim_defaulting_reconciler.go
negz: Exactly. Also, it's closer to what storage classes do. I think we decided on a label rather than an annotation for our contemporary implementation because you can't filter a list on an annotation. In retrospect I figure it's not too bad to just list all the classes and filter them manually.

https://kubernetes.io/docs/tasks/administer-cluster/change-default-storage-class/

pkg/resource/configurator.go
negz: I haven't looked at this for some time, but as far as I can tell it works. Currently we almost always dynamically provision a managed resource in a separate namespace from the resource claim, yet deleting the resource claim automatically deletes the managed resource (via owner reference based garbage collection).

What's interesting is that I have no idea _how_ this works - I'm not sure how the garbage collector figures out what to delete without knowing the owner's namespace.

hasheddan: I am going to dive into this a little bit more, but I don't think my investigation needs to block here :)

pkg/resource/claim_scheduling_reconciler.go
negz: Do you think there would be much value in doing this in the defaulting controller too?

I originally had that randomly picking one default when multiple classes were annotated as default (pretty much the same logic as the scheduling controller), but then I figured it was acceptable (and simpler) to just pick the first class given that the existence of multiple default classes is a misconfiguration.

That said, it would not be that hard to pick one randomly (just copy the logic from the scheduling controller), and doing so could conceivably be useful to some users (i.e. if you wanted to randomly spread unopinionated claims across N different "default" resource classes)/

hasheddan: I like the idea of picking a random one because it removes the ability for a user to either:

1. Have multiple defaults configured and not fix it because they know the one they want is going to get picked (not sure if the `List()` returns predictable order?).
2. Have multiple defaults configured and not be aware of it because the consistently get the same one.

TLDR: I would support doing this in the defaulting controller :)

negz: I think I agree, but I might just add a TODO for now and address this in a future PR.

apis/core/v1alpha1/resource.go
```diff
@@ -124,10 +158,15 @@ type ResourceStatus struct {
...
-// resource class.
-type NonPortableClassSpecTemplate struct {
+// A ClassSpecTemplate defines a template that will be used to create the
+// specifications of managed resources dynamically provisioned using a resource
+// class.
+type ClassSpecTemplate struct {
+	// WriteConnectionSecretsToNamespace specifies the namespace in which the
+	// connection secrets of managed resources dynamically provisioned using
+	// this claim will be created.
+	WriteConnectionSecretsToNamespace string `json:"writeConnectionSecretsToNamespace"`
```
hasheddan: Is this the remedy to the problem proposed in this slack message: https://crossplane.slack.com/archives/CKXQHM7U3/p1571426744220400 ? If so, I like it! I appreciate how it allows the user to choose the secret's namespace, even if we may recommend that they usually choose the same namespace as the `Provider` secret.

pkg/resource/claim_defaulting_reconciler.go
```diff
@@ -0,0 +1,122 @@
...
+		}
+	}
+
+	if len(defaults) == 0 {
+		// None of our classes are annotated as the default.
+		// There's nothing for us to do.
+		return reconcile.Result{Requeue: false}, nil
+	}
+
+	claim.SetClassReference(meta.ReferenceTo(&defaults[0], schema.GroupVersionKind(r.classKind)))
```
hasheddan: Can we add that `TODO` you mentioned here? Alternatively, I believe this would be fixed just by mimicking the claim scheduling reconciler randomness:
```go
random := rand.New(rand.NewSource(time.Now().UnixNano()))
selected := defaults[random.Intn(len(defaults))]
claim.SetClassReference(meta.ReferenceTo(&selected, schema.GroupVersionKind(r.classKind)))
```

apis/core/v1alpha1/resource.go
negz: Yeah, this was option two proposed in that discussion. I went with this rather than option 3 because it seemed simpler, more flexible, and more explicit/less magic. I agree that in practice we probably just want to recommend all providers and classes use `crossplane-system` for their secrets.

pkg/resource/claim_defaulting_reconciler.go
negz: Okay fiiine I just implemented this.

negz:
Here's an example of simple resource class selection in action:

```console
$ kubectl get redisclass -l classy="true"
NAME                     PROVIDER-REF   RECLAIM-POLICY   AGE
azure-redis-labelled-a   example        Delete           51m
azure-redis-labelled-b   example        Delete           51m
azure-redis-labelled-c   example        Delete           51m
azure-redis-labelled-d   example        Delete           51m
azure-redis-labelled-e   example        Delete           51m

$ kubectl get cloudmemorystoreinstanceclass -l classy="true"
NAME                   PROVIDER-REF   RECLAIM-POLICY   AGE
gcp-redis-labelled-a   example        Delete           50m
gcp-redis-labelled-b   example        Delete           49m
gcp-redis-labelled-c   example        Delete           49m
gcp-redis-labelled-d   example        Delete           49m
gcp-redis-labelled-e   example        Delete           49m

$ kubectl get rediscluster
NAME          STATUS   CLASS-KIND                      CLASS-NAME               RESOURCE-KIND              RESOURCE-NAME               AGE
app-redis-0            RedisClass                      azure-redis-labelled-b   Redis                      default-app-redis-0-8rhkr   4m31s
app-redis-1            RedisClass                      azure-redis-labelled-b   Redis                      default-app-redis-1-tbn7g   4m11s
app-redis-2            CloudMemorystoreInstanceClass   gcp-redis-labelled-a     CloudMemorystoreInstance   default-app-redis-2-mrqpm   3m52s
app-redis-3            RedisClass                      azure-redis-labelled-c   Redis                      default-app-redis-3-glr4s   3m36s
app-redis-4            RedisClass                      azure-redis-labelled-b   Redis                      default-app-redis-4-vdzc9   3m27s
app-redis-5            CloudMemorystoreInstanceClass   gcp-redis-labelled-b     CloudMemorystoreInstance   default-app-redis-5-2qt5r   3m20s
app-redis-6            RedisClass                      azure-redis-labelled-a   Redis                      default-app-redis-6-t7g59   3m11s
app-redis-7            CloudMemorystoreInstanceClass   gcp-redis-labelled-a     CloudMemorystoreInstance   default-app-redis-7-cwmpl   2m25s
app-redis-8            CloudMemorystoreInstanceClass   gcp-redis-labelled-b     CloudMemorystoreInstance   default-app-redis-8-9dcm4   2m25s
app-redis-9            CloudMemorystoreInstanceClass   gcp-redis-labelled-b     CloudMemorystoreInstance   default-app-redis-9-fdxqf   2m25s
```

The claims are all variants of:
```yaml
---
apiVersion: cache.crossplane.io/v1alpha1
kind: RedisCluster
metadata:
  name: app-redis-0
spec:
  classSelector:
    matchLabels:
      classy: "true"
  writeConnectionSecretToRef:
    name: app-redis-0
```

PR #54, opened 2019-10-22
Add port key to connection secret key constants

muvaf:

We get port information alongside endpoint from the providers but currently assume that the user can figure out the port I guess. This defines a connection secret key that we can use to represent that information.

### Checklist

I have:

### DIFF
--- a/apis/core/v1alpha1/resource.go
+++ b/apis/core/v1alpha1/resource.go
@@ -23,6 +23,8 @@ import (
 const (
 	// ResourceCredentialsSecretEndpointKey is the key inside a connection secret for the connection endpoint
 	ResourceCredentialsSecretEndpointKey = "endpoint"
+	// ResourceCredentialsSecretPortKey is the key inside a connection secret for the connection port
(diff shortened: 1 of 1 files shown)


muvaf:
GCP [CloudMemoryInstance](https://github.com/crossplaneio/stack-gcp/blob/master/pkg/clients/cloudmemorystore/cloudmemorystore.go#L123) and AWS [Elasticache](https://github.com/crossplaneio/stack-aws/blob/master/pkg/controller/cache/managed.go#L131) resources do expose port information but we don't make use of them. I think we can expect some more resources expose port for connection. Though usually it's not user-defined, I think it's helpful to expose it.

negz:
This seems reasonable to me. It's a little undefined how the existing endpoint could be used, so you could argue the endpoint should be `redis://example.org:1234` or something rather than including a separate port. I like the flexibility of having a distinct port field though, given I expect the endpoint to vary depending on how clients typically consume it (i.e. as a URL, a raw IP, a DNS name, etc).

PR #55, opened 2019-10-24
Fix static provisioning edge cases

### LINKED ISSUE #28: Static provisioning requires nonsensical managed resources
### What happened?

Changes were made to our resource claim controller watch predicates recently in https://github.com/crossplaneio/crossplane-runtime/pull/18 and https://github.com/crossplaneio/crossplane-runtime/pull/24. The latter refactored the watch predicates used by resource claim controllers with the intention of supporting both statically and dynamically provisioned managed resources. It does this by accepting resources that satisfy any of three watch predicates, applied in the following order:

1. `resource.HasManagedResourceReferenceKind` accepts resource claims with a `.spec.resourceRef` to the kind of managed resource the reconciler is concerned with.
1. `resource.HasDirectClassReferenceKind` accepts managed resources with a `.spec.classRef` to the kind of non-portable class the reconciler is concerned with.
1. `resource.HasIndirectClassReferenceKind` accepts resource claims with a `.spec.classRef` to a portable class with a `.classRef` to the kind of non-portable class the reconciler is concerned with.

In retrospect `resource.HasDirectClassReferenceKind` only makes sense in a dynamic provisioning context, because statically provisioned managed resources don't use a resource class, and thus typically don't set their `.spec.classRef`. Based on my read of the current predicate logic (I have not confirmed this) I expect our resource claim reconcilers would never notice changes to statically provisioned managed resources.

This is not too big an issue given the resource claim reconciler only watches managed resources to wait for them to become bindable, which they typically do before the resource claim is created in a static provisioning scenario. I'm guessing that's why I didn't encounter this case when I tested static provisioning per https://github.com/crossplaneio/crossplane-runtime/pull/24#issuecomment-531460570.

### How can we reproduce it?

1. Create a resource claim that references a managed resource that does not yet exist.
1. Create the referenced managed resource.

My guess is that the claim will never bind.

There are two workarounds for this:
1. Create the managed resource and wait for it its `.status.bindingPhase` to become `Unbound` before creating a resource claim that references it.
1. Set the managed resource's `.spec.classRef.kind` and `.spec.classRef.apiVersion` to its corresponding non-portable resource class kind.

crossplane-runtime version: 26a458d

negz:

Fixes #28

Quoting the above issue:

> There are three scenarios we need to make sure the claim reconciler handles:
>
> 1. Dynamic provisioning. No managed resource exists. A resource claim is created with a portable class reference and without a managed resource reference. In this case the claim reconciler should dynamically provision and then bind to a managed resource.
> 1. Resource-first static provisioning. A managed resource is created explicitly, without using a resource class. Later (or simultaneously), a resource claim that references (via its resource reference) that managed resource is created. We should skip dynamic provisioning and go straight to binding.
> 1. Claim-first static provisioning. A resource claim is created that references (via its resource reference) a managed resource that does not yet exist. We should fail to reconcile that resource claim until such time as the referenced managed resource is created (out of band) and becomes bindable, at which point we should bind to it.

While investigating I found that scenario 3 lead to a nil pointer exception, and scenario 2 was partially broken; static provisioning would only work when the statically provisioned managed resource was available for binding at the time the claim was submitted. If the resource was still being created when the claim was submitted it would never bind. I also realised we were susceptible to a nil pointer in a pathological case in which a managed resource referenced a claim that had neither a resource reference or a class reference. I believe this PR fixes all of these issues.

### Checklist

I have:

### DIFF
--- a/pkg/resource/api.go
+++ b/pkg/resource/api.go
@@ -147,11 +147,12 @@ func (a *APIManagedConnectionPropagator) PropagateConnection(ctx context.Context
 // using the status subresource; such objects should use APIManagedStatusBinder.
 type APIManagedBinder struct {
 	client client.Client
+	typer  runtime.ObjectTyper
 }
 
 // NewAPIManagedBinder returns a new APIManagedBinder.
-func NewAPIManagedBinder(c client.Client) *APIManagedBinder {
-	return &APIManagedBinder{client: c}
+func NewAPIManagedBinder(c client.Client, t runtime.ObjectTyper) *APIManagedBinder {
+	return &APIManagedBinder{client: c, typer: t}
 }
 
 // Bind the supplied resource to the supplied claim.
@@ -164,6 +165,12 @@ func (a *APIManagedBinder) Bind(ctx context.Context, cm Claim, mg Managed) error
 			return errors.Wrap(err, errUpdateClaim)
 		}
 	}
+
+	// This claim reference will already be set for dynamically provisioned
+	// managed resources, but we need to make sure it's set for statically
+	// provisioned resources too.
+	cmr := meta.ReferenceTo(cm, MustGetKind(cm, a.typer))
+	mg.SetClaimReference(cmr)
 	mg.SetBindingPhase(v1alpha1.BindingPhaseBound)
 	if err := a.client.Update(ctx, mg); err != nil {
 		return errors.Wrap(err, errUpdateManaged)
@@ -177,11 +184,12 @@ func (a *APIManagedBinder) Bind(ctx context.Context, cm Claim, mg Managed) error
 // APIManagedBinder.
 type APIManagedStatusBinder struct {
 	client client.Client
+	typer  runtime.ObjectTyper
 }
 
 // NewAPIManagedStatusBinder returns a new APIManagedStatusBinder.
-func NewAPIManagedStatusBinder(c client.Client) *APIManagedStatusBinder {
-	return &APIManagedStatusBinder{client: c}
+func NewAPIManagedStatusBinder(c client.Client, t runtime.ObjectTyper) *APIManagedStatusBinder {
+	return &APIManagedStatusBinder{client: c, typer: t}
 }
 
 // Bind the supplied resource to the supplied claim.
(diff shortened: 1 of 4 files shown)


negz:
```bash
$ make reviewable test build
14:52:24 [ .. ] go mod vendor
14:52:25 [ OK ] go mod vendor
14:52:25 [ .. ] go generate darwin_amd64
apis/core/v1alpha1/zz_generated.deepcopy.go
14:52:26 [ OK ] go generate darwin_amd64
14:52:26 [ .. ] verify dependencies have expected content
all modules verified
14:52:30 [ OK ] go modules dependencies verified
14:52:30 [ .. ] golangci-lint
14:52:38 [ OK ] golangci-lint
14:52:38 [ .. ] verify dependencies have expected content
all modules verified
14:52:40 [ OK ] go modules dependencies verified
14:52:40 [ .. ] go test unit-tests
?       github.com/crossplaneio/crossplane-runtime/pkg/logging  [no test files]
ok      github.com/crossplaneio/crossplane-runtime/pkg/meta     0.257s  coverage: 100.0% of statements
ok      github.com/crossplaneio/crossplane-runtime/pkg/resource 0.287s  coverage: 93.9% of statements
?       github.com/crossplaneio/crossplane-runtime/pkg/test     [no test files]
ok      github.com/crossplaneio/crossplane-runtime/pkg/util     0.370s  coverage: 78.6% of statements
?       github.com/crossplaneio/crossplane-runtime/apis [no test files]
ok      github.com/crossplaneio/crossplane-runtime/apis/core/v1alpha1   0.302s  coverage: 27.3% of statements
14:52:45 [ OK ] go test unit-tests
14:52:46 [ .. ] verify dependencies have expected content
all modules verified
14:52:47 [ OK ] go modules dependencies verified
14:52:47 [ .. ] golangci-lint
14:52:54 [ OK ] golangci-lint
14:52:54 [ .. ] go build linux_amd64
14:52:54 [ OK ] go build linux_amd64
```

negz:
> Have you had the chance to test it out with a real scenario?

I have. I built stack-gcp against this branch and tested that all scenarios discussed in this issue work as expected.

hasheddan:
@negz awesome! I am comfortable with this being merged :) also, we will need to bump dependencies in all stacks

PR #56, opened 2019-10-25
Publicize mocks of our interfaces

muvaf:

They are helpful for the tests we want to write for the following checks where the test object should adhere to `Managed` interface but not the actual type:
```
cr, ok := mg.(*v1alpha2.RDSInstance)
if !ok {
	return resource.ExternalObservation{}, errors.New(errNotRDSInstance)
}
```

### Checklist

I have:

### DIFF
--- a/pkg/resource/api_test.go
+++ b/pkg/resource/api_test.go
@@ -18,8 +18,9 @@ package resource
 
 import (
 	"context"
-	"github.com/crossplaneio/crossplane-runtime/pkg/resource/fake"
 	"testing"
+
+	"github.com/crossplaneio/crossplane-runtime/pkg/resource/fake"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/pkg/errors"
@@ -72,7 +73,7 @@ func TestCreate(t *testing.T) {
 				client: &test.MockClient{
 					MockCreate: test.NewMockCreateFn(errBoom),
 				},
-				typer: MockSchemeWith(&fake.MockClaim{}, &fake.MockClass{}, &fake.MockManaged{}),
+				typer: fake.MockSchemeWith(&fake.MockClaim{}, &fake.MockClass{}, &fake.MockManaged{}),
 			},
 			args: args{
 				ctx: context.Background(),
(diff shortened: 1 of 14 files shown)


negz:
This seems like a good direction to me. Does it block your work on 0.4 @muvaf? It might be nice to move it into a distinct `fake` package, so we have `fake.Managed{}` or similar.

negz:
I don't think this is a priority, but I'd still be happy to approve this PR if it was rebased and the interfaces were moved into a `fake` package. Let me know if you're still interested in moving forward with this @muvaf.

muvaf:
@negz I just ran into another case where I need these guys in the tests. Updated the PR with your suggestion, it should be ready to merge.

negz:
@muvaf Looks like there's a merge conflict.

pkg/resource/api_test.go
```diff
@@ -20,6 +20,8 @@ import (
 	"context"
 	"testing"
 
+	"github.com/crossplaneio/crossplane-runtime/pkg/resource/fake"
```
negz: Would prefer if this import wasn't floating by itself.

pkg/resource/api_test.go
```diff
@@ -71,13 +73,13 @@ func TestCreate(t *testing.T) {
 				client: &test.MockClient{
 					MockCreate: test.NewMockCreateFn(errBoom),
 				},
-				typer: MockSchemeWith(&MockClaim{}, &MockClass{}, &MockManaged{}),
+				typer: fake.MockSchemeWith(&fake.MockClaim{}, &fake.MockClass{}, &fake.MockManaged{}),
```
negz: `fake.MockFoo` feels a bit stutter-ey.  Now that these are in their own package I'd prefer they were named just `fake.Scheme`, `fake.Claim`, etc.

pkg/resource/api_test.go
negz: Looks like a few other imports have this issue too.

negz:
LGTM in general, but I'd prefer if we could remove the `Fake` prefix from the type names (since they're now in package `fake`). Could you also add a package doc comment to the new `fake` package? Take a look at the other package doc strings for examples.

muvaf:
@negz Thanks for the naming suggestion! It should be ready to merge.

Issue #63, opened 2019-10-28
Managed reconciler should set the finalizer only after resource is created


muvaf:
### What problem are you facing?

Currently, for some reason, if the steps before `Create` call fails and the reason also causes `Delete` to fail, the resource is stuck and we can't delete it. Examples:
* Insufficient permissions of the provider's credential causes even the first `Observe` call to fail, but since finalizer is already set, user cannot delete the CR.
* If references cannot be solved forever, there is no way to delete the resource since the finalizer is set already and `Delete` is never executed. #62 

This started to happen after the merge of https://github.com/crossplaneio/crossplane-runtime/pull/45/files#diff-c7df5a255a05fb18975e474f9af689a9L400 since we set the finalizer _no matter what_ on this PR.

### How could Crossplane help solve your problem?

We can set the finalizer _only_ after the `Observe` call returns that the resource exists and remove it after deletion signal is received and resource doesn't exist. But I think these calls should be native part of the managed reconciler reconcile loop instead of yet another `Initializer` or `Finalizer` objects. There has been some discussion around this https://github.com/crossplaneio/crossplane-runtime/pull/45#discussion_r333972416 but I think we should fix it for 0.5 of Crossplane.

negz:
> But I think these calls should be native part of the managed reconciler reconcile loop instead of yet another Initializer or Finalizer objects.

I can see why we might want to do this, because all the `Initializer` and `Finalizer` terminology kind of papers and over and abstracts the fact that we're adding and removing finalizers here. I implemented finalizer addition this way originally because:

1. Hypothetically a controller might not need or want to set a finalizer? In practice I can't really imagine any controller dealing with an external system that would not need to use a finalizer, though.
1. Hypothetically a controller might want to set more than one finalizer? We haven't seen this in practice so far.
1. Updating finalizers requires updating the managed resource itself, not its status. This means one more call every managed resource reconciler must make, one more thing we must mock out for every test, etc.

negz:
I'm going to play with this today to see if there's something we could do to fix this and #62 in time for possible 0.4.1 release.

negz:
> We can set the finalizer only after the Observe call returns that the resource exists

[...]

PR #64, opened 2019-10-28
Make 'Observe' method conditional on having all the references resolved

### LINKED ISSUE #62: Managed Reconciler: A resource with unresolved references never gets deleted
### What happened?

### How can we reproduce it?

soorena776:
Signed-off-by: soorena776 <javad@upbound.io>

Fixes #62 
### Checklist

I have:

### DIFF
--- a/pkg/resource/managed_reconciler_test.go
+++ b/pkg/resource/managed_reconciler_test.go
@@ -53,6 +53,13 @@ func TestManagedReconciler(t *testing.T) {
 	errBoom := errors.New("boom")
 	errNotReady := &referencesAccessErr{[]ReferenceStatus{{Name: "cool-res", Status: ReferenceNotReady}}}
 	now := metav1.Now()
+
+	setReferencesResolvedFn := func(obj runtime.Object) error {
+
+		obj.(Managed).SetConditions(v1alpha1.ReferenceResolutionSuccess())
+		return nil
+	}
+
 	testFinalizers := []string{"finalizer.crossplane.io"}
 	testConnectionDetails := ConnectionDetails{
 		"username": []byte("crossplane.io"),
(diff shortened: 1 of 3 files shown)


pkg/resource/managed_reconciler_test.go
```diff
@@ -229,576 +245,6 @@ func TestManagedReconciler(t *testing.T) {
 			},
 			want: want{result: reconcile.Result{RequeueAfter: defaultManagedShortWait}},
 		},
-		"DeletedButResourceExistsAndReclaimRetain": {
```
soorena776: fixing the unittests

muvaf:
I think the issue here is a more general one; we shouldn't set the finalizer unless we are sure the resource actually exists, either creating or in any form. See https://github.com/crossplaneio/crossplane-runtime/issues/63

soorena776:
@muvaf I saw #63 and I agree. This fix however is focused on where ReferenceResolver should be placed in the Reconcile loop, as it currently always exits the loop early if a reference is not resolved

muvaf:
> // If the references are not resolved the resource cannot exist, assuming
// that the reference fields are immutable.

I am not sure about both of the statements above hold true in our use cases.

On another note, I'm a bit hesitant to make a change on managed reconciler this late to the 0.4 release. How about moving reference resolve block to after `if meta.WasDeleted(managed)` block to solve this specific issue if it's a really pressing one?

negz:
I think we should avoid fixing this for 0.4, which we plan to ship today. My rationale is:

* Fixing it would require last minute changes to the managed resource reconciler, which is used by well over half of our controllers these days. The blast radius of unintended effects on 0.4 would be large.
* Fixing the managed resource reconciler won't fix the issue for the Kubernetes cluster, bucket, and Azure Redis controllers (assuming they are also affected), which do not use the managed resource reconciler.
* I believe it's possible to work around the issue (albeit manually) by editing it to either remove the finalizer, or remove the offending references.
* I'm not certain the change proposed by this PR is the best path forward.

I suggest we consider the bug raised by #62 together with that raised by #63 holistically and see if we can come up with a bug fix for 0.4.1.

negz:
We ended up taking a different path for this in https://github.com/crossplaneio/crossplane-runtime/pull/70, so I'm going to close this PR. I appreciate the effort coming up with a quick fix, even though we did not run with it for 0.4!

PR #65, opened 2019-10-31
Managed resource reconciler refactoring

### LINKED ISSUE #51: Make use of finalizer mechanism mandatory

Currently, finalizer add/remove mechanism is [enabled by default](https://github.com/crossplaneio/crossplane-runtime/blob/master/pkg/resource/managed_reconciler.go#L232) in all generic managed reconciler instances. However, with the introduction of https://github.com/crossplaneio/crossplane-runtime/pull/45 there will be some cases that developer doesn't want to use the `ManagedNameAsExternalName` initializer for development of resources that get their generated name from the provider. This could potentially lead to forgetting to supply `APIManagedFinalizerAdder` in the initialization chain. 

I think there is no case that we don't want to use `APIManagedFinalizerAdder` or `APIManagedFinalizerRemover`. So, we can just make them mandatory by calling directly or injecting into the supplied initialization chain manually.

### LINKED ISSUE #53: ReferenceResolver: Remove the condition on `ReferenceResolutionSuccess` to resolve references

origin: https://github.com/crossplaneio/crossplane-runtime/pull/46#discussion_r337195043

One downside of gating reference resolution on a condition being set is that we'll only ever process the resolution once. In the current implementation if someone edits a managed resource and changes a reference that corresponds to a mutable field, we'll ignore it.
Perhaps we should just run reference resolution every time we reconcile the managed resource? Definitely happy to defer this to a future PR in the interest of getting this iteration merged.

Per @negz, we could avoid this by updating the AttributeReferencer implementations to be no-ops if the fields they would assign were already set. This would avoid the need to check the TypeReferencesResolved condition, but would have the same issue in that it wouldn't support updating references after creation time.

### LINKED ISSUE #62: Managed Reconciler: A resource with unresolved references never gets deleted

### LINKED ISSUE #63: Managed reconciler should set the finalizer only after resource is created

negz:

Opening this as a draft for now, because it's fairly wide ranging and should probably be more than one PR. In the meantime everything is broken down by commit fairly cleanly.

The PR:

* Adds finalizers right before the creation block - see rationale in that commit message.
* Triggers reference resolution on every reconcile - this will require updates to referencers to ensure they're idempotent (i.e. only append the resolved value to a slice once).
* Avoids updating the managed resource if it doesn't change during the reference resolution process.
* Removes the check for the `attributereferencer` struct tag - see rationale in that commit message.
* Cleans up the managed reconciler and reference resolver tests.

Fixes #63
Fixes #62 
Fixes #53 
Fixes #51 

I've tested this all out by creating a GKE cluster and CloudSQL database using the `examples/workload` manifests.

### Checklist

I have:

### DIFF
--- a/pkg/resource/managed_reconciler.go
+++ b/pkg/resource/managed_reconciler.go
@@ -394,14 +394,6 @@ func (r *ManagedReconciler) Reconcile(req reconcile.Request) (reconcile.Result, 
 		return reconcile.Result{RequeueAfter: r.shortWait}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
 	}
 
-	if err := r.managed.Initialize(ctx, managed); err != nil {
-		// If this is the first time we encounter this issue we'll be requeued
-		// implicitly when we update our status with the new error condition.
-		// If not, we want to try again after a short wait.
-		managed.SetConditions(v1alpha1.ReconcileError(err))
-		return reconcile.Result{RequeueAfter: r.shortWait}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
-	}
-
 	observation, err := external.Observe(ctx, managed)
 	if err != nil {
 		// We'll usually hit this case if our Provider credentials are invalid
@@ -483,6 +475,16 @@ func (r *ManagedReconciler) Reconcile(req reconcile.Request) (reconcile.Result, 
 		return reconcile.Result{RequeueAfter: r.shortWait}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
 	}
 	managed.SetConditions(v1alpha1.ReferenceResolutionSuccess())
+
+	// TODO(negz): Is this really where we want to set the external name?
+	// We'll probably want to use it during observation.
+	if err := r.managed.Initialize(ctx, managed); err != nil {
+		// If this is the first time we encounter this issue we'll be requeued
+		// implicitly when we update our status with the new error condition.
+		// If not, we want to try again after a short wait.
+		managed.SetConditions(v1alpha1.ReconcileError(err))
+		return reconcile.Result{RequeueAfter: r.shortWait}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
+	}
 
 	if !observation.ResourceExists {
 		creation, err := external.Create(ctx, managed)
--- a/pkg/resource/managed_reconciler.go
+++ b/pkg/resource/managed_reconciler.go
@@ -50,32 +50,34 @@ const (
 // resource, for example usernames, passwords, endpoints, ports, etc.
 type ConnectionDetails map[string][]byte
 
-// A ManagedConnectionPublisher manages the supplied ConnectionDetails for the supplied
-// Managed resource. ManagedPublishers must handle the case in which the
-// supplied ConnectionDetails are empty.
+// A ManagedConnectionPublisher manages the supplied ConnectionDetails for the
+// supplied Managed resource. ManagedPublishers must handle the case in which
+// the supplied ConnectionDetails are empty.
 type ManagedConnectionPublisher interface {
 	PublishConnection(ctx context.Context, mg Managed, c ConnectionDetails) error
 	UnpublishConnection(ctx context.Context, mg Managed, c ConnectionDetails) error
 }
 
-// ManagedConnectionPublisherFns is the pluggable struct to produce objects with ManagedConnectionPublisher interface.
+// ManagedConnectionPublisherFns satisies the ManagedConnectionPublisher
+// interface.
 type ManagedConnectionPublisherFns struct {
 	PublishConnectionFn   func(ctx context.Context, mg Managed, c ConnectionDetails) error
 	UnpublishConnectionFn func(ctx context.Context, mg Managed, c ConnectionDetails) error
 }
 
-// PublishConnection calls plugged PublishConnectionFn.
+// PublishConnection details.
 func (fn ManagedConnectionPublisherFns) PublishConnection(ctx context.Context, mg Managed, c ConnectionDetails) error {
 	return fn.PublishConnectionFn(ctx, mg, c)
 }
 
-// UnpublishConnection calls plugged UnpublishConnectionFn.
+// UnpublishConnection details.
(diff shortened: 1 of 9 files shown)


muvaf:
I think I need some strong reasons to introduce `Annotator` chain. It seems that the reason we separate annotation process from initialization is we want to separate the identification from the life-cycle management that we achieve through k8s finalizers. But per https://github.com/crossplaneio/crossplane-runtime/issues/51 there is no need to make k8s finalizer mechanism an `Initializer` at all. So, this removes the life-cycle responsibility of `Initialize` call, meaning we don't need to call the `Initialize` after `Observe` anymore. We can keep the `Initialize` call as is and add the finalizer add/delete to the appropriate places.

Something like this;
* `Annotate` call is `Initialize` again.
* APIFinalizerAdder & APIFinalizerRemover are made internal objects.
* We call `addKubernetesFinalizer` in the place of `Initialize` call as in this PR(right before `if !observation.ResourceExists`)
* We call `removeKubernetesFinalizer` right after `Finalize` call.

pkg/resource/claim_binding_reconciler.go
```diff
@@ -138,6 +138,20 @@ type ManagedFinalizer interface {
 	Finalize(ctx context.Context, cm Managed) error
 }
 
+// A FinalizerChain chains multiple managed finalizers.
+type FinalizerChain []ManagedFinalizer
```
muvaf: Noticed it the first time in this PR but why do we call `r.managed.Finalize` in claim reconciler? Shouldn't claim reconciler just send the deletion signal and checks whether object is deleted, i.e `(NotFound)` and then calls its claim's finalizer? Pretty much like how managed reconciler treats its external resource.

pkg/resource/managed_reconciler.go
```diff
@@ -235,17 +278,16 @@ type mrManaged struct {
 	ManagedFinalizer
 	ManagedInitializer
 	ManagedReferenceResolver
+	ManagedAnnotator
 }
 
 func defaultMRManaged(m manager.Manager) mrManaged {
 	return mrManaged{
 		ManagedConnectionPublisher: NewAPISecretPublisher(m.GetClient(), m.GetScheme()),
 		ManagedFinalizer:           NewAPIManagedFinalizerRemover(m.GetClient()),
```
muvaf: It seems #51 is not fixed in this PR.

pkg/resource/managed_reconciler.go
```diff
@@ -339,7 +405,7 @@ func NewManagedReconciler(m manager.Manager, of ManagedKind, o ...ManagedReconci
 
 // Reconcile a managed resource with an external resource.
 func (r *ManagedReconciler) Reconcile(req reconcile.Request) (reconcile.Result, error) { // nolint:gocyclo
-	// NOTE(negz): This method is a little over our cyclomatic complexity goal.
+	// NOTE(negz): This method is a well over our cyclomatic complexity goal.
```
muvaf: Sad but true.

pkg/resource/managed_reconciler.go
negz: This PR does not literally make the finalizer mandatory, but it does return it to being on by default and less easy to accidentally turn off. The spirit of #51 seemed to be that it had become easy to accidentally turn off finalizers unintentionally, and I think that's no longer the case with this PR.

pkg/resource/claim_binding_reconciler.go
negz: In both of these reconcilers the "Finalize" interface is not intended explicitly to mean "remove a finalizer" in the metadata sense, but rather "finalize the lifecycle of this object". I can see how this is confusing, but struggled to find a better name for the interface.

The managed resource reconciler finalizes its work on a managed resource by removing its finalizer. The claim binding reconclise finalizes its work on a managed resource by removing its claim ref and setting it to unbound.

negz:
> But per #51 there is no need to make k8s finalizer mechanism an Initializer at all.

I disagree with this. You're right that in practice pretty much no-one is going to disable adding and removing finalizers, but I think we should keep them abstracted in something like an `Initializer`. Currently the responsibility of the managed reconciler's `Reconcile` method is:

* Update the `Synced` condition.
* Update the managed resource status (only, never anything else).
* Orchestrate various interfaces that do the real work.

I'd like to keep the responsibility of updating finalizer metadata contained within a separate type that we can stub out. However per our discussion this morning I think we could probably make this less obtuse by:

* Removing the Annotator interface, and keeping Initializer. This would set the external name.
* Naming the interfaces that add and remove the finalizer more specifically `FinalizerAdder` and `FinalizerRemover`.

negz:
I've broken out two PRs from this one:

* https://github.com/crossplaneio/crossplane-runtime/pull/67
* https://github.com/crossplaneio/crossplane-runtime/pull/66

I'll break some more out once the above are merged, since much of my updates assume the new testing style.

muvaf:
> Removing the Annotator interface, and keeping Initializer. This would set the external name.
    Naming the interfaces that add and remove the finalizer more specifically FinalizerAdder and FinalizerRemover.

This seems like a good path forward to me as discussed. Having the `FinalizerAdder` and `FinalizerRemover` pluggable would make the function more testable as you pointed out.

About the naming of the `ManagedFinalizer` interface and `Finalize` function, what do you think about using `Release`/`TearDown`/`Complete` instead of `Finalize`? It's just so easy to mix this `Finalize` word with k8s finalizers.

negz:
#68 and #70 cover all the changes I was pondering in this draft, so I'm closing this in favor of those smaller PRs.

Issue #21, opened 2019-07-09
Implement updated reclaim policy strategy


hasheddan:
Community discussion has settled on using `reclaimPolicy` in a similar manner to how Kubernetes `PersistentVolume` `reclaimPolicy` works. This is a change from the current status of `reclaimPolicy` in Crossplane. For more information, see the [**Goals** section](https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-resource-reclaim-policy.md#goals) of the [reclaim policy one-pager](https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-resource-reclaim-policy.md).

Implementation of sentiment expressed in crossplaneio/crossplane#550

PR #70, opened 2019-11-01
Add managed resource finalizer immediately before creating

### LINKED ISSUE #25: Statically provisioned resources should be unbound when their claim is deleted
### What happened?

The resource claim reconciler supports either dynamically or statically provisioning managed resources. The former involves using a resource class to create a new managed resource, while the latter involves binding to an existing resource.

A resource claim is the Kubernetes owner reference of any managed resource it dynamically provisions, but not the owner reference of any statically provisioned managed resource it later binds to. When a resource claim is deleted its underlying managed resource will only be deleted if it was dynamically provisioned and the user did not supply the `--cascade` option to `kubectl delete`.

While it is possible to delete a resource claim without deleting its managed resource, the resource claim reconciler doesn't currently update the managed resource in this scenario. It remains in binding state "bound", with a claim reference to the (now deleted) claim that it was bound to. This is probably not the behaviour we want.

### How can we reproduce it?

1. Statically provision a Crossplane managed resource
1. Create a claim that specifies the managed resource from step 1 in its `.spec.resourceRef`
1. Wait for the claim and resource to bind.
1. Delete the claim created in step 2
1. Observe that the managed resource created in step 1 continues to be "bound" to the now deleted resource claim.

Crossplane version:

### LINKED ISSUE #51: Make use of finalizer mechanism mandatory

### LINKED ISSUE #62: Managed Reconciler: A resource with unresolved references never gets deleted

### LINKED ISSUE #63: Managed reconciler should set the finalizer only after resource is created

negz:

Fixes #63
Fixes #62
Fixes #51
Fixes #25 

This commit moves where we set the finalizer for managed resources to right before creating them, not at the beginning of the reconcile. This means we'll be less likely to encounter issues where we can't delete a managed resource because we could never create it in the first place, but we added a finalizer.

* By the time we get here we know our Observe call worked. If (for example) our cloud provider credentials were completely wrong, we'd never proceed far enough to add the finalizer.
* If Observe works but Create fails (for example because we had RO cloud provider credentials) we would already have added the finalizer, but...
* When the managed resource was deleted we'd be able to Observe that the external resource does not exist (because we were never able to Create it) and thus would not call Delete on the external resource and go straight to unpublishing credentials and removing the finalizer.

This commit also renames and refactors a bunch of our interfaces to use less obtuse names. Previously sometimes a "finalize" method unbound a managed resource, while at other times it removed the finalizer. Similarly, finalizers were added in "initialize". We now have a 'Binder' interface with bind and unbind methods, and two 'Finalizer' interfaces (one for Claim, and one for Managed) that add and remove finalizers, as you would expect.

### Checklist

I have:

### DIFF
--- a/pkg/resource/claim_binding_reconciler.go
+++ b/pkg/resource/claim_binding_reconciler.go
@@ -193,7 +193,6 @@ type crManaged struct {
 	ManagedConfigurator
 	ManagedCreator
 	ManagedConnectionPropagator
-	Binder
 	ManagedFinalizer
 }
 
@@ -202,16 +201,19 @@ func defaultCRManaged(m manager.Manager) crManaged {
 		ManagedConfigurator:         NewObjectMetaConfigurator(m.GetScheme()),
 		ManagedCreator:              NewAPIManagedCreator(m.GetClient(), m.GetScheme()),
 		ManagedConnectionPropagator: NewAPIManagedConnectionPropagator(m.GetClient(), m.GetScheme()),
-		Binder:                      NewAPIBinder(m.GetClient(), m.GetScheme()),
 	}
 }
 
 type crClaim struct {
 	ClaimFinalizer
+	Binder
 }
 
 func defaultCRClaim(m manager.Manager) crClaim {
-	return crClaim{ClaimFinalizer: NewAPIClaimFinalizer(m.GetClient(), claimFinalizerName)}
+	return crClaim{
+		ClaimFinalizer: NewAPIClaimFinalizer(m.GetClient(), claimFinalizerName),
+		Binder:         NewAPIBinder(m.GetClient(), m.GetScheme()),
+	}
 }
 
 // A ClaimReconcilerOption configures a ClaimReconciler.
@@ -246,7 +248,7 @@ func WithManagedConnectionPropagator(p ManagedConnectionPropagator) ClaimReconci
 // resources to their claim.
 func WithBinder(b Binder) ClaimReconcilerOption {
 	return func(r *ClaimReconciler) {
-		r.managed.Binder = b
+		r.claim.Binder = b
 	}
 }
 
@@ -329,7 +331,7 @@ func (r *ClaimReconciler) Reconcile(req reconcile.Request) (reconcile.Result, er
 	}
 
 	if meta.WasDeleted(claim) {
-		if err := r.managed.Unbind(ctx, claim, managed); err != nil {
+		if err := r.claim.Unbind(ctx, claim, managed); err != nil {
 			// If we didn't hit this error last time we'll be requeued
 			// implicitly due to the status update. Otherwise we want to retry
 			// after a brief wait, in case this was a transient error.
@@ -382,14 +384,6 @@ func (r *ClaimReconciler) Reconcile(req reconcile.Request) (reconcile.Result, er
 			return reconcile.Result{RequeueAfter: aShortWait}, errors.Wrap(r.client.Status().Update(ctx, claim), errUpdateClaimStatus)
 		}
 
-		if err := r.claim.AddFinalizer(ctx, claim); err != nil {
-			// If we didn't hit this error last time we'll be requeued
-			// implicitly due to the status update. Otherwise we want to retry
-			// after a brief wait, in case this was a transient error.
-			claim.SetConditions(v1alpha1.Creating(), v1alpha1.ReconcileError(err))
-			return reconcile.Result{RequeueAfter: aShortWait}, errors.Wrap(r.client.Status().Update(ctx, claim), errUpdateClaimStatus)
-		}
-
 		if err := r.managed.Create(ctx, claim, class, managed); err != nil {
 			// If we didn't hit this error last time we'll be requeued
 			// implicitly due to the status update. Otherwise we want to retry
@@ -428,7 +422,15 @@ func (r *ClaimReconciler) Reconcile(req reconcile.Request) (reconcile.Result, er
 			return reconcile.Result{RequeueAfter: aShortWait}, errors.Wrap(r.client.Status().Update(ctx, claim), errUpdateClaimStatus)
 		}
 
-		if err := r.managed.Bind(ctx, claim, managed); err != nil {
+		if err := r.claim.AddFinalizer(ctx, claim); err != nil {
+			// If we didn't hit this error last time we'll be requeued
+			// implicitly due to the status update. Otherwise we want to retry
+			// after a brief wait, in case this was a transient error.
+			claim.SetConditions(v1alpha1.Creating(), v1alpha1.ReconcileError(err))
+			return reconcile.Result{RequeueAfter: aShortWait}, errors.Wrap(r.client.Status().Update(ctx, claim), errUpdateClaimStatus)
+		}
+
+		if err := r.claim.Bind(ctx, claim, managed); err != nil {
 			// If we didn't hit this error last time we'll be requeued implicitly
 			// due to the status update. Otherwise we want to retry after a brief
 			// wait, in case this was a transient error.
--- a/pkg/resource/claim_binding_reconciler.go
+++ b/pkg/resource/claim_binding_reconciler.go
@@ -120,57 +120,54 @@ func (fn ManagedConnectionPropagatorFn) PropagateConnection(ctx context.Context,
 	return fn(ctx, cm, mg)
 }
 
-// A ManagedBinder binds a resource claim to a managed resource.
-type ManagedBinder interface {
+// A Binder binds a resource claim to a managed resource.
+type Binder interface {
+	// Bind the supplied Claim to the supplied Managed resource.
 	Bind(ctx context.Context, cm Claim, mg Managed) error
+
+	// Unbind the supplied Claim from the supplied Managed resource.
+	Unbind(ctx context.Context, cm Claim, mg Managed) error
 }
 
-// A ManagedBinderFn is a function that satisfies the ManagedBinder interface.
-type ManagedBinderFn func(ctx context.Context, cm Claim, mg Managed) error
-
-// Bind the supplied resource claim to the supplied managed resource.
-func (fn ManagedBinderFn) Bind(ctx context.Context, cm Claim, mg Managed) error {
-	return fn(ctx, cm, mg)
+// BinderFns satisfy the Binder interface.
+type BinderFns struct {
+	BindFn   func(ctx context.Context, cm Claim, mg Managed) error
+	UnbindFn func(ctx context.Context, cm Claim, mg Managed) error
 }
 
-// A ManagedFinalizer finalizes the deletion of a resource claim.
-type ManagedFinalizer interface {
-	Finalize(ctx context.Context, cm Managed) error
+// Bind the supplied Claim to the supplied Managed resource.
+func (b BinderFns) Bind(ctx context.Context, cm Claim, mg Managed) error {
+	return b.BindFn(ctx, cm, mg)
 }
 
-// A FinalizerChain chains multiple managed finalizers.
-type FinalizerChain []ManagedFinalizer
-
-// Finalize calls each ManagedFinalizer serially. It returns the first
-// error it encounters, if any.
-func (cc FinalizerChain) Finalize(ctx context.Context, mg Managed) error {
-	for _, c := range cc {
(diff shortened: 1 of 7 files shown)


hasheddan:
Haven't finished reviewing yet, but love the addition of the `release/breaking-api-change` label to this repo!

pkg/resource/claim_binding_reconciler.go
```diff
@@ -394,6 +382,14 @@ func (r *ClaimReconciler) Reconcile(req reconcile.Request) (reconcile.Result, er
 			return reconcile.Result{RequeueAfter: aShortWait}, errors.Wrap(r.client.Status().Update(ctx, claim), errUpdateClaimStatus)
 		}
 
+		if err := r.claim.AddFinalizer(ctx, claim); err != nil {
```
muvaf: This would add finalizer to claim only if the managed resource is not created yet. I know that we have scenarios where the managed resource exists and our newly created claim binds to it, so, in those cases we'd see an immediate deletion of the claim. What do you think about calling `r.claim.AddFinalizer ` right before `if !meta.WasCreated(managed) && claim.GetClassReference() != nil` just like managed reconciler?

muvaf:
Linking https://github.com/crossplaneio/crossplane-runtime/pull/65 for the discussions happened on that PR.

pkg/resource/claim_binding_reconciler.go
negz: Now that you mention it it's likely that this is why statically provisioned managed resources are never properly unbound per https://github.com/crossplaneio/crossplane-runtime/issues/25. I agree that this isn't the right place to add the finalizer, but I actually think we should set it right before binding to our managed resource.

The managed resource reconciler uses the finalizer to delete the external resource (if necessary) whereas the claim reconciler uses the finalizer to _unbind_ the managed resource. If we never bound to the resource there's nothing to do. Keep in mind we currently rely on garbage collection to ensure dynamically provisioned managed resources are deleted when the claim is deleted.

negz:
Confirmed I can delete a managed resource that fails to create due to bad credentials. It gets the finalizer, but skips over the delete block because it can't observe the resource:

```
$ kubectl describe network
Name:         sample-network
Namespace:    
Labels:       <none>
Annotations:  crossplane.io/external-name: sample-network
              kubectl.kubernetes.io/last-applied-configuration:
                {"apiVersion":"compute.gcp.crossplane.io/v1alpha3","kind":"Network","metadata":{"annotations":{},"name":"sample-network"},"spec":{"autoCre...
API Version:  compute.gcp.crossplane.io/v1alpha3
Kind:         Network
Metadata:
  Creation Timestamp:  2019-11-02T02:55:22Z
  Finalizers:
    finalizer.managedresource.crossplane.io
  Generation:        1
  Resource Version:  329299
  Self Link:         /apis/compute.gcp.crossplane.io/v1alpha3/networks/sample-network
  UID:               551bcd1c-d87a-4836-a953-42971069353f
Spec:
  Auto Create Subnetworks:  false
  Name:                     my-cool-network
  Provider Ref:
    Name:          example
  Reclaim Policy:  Delete
  Routing Config:
    Routing Mode:  REGIONAL
Status:
  Conditions:
    Last Transition Time:  2019-11-02T02:55:23Z
    Reason:                Successfully resolved managed resource references to other resources
    Status:                True
    Type:                  ReferencesResolved
    Last Transition Time:  2019-11-02T02:55:23Z
    Reason:                Managed resource is being created
    Status:                False
    Type:                  Ready
    Last Transition Time:  2019-11-02T02:55:23Z
    Message:               creation of Network resource has failed: googleapi: Error 404: Failed to find project playgroundz, notFound
    Reason:                Encountered an error during managed resource reconciliation
    Status:                False
    Type:                  Synced
Events:                    <none>

$ kubectl delete -f network.yaml 
network.compute.gcp.crossplane.io "sample-network" deleted
```

negz:
Also confirmed I can delete a managed resource that cannot resolve its references:

```
$ kubectl describe subnetwork                                                                                                                         
Name:         sample-subnetwork                                                                                                                                             
Namespace:                                                                                                                                                                         
Labels:       <none>                                                           
Annotations:  crossplane.io/external-name: sample-subnetwork                             
              kubectl.kubernetes.io/last-applied-configuration:                                                                                                                    
                {"apiVersion":"compute.gcp.crossplane.io/v1alpha3","kind":"Subnetwork","metadata":{"annotations":{},"name":"sample-subnetwork"},"spec":{"i...                      
API Version:  compute.gcp.crossplane.io/v1alpha3                                         
Kind:         Subnetwork                                                                                                                                                          
Metadata:                                   
  Creation Timestamp:  2019-11-02T02:49:30Z                                              
  Generation:          1                                                                                                                                                    
  Resource Version:    328806                                                                                                                                                      
  Self Link:           /apis/compute.gcp.crossplane.io/v1alpha3/subnetworks/sample-subnetwork                                                                                      
  UID:                 41e54e35-ad67-4d01-893c-1b5d02103404                              
Spec:                                                                                                                                                                              
  Ip Cidr Range:  192.168.0.0/24                                               
  Name:           my-cool-subnetwork                                                     
  Network Ref:                                                                                                                                                                     
    Name:                    sample-netwark                                    
  Private Ip Google Access:  true                                                                                                                                           
  Provider Ref:                                                                          
    Name:          example                                                                                                                                                         
  Reclaim Policy:  Delete                                                                                        
  Region:          us-central1                                                                                                                                                     
  Secondary Ip Ranges:                                                          
    Ip Cidr Range:  10.0.0.0/8                                                                                                                                                    
    Range Name:     pods
    Ip Cidr Range:  172.16.0.0/16
    Range Name:     services
Status:
  Conditions:
    Last Transition Time:  2019-11-02T02:49:31Z
    Message:               [{reference:sample-netwark status:NotFound}]
    Reason:                One or more referenced resources do not exist, or are not yet Ready                                                                                         Status:                False
    Type:                  ReferencesResolved
Events:                    <none>

$ kubectl delete subnetwork --all
subnetwork.compute.gcp.crossplane.io "sample-subnetwork" deleted
```

negz:
Confirmed this fixes #25. `standard-mysql-negz-again` is a statically provisioned managed resource.

```
$ kubectl get -o json cloudsqlinstance standard-mysql-negz-again | jq .status.bindingPhase
"Unbound"

$ kubectl get -o json cloudsqlinstance standard-mysql-negz-again | jq .spec.claimRef
null

$ kubectl apply -f database.yaml 
mysqlinstance.database.crossplane.io/sql created

$ kubectl get -o json cloudsqlinstance standard-mysql-negz-again | jq .status.bindingPhase
"Bound"
$ kubectl get -o json cloudsqlinstance standard-mysql-negz-again | jq .spec.claimRef
{
  "apiVersion": "database.crossplane.io/v1alpha1",
  "kind": "MySQLInstance",
  "name": "sql",
  "namespace": "default",
  "uid": "f5b0e2ef-e79b-4171-b19c-9bae369fa867"
}

messremb(negz@wordpress)$ kubectl delete -f database.yaml 
mysqlinstance.database.crossplane.io "sql" deleted

$ kubectl get -o json cloudsqlinstance standard-mysql-negz-again | jq .status.bindingPhase
"Unbound"

$ kubectl get -o json cloudsqlinstance standard-mysql-negz-again | jq .spec.claimRef
null
```

pkg/resource/claim_binding_reconciler.go
muvaf: > the claim reconciler uses the finalizer to unbind the managed resource

> we currently rely on garbage collection to ensure dynamically provisioned managed resources are deleted when the claim is deleted.

Well, aren't these anti-pattern anyway? Bind/Unbind mechanism should only be meaningful if the reclaim policy between claim and managed resource is set to `Retain`. If it's `Delete`, we should just delete it explicitly. But there is no claim<>managed reclaim policy and all this seems to be related to #21 , so, it should be OK for this PR not to touch that.

IMHO, sounds to me it's time we finally fix #21 and #27 . I think I can work on a one-pager that covers both. Not much to talk about #21 as it will just work like PVC/PV model but I think it'd be good to cover both in the same document to explain the mechanisms better. Maybe after completing Azure redis & db v1beta1. @prasek @negz what do you guys think?

pkg/resource/claim_binding_reconciler.go
```diff
@@ -202,16 +201,19 @@ func defaultCRManaged(m manager.Manager) crManaged {
 		ManagedConfigurator:         NewObjectMetaConfigurator(m.GetScheme()),
 		ManagedCreator:              NewAPIManagedCreator(m.GetClient(), m.GetScheme()),
 		ManagedConnectionPropagator: NewAPIManagedConnectionPropagator(m.GetClient(), m.GetScheme()),
-		Binder:                      NewAPIBinder(m.GetClient(), m.GetScheme()),
 	}
 }
 
 type crClaim struct {
 	ClaimFinalizer
+	Binder
```
muvaf: I like that change. It sounds to me binding act belongs to claim, so, it makes sense for claim to `crClaim` manage it.

negz:
> I realized `APIStatusBinder` is just same as `APIBinder`

@muvaf Not quite, unfortunately. `APIStatusBinder` updates both the managed resource and its status (subresource), while `APIBinder` updates only the managed resource. `APIStatusBinder` is for managed resources that use the status subresource, while `APIBinder`  is for older managed resources that do not. I hope we'll be able to remove `APIBinder` once we no longer have managed resources that don't use the status subresource.

muvaf:
@negz Ah yes, I always forget that normal `Update` call cannot update the status subresource. Thanks for the clarification!

pkg/resource/claim_binding_reconciler.go
negz: > Well, aren't these anti-pattern anyway?

It is. I agree that we will likely need to refactor how we add and remove claim finalizers once we're not relying on garbage collection to delete managed resources when the claim is deleted.

> IMHO, sounds to me it's time we finally fix #21 and #27 . I think I can work on a one-pager that covers both. Not much to talk about #21 as it will just work like PVC/PV model but I think it'd be good to cover both in the same document to explain the mechanisms better.

I can see your point, but keep in mind @hasheddan already produced a one pager for #21. I agree that design done for #27 should consider how it relates to reclaim policies but let's be careful not to re-do that work if we don't need to.

> Maybe after completing Azure redis & db v1beta1. @prasek @negz what do you guys think?

My suspicion is that #21 will mostly involve changes to the resource claim reconciler, while #27 could impact how we write `resource.ExternalClient` implementations. On one hand I can see that these are both potentially "it's never the right time" kind of issues, but I feel like they're lower priority than getting all of our managed resource controllers ported to the managed resource reconciler (not necessarily v1beta1).

PR #75, opened 2019-11-05
Don't resolve references when being deleted

### LINKED ISSUE #74: References to deleting managed resources blocks deletion.
### What happened?

In https://github.com/crossplaneio/crossplane-runtime/pull/66 we updated the managed resource reconciler such that managed resources resolve their references on every reconcile. We resolve our references before we observe, create, update, or delete the external resource that corresponds to our managed resource. This is because:

1. We must resolve references before we observe the external resource, e.g. for resources like `connection.servicenetworking.gcp.crossplane.io` that may need its `.spec.networkRef` set in order to observe the external resource.
1. We must observe the external resource in order to know whether we need to create, update, or delete it.

Cross resource resolution returns an error and does not proceed when the managed resource under reconciliation is dependent upon a resource that does not exist, or does not have condition `Ready=True`. This means that when, for example:

* A Subnetwork references a Network.
* A Network cannot be deleted until all of its Subnetworks have been deleted.
* A Subnetwork cannot be deleted when its Network is deleting.

We end up stuck. The Subnetwork can't be deleted because the Network (managed resource) is deleting, and the Network can't be deleted because the Subnetwork (external resource) hasn't been deleted.

### How can we reproduce it?

1. Create GCP Network managed resource.
1. Create a GCP Subnetwork managed resource that references said Network.
1. Delete the Network resource.
1. Delete the Subnetwork resource.

Observe that neither resource ever successfully deletes.

Crossplane version:

negz:

Fixes #74 (at least, fixes it enough for now)

Per the comment, there's too high a chance we'll get stuck and not process the delete because we reference resources that are also being deleted. I considered:

* Processing references at delete time but not blocking on accessor errors. I felt this introduced too much complexity for little gain compared to just not trying resolution at all.
* Making AttributeReferencers no-ops when the field they would set already had a value. I think this is an avenue we should investigate (issue forthcoming) but it is awkward to implement with the current AttributeReferencer interface.
* Enforcing ordered deletes, such that a referenced resource cannot be deleted (at least not by Crossplane) until its referencers have all been deleted. This is the most bulletproof, but also the most complicated solution and would require further design investigation to pursue.

### Checklist

I have:

### DIFF
--- a/pkg/resource/managed_reconciler.go
+++ b/pkg/resource/managed_reconciler.go
@@ -455,20 +455,28 @@ func (r *ManagedReconciler) Reconcile(req reconcile.Request) (reconcile.Result, 
 	// We resolve any references before observing our external resource because
 	// in some rare examples we need a spec field to make the observe call, and
 	// that spec field could be set by a reference.
-	if err := r.managed.ResolveReferences(ctx, managed); err != nil {
-		condition := v1alpha1.ReconcileError(err)
-		if IsReferencesAccessError(err) {
-			condition = v1alpha1.ReferenceResolutionBlocked(err)
-		}
+	//
+	// We do not resolve references when being deleted because it is likely that
+	// the resources we reference are also being deleted, and would thus block
+	// resolution due to being unready or non-existent. It is unlikely (but not
+	// impossible) that we need to resolve a reference in order to process a
+	// delete, and that reference is stale at delete time.
+	if !meta.WasDeleted(managed) {
+		if err := r.managed.ResolveReferences(ctx, managed); err != nil {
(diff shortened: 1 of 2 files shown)


negz:
I tested this by using the `stack-gcp` release branch with a `replace` statement in my `go.mod` to this branch, and everything seems to work as expected. I created stuff out of order (i.e. subnet before network) and confirmed they blocked until references were resolvable, then deleted stuff (network, subnetwork, globaladdress, gkecluster, cloudsqlinstance, etc) in random order and confirmed they were all (eventually) deleted successfully.

PR #80, opened 2019-11-15
mods: bump controller-runtime to v0.3.0

hasheddan:

Bumps `controller-runtime` to v0.3.0 and updates related k8s dependencies to `kubernetes-1.15.4`.

### Checklist

I have:


hasheddan:
Note: only breaking change in the crossplane-runtime version bump is removal of `dep`

negz:
@hasheddan Can you provide a little context around the need for this change? Do we have to bump the Kubernetes dependencies along with controller-runtime?

hasheddan:
Sure! So from what I can tell, the controller-runtime bump has k8s dependencies bumped, which included a breaking change with client-go. In general, I think it would be good to be using the same version of all k8s.io dependencies across our controllers, but the version is going to depend on what version controller-runtime uses as long as we depend on it. In this case it is v1.15.4

PR #83, opened 2019-11-18
Don't set controller references for dynamically provisioned resources

negz:

Resource claims are namespaced, but the managed resources they dynamically provision are cluster scoped. A namespaced resource cannot, by design, own a cluster scoped resource. This has been a known issue for a long time, but we recently learned that it can result in the managed resources being unintentionally garbage collected by Kubernetes per https://github.com/crossplaneio/stack-gcp/issues/99.

Resource claims and managed resources already reference each other via their resourceRef and claimRef; we had set the owner reference purely to delete dynamically provisioned managed resources along with their claim. Dynamically provisioned managed resources will now always stick around until explicitly cleaned up. We intend to address this by repurposing the existing reclaimPolicy field per https://github.com/crossplaneio/crossplane-runtime/issues/21

### Checklist

I have:

### DIFF
--- a/pkg/resource/configurator.go
+++ b/pkg/resource/configurator.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"fmt"
 
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 
 	"github.com/crossplaneio/crossplane-runtime/pkg/meta"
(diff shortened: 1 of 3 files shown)


pkg/resource/configurator.go
```diff
@@ -41,33 +40,30 @@ func (cc ConfiguratorChain) Configure(ctx context.Context, cm Claim, cs Class, m
...
-	typer runtime.ObjectTyper
-}
+// provisioned resource, deriving it from a class and claim. It is deprecated;
+// use ConfigureNames instead.
+type ObjectMetaConfigurator struct{}
 
 // NewObjectMetaConfigurator returns a new ObjectMetaConfigurator.
-func NewObjectMetaConfigurator(t runtime.ObjectTyper) *ObjectMetaConfigurator {
-	return &ObjectMetaConfigurator{typer: t}
+func NewObjectMetaConfigurator(_ runtime.ObjectTyper) *ObjectMetaConfigurator {
```
hasheddan: I like how this makes it such that `NewObjectMetaConfigurator` code does not have to be modified! What do you think our policy for removing deprecated code here should be given this library is pre-v1.0?

negz:
I've tested this change by:

1. Dynamically provisioning a CloudSQLInstance via a PostgreSQLInstance claim
1. Deleting the PostgreSQLInstance claim.
1. Confirming the CloudSQLInstance remains, but returns to the Unbound binding phase, with a nil claim reference.

pkg/resource/configurator.go
negz: As far as I can tell https://semver.org/ does not restrict us in any way while we're below version 1. I think a good compromise would be:

1. Don't break API compatibility on a patch release.
1. It's okay to mark an API as deprecated on a patch release.
1. A little subjective, but it's okay to make trivial behaviour changes on a patch release.

muvaf: IMHO, it should be fine to release a deprecation notice in a release and after that one break the function signature directly. The consumers will see the error on build time, so, it should be fine and fairly easy for them to fix it.

negz:
> @negz Why not keep the current behavior by just deleting the managed resource when the claim is deleted via resourceRef instead of owner reference?

@muvaf This was the minimal change to fix the bug; I didn't want to do too much in one PR. I will have another PR open in a few minutes that reintroduces support for deleting managed resources when their claim is deleted.

negz:
> I will have another PR open in a few minutes

https://github.com/crossplaneio/crossplane-runtime/pull/87

PR #85, opened 2019-11-18
Revert "mods: bump controller-runtime to v0.3.0"

negz:
Reverts crossplaneio/crossplane-runtime#80

I'm going to revert this for the time being, unless someone can figure out how to build a stack with it without changing too much. My reasoning:

* As far as I know we're only doing this bump as a best practice, rather than because it's needed for anything in particular right now.
* I've been trying unsuccessfully to build stack-gcp against this change for over an hour.
* We don't yet have a release branch strategy for crossplane-runtime, which means it's difficult to release a patch to crossplane-runtime without requiring stacks to figure out this dependency update.

It's not quite obvious to me what is making this set of dependencies so hard to reconcile.


negz:
Some further context:

* The driver for this change was @hasheddan's work on integration testing.
* It's possible to get a stack to build using `replace` statements per https://github.com/crossplaneio/stack-gcp/pull/102, but that seems less than ideal.
* I believe one theory as to why this was so tricky is that stacks have a direct dependency on controller-runtime and friends and also an indirect dependency on them via the `github.com/crossplaneio/crossplane` module. This puts us in the unfortunate position of needing to update the crossplane module's dependency on controller-runtime before we can update stack dependencies on crossplane-runtime. I've raised https://github.com/crossplaneio/crossplane/issues/1085 to suggest an improvement to this.

Issue #22, opened 2019-04-19
Allow existing resources to be "imported" into Crossplane to be managed


ichekrygin:
### Overview
One of the key Crossplane functionality is support for provisioning and life-cycle management of the public cloud managed resources. Crossplane offers two modes of resource creation:
- Static: Resource is created using a dedicated resource type (CR)
- Dynamic: Resource is created using ResourceClaim type (CR)
In either case, crossplane takes "full lifecycle" ownership of the created resource, and based on the specified Reclaim Policy, can delete managed resource upon the clean-up,

There are, however, use cases when the user would like to represent existing resources (i.e. resources provisioned outside the Crossplane) as Crossplane Resource instances.  To achieve that, Crossplane must support resource import functionality.

### Further Consideration
In order for Crossplane to provide resource import functionality following questions need to be addressed:

#### Resource Lifecycle
Will the crossplane become a "full" owner of the managed resource lifecycle.

###### Deletion
If crossplane imports a resource, can this resource be specified with ReclaimPolicy: Delete. Initial thinking, no, it cannot, i.e. all imported resources automatically get assigned (and hopefully at some point enforced) Reclaim Policy: Retain ("cannot delete something you didn't create")

##### Update
Can crossplane perform manage resource update, i.e. changing resource state/properties with the potential impact on any/all existing downstream resource consumers?

#### Cloud Provider compatibility
There should be a special concideration that the imported resource could be "fully" managed by the existing cloud provider (service account scopes/permissions, etc)

muvaf:
I think https://github.com/crossplaneio/crossplane-runtime/pull/45 makes it pretty easy to import a resource. Just tried with [new CloudSQL controller](https://github.com/crossplaneio/stack-gcp/pull/46) that incorporated the late initialize pattern from https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-managed-resource-api-design.md#high-fidelity

All you have to do:
- Have your resource class `Spec.ForProvider` struct empty so that after acquiring the resource, you don't try to make updates immediately. All fields would get late-initialized with the values received from the provider.
- Create a claim with annotation `crossplane.io/external-name: <your resource name>`

Then it will acquire the resource spec (into managed resource), bind it to the claim and propagate its credentials to claim's namespace. That goes for a resource (CloudSQL) that allows you to retrieve the credentials even after the creation though. So, for resources that send you the credentials only during creation there would be some manual steps getting them into managed resource's secret.

[...]

PR #87, opened 2019-11-19
Mirror PersistentVolume ReclaimPolicy semantics

### LINKED ISSUE #21: Implement updated reclaim policy strategy

negz:

Fixes #21 

This commit changes the meaning of the managed resource reclaim policy to match https://kubernetes.io/docs/concepts/storage/persistent-volumes/#reclaiming as closely as possible, minus the deprecated 'Recycle' policy. Previously the reclaim policy dictated only what happened to the external resource when its managed resource was deleted.

Note that the implementation here differs slightly from some of our discussions around the reclaim policy, but I believe that is due to our having misunderstood how the persistent volume reclaim policy that we wanted to mirror works. Based on my read of the above link, the behaviour I intend for this PR is:

* Managed resources using the `Delete` policy are deleted when their bound resource claim is deleted.
* Managed resources using the `Delete` policy delete their external resource when they are deleted.
* Managed resources using the `Retain` policy are _not_ deleted when their bound resource claim is deleted.
* Managed resources using the `Retain` policy do _not_ delete their external resource when they are deleted.
* All managed resources transition to binding phase `Released` (not `Unbound`) when their resource claim is deleted.
* A `Released` managed resource cannot be reused; it sticks around only so the infrastructure operator can perform any cleanup tasks they need to before manually deleting it.
* A managed resource without a reclaim policy defaults to `Retain`, in that the absence of a reclaim policy is treated as `Retain`.
* A resource class without a reclaim policy will dynamically provision managed resources with the `Delete` reclaim policy.

### Testing Checklist

Before merging this PR I intend to:

- [x] Ensure statically provisioned managed resources default to `Retain` by dynamically provisioning a managed resource without a reclaim policy and observing that the managed resource's reclaim policy is unset.
- [x] Ensure dynamically provisioned managed resources default to `Delete` by dynamically provisioning a managed resource using a resource claim without a resource policy and observing that the managed resource's reclaim policy is `Delete`.
- [x] Ensure managed resources with policy `Retain` (or unset) are retained in binding phase `Released` when their claim is deleted.
- [x] Ensure managed resources with policy `Retain` (or unset) retain their external resources when deleted.
- [x] Ensure managed resources with policy `Delete` are deleted when their claim is deleted.
- [x] Ensure managed resources with policy `Delete` delete their external resources when deleted.

### Checklist

I have:

### DIFF
--- a/apis/core/v1alpha1/resource.go
+++ b/apis/core/v1alpha1/resource.go
@@ -150,13 +150,18 @@ type ResourceSpec struct {
 	// observe, update, and delete this managed resource.
 	ProviderReference *corev1.ObjectReference `json:"providerRef"`
 
-	// ReclaimPolicy specifies what will happen to the external resource this
-	// managed resource manages when the managed resource is deleted. "Delete"
-	// deletes the external resource, while "Retain" (the default) does not.
-	// Note this behaviour is subtly different from other uses of the
-	// ReclaimPolicy concept within the Kubernetes ecosystem per
-	// https://github.com/crossplaneio/crossplane-runtime/issues/21
+	// ReclaimPolicy specifies what will happen to this managed resource when
+	// its resource claim is deleted, and what will happen to the underlying
+	// external resource when the managed resource is deleted. The "Delete"
+	// policy causes the managed resource to be deleted when its bound resource
+	// claim is deleted, and in turn causes the external resource to be deleted
+	// when its managed resource is deleted. The "Retain" policy causes the
+	// managed resource to be retained, in binding phase "Released", when its
+	// resource claim is deleted, and in turn causes the external resource to be
+	// retained when its managed resource is deleted. The "Retain" policy is
+	// used when no policy is specified.
 	// +optional
+	// +kubebuilder:validation:Enum=Retain;Delete
 	ReclaimPolicy ReclaimPolicy `json:"reclaimPolicy,omitempty"`
 }
 
@@ -180,12 +185,18 @@ type ClassSpecTemplate struct {
 	// provisioned using this resource class.
 	ProviderReference *corev1.ObjectReference `json:"providerRef"`
 
-	// ReclaimPolicy specifies what will happen to external resources when
-	// managed resources dynamically provisioned using this resource class are
-	// deleted. "Delete" deletes the external resource, while "Retain" (the
-	// default) does not. Note this behaviour is subtly different from other
-	// uses of the ReclaimPolicy concept within the Kubernetes ecosystem per
-	// https://github.com/crossplaneio/crossplane-runtime/issues/21
+	// ReclaimPolicy specifies what will happen to managed resources dynamically
+	// provisioned using this class when their resource claims are deleted, and
+	// what will happen to their underlying external resource when they are
+	// deleted. The "Delete" policy causes the managed resource to be deleted
+	// when its bound resource claim is deleted, and in turn causes the external
+	// resource to be deleted when its managed resource is deleted. The "Retain"
+	// policy causes the managed resource to be retained, in binding phase
+	// "Released", when its resource claim is deleted, and in turn causes the
+	// external resource to be retained when its managed resource is deleted.
+	// The "Retain" policy is used when no policy is specified, however the
+	// "Delete" policy is set at dynamic provisioning time if no policy is set.
 	// +optional
+	// +kubebuilder:validation:Enum=Retain;Delete
 	ReclaimPolicy ReclaimPolicy `json:"reclaimPolicy,omitempty"`
 }
--- a/pkg/resource/api.go
+++ b/pkg/resource/api.go
@@ -41,6 +41,7 @@ const (
 	errCreateOrUpdateSecret = "cannot create or update connection secret"
 	errUpdateManaged        = "cannot update managed resource"
 	errUpdateManagedStatus  = "cannot update managed resource status"
+	errDeleteManaged        = "cannot delete managed resource"
 )
 
 const claimFinalizerName = "finalizer." + claimControllerName
@@ -177,14 +178,25 @@ func (a *APIBinder) Bind(ctx context.Context, cm Claim, mg Managed) error {
 	return errors.Wrap(a.client.Update(ctx, cm), errUpdateClaim)
 }
 
-// Unbind the supplied Claim from the supplied Managed resource.
+// Unbind the supplied Claim from the supplied Managed resource by removing the
+// managed resource's claim reference, transitioning it to binding phase
+// "Released", and if the managed resource's reclaim policy is "Delete",
+// deleting it.
 func (a *APIBinder) Unbind(ctx context.Context, _ Claim, mg Managed) error {
-	// TODO(negz): We probably want to delete the managed resource here if its
-	// reclaim policy is delete, rather than relying on garbage collection, per
-	// https://github.com/crossplaneio/crossplane/issues/550
-	mg.SetBindingPhase(v1alpha1.BindingPhaseUnbound)
+	mg.SetBindingPhase(v1alpha1.BindingPhaseReleased)
 	mg.SetClaimReference(nil)
-	return errors.Wrap(IgnoreNotFound(a.client.Update(ctx, mg)), errUpdateManaged)
+	if err := a.client.Update(ctx, mg); err != nil {
+		return errors.Wrap(IgnoreNotFound(err), errUpdateManaged)
+	}
+
+	// We go to the trouble of unbinding the managed resource before deleting it
+	// because we want it to show up as "released" (not "bound") if its managed
+	// resource reconciler is wedged or delayed trying to delete it.
+	if mg.GetReclaimPolicy() != v1alpha1.ReclaimDelete {
+		return nil
+	}
+
+	return errors.Wrap(IgnoreNotFound(a.client.Delete(ctx, mg)), errDeleteManaged)
 }
 
 // An APIStatusBinder binds resources to claims by updating them in a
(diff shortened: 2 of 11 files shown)


pkg/resource/resource.go
```diff
@@ -142,9 +142,14 @@ func ResolveClassClaimValues(classValue, claimValue string) (string, error) {
 }
 
 // SetBindable indicates that the supplied Bindable is ready for binding to
-// another Bindable, such as a resource claim or managed resource.
+// another Bindable, such as a resource claim or managed resource by setting its
+// binding phase to "Unbound". It is a no-op for Bindables in phases "Bound" or
+// "Released", because these phases may not transition back to "Unbound".
 func SetBindable(b Bindable) {
-	if b.GetBindingPhase() == v1alpha1.BindingPhaseBound {
+	switch b.GetBindingPhase() {
```
muvaf: Using `switch` instead of just an `OR` operator feels odd to me here.

apis/core/v1alpha1/resource.go
```diff
@@ -180,12 +185,18 @@ type ClassSpecTemplate struct {
...
+	// ReclaimPolicy specifies what will happen to managed resources dynamically
+	// provisioned using this class when their resource claims are deleted, and
+	// what will happen to their underlying external resource when they are
+	// deleted. The "Delete" policy causes the managed resource to be deleted
+	// when its bound resource claim is deleted, and in turn causes the external
+	// resource to be deleted when its managed resource is deleted. The "Retain"
+	// policy causes the managed resource to be retained, in binding phase
+	// "Released", when its resource claim is deleted, and in turn causes the
+	// external resource to be retained when its managed resource is deleted.
+	// The "Retain" policy is used when no policy is specified, however the
```
muvaf: I think we should choose one default for all cases where policy isn't specified and stick with it.

negz: I'd prefer to mirror the semantics of the PersistentVolume reclaim policy, which this does.

pkg/resource/api.go
```diff
@@ -228,19 +240,29 @@ func (a *APIStatusBinder) Bind(ctx context.Context, cm Claim, mg Managed) error
 	return errors.Wrap(a.client.Update(ctx, cm), errUpdateClaim)
 }
 
-// Unbind the supplied Claim from the supplied Managed resource.
+// Unbind the supplied Claim from the supplied Managed resource by removing the
+// managed resource's claim reference, transitioning it to binding phase
+// "Released", and if the managed resource's reclaim policy is "Delete",
+// deleting it.
 func (a *APIStatusBinder) Unbind(ctx context.Context, _ Claim, mg Managed) error {
```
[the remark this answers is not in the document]
negz: `Release` might be a little better a name, but I think `Unbind` is still decent and would thus prefer to avoid the API churn of renaming the method.

pkg/resource/resource.go
negz: I don't feel strongly either way on this one. I tend to prefer switches because I think line wrapped if statements are ugly, and it's easy for `||` if statements to get long in Go.

negz:
> With this change, the behavior of ReclaimPolicy is extended to cover the relation between claim and managed, too. IMO, this is a public interface change, which would require a version bump on CRDs.

The Kubernetes [API versioning guidance](https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-versioning) is not super explicit about this. For beta resources it states:

> The schema and/or semantics of objects may change in incompatible ways in **a subsequent beta or stable release**.

As opposed to alpha APIs, which state:

> The API may change in incompatible ways in **a later software release** without notice.

If _a subsequent beta or stable release_ is interpreted to refer to the custom resource and not the software, then I agree that this is a semantic change and thus warrants a version bump for beta resources.

> However, we know that this is not the final destination we want to arrive per #21 and #27 . At some point, we want to separate the external resource policy from reclaim policy, which would then again be a breaking change. So, I think we should just implement external policy as well and then do the bump in this release only once.

I'm much less confident these days that the `externalResourcePolicy` suggestion I included in #21 and #27 is a good direction. It was informed by my wanting to mirror the semantics of the persistent volume reclaim policy, but misunderstanding that said policy actually controls the relationship between claim and volume _and between volume and external resource. @hasheddan spotted that a while back per https://github.com/crossplaneio/crossplane-runtime/issues/21#issuecomment-514656798 but I didn't grok it until today.

Presuming we still want to mirror the persistent volume semantics, then we don't need to (or indeed want to) introduce a second setting that determines whether external resources are deleted along with their managed resources; that is covered by `reclaimPolicy`.This leaves us potentially needing;

1. A setting that determines what we do when we want to create a managed resource that already exists.
1. A setting that determines what we do when an external resource's configuration drifts from its managed resource.

I'm quite strongly convinced these days that Crossplane should _always_ be the source of truth for the managed resources it manages. For the above this means:

1. The options are either to return an error or adopt the external resource into Crossplane management. If we adopt the external resource we immediately start driving it to our desired state as specified by the spec of the managed resource that adopted it.
1. Should not be configurable. Crossplane is always the source of truth, and we always drive the external resource toward our source of truth when it diverges.

In summary, I feel we should move forward with this PR and don't expect to change how `reclaimPolicy` works again anytime soon. I do think we'll still want to introduce something like an `adoptionPolicy` that controls whether we return an error or adopt an external resource we encounter when trying to create a managed resource.

muvaf:
> If a subsequent beta or stable release is interpreted to refer to the custom resource and not the software and we deem this semantic/behaviour change to be incompatible with the previous behaviour, then I agree that this warrants a version bump for beta resources.

I think this is indeed the case. Quoting from the versioning guideline:
> We chose to version at the API level rather than at the resource or field level to ensure that the API presents a clear, consistent view of system resources and behavior, and to enable controlling access to end-of-life and/or experimental APIs.
...
Note that API versioning and Software versioning are only indirectly related.

If it was a new field, I'd say we don't need the bump. But it's clear to me that the user gets a different behavior now when they choose `Retain` policy. Before, we'd delete the managed resource CR when a claim is deleted no matter what, but now we don't.

I feel like this is more of a product direction discussion rather than technical. I don't want to block this PR. Let's discuss it under https://github.com/crossplaneio/crossplane-runtime/issues/22

muvaf:
I am not sure about the fact that when reclaim policy is set to `Retain`, even though the resource is retained after claim deletion, it cannot be bound to another claim. I know this is how it works in PVC/PV model but I think the cloud resources are not as application-specific/low-level as volumes. It seems like fair game to me for another application to claim a managed resource and use it.

hasheddan:
Note: this will also require a slight change to core Crossplane as the `ResourceClaim` struct embeds `BindingStatus`. However, the claims would never be set to `Released` so I am not sure how urgent the change would be.

hasheddan:
@muvaf once we solve #22 then this could still be accomplished. The retained managed resources could be manually cleaned up, then the remaining external resources could be "imported" and used by another claim.

pkg/resource/api.go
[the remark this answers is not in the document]
muvaf: Yeah I see that, but still I'd expect the resource to be in `Unbound` state when I call `Unbind`

negz:
> Before, we'd delete the managed resource CR when a claim is deleted no matter what, but now we don't.

A small clarification - we'd delete the managed resource CR _by default_, but not no matter what. Whether or not the managed resource was deleted was determined by the [cascading deletion policy](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#setting-the-cascading-deletion-policy), so `kubectl delete --cascade=false someclaim` would not delete the managed resource.

> I am not sure about the fact that when reclaim policy is set to Retain, even though the resource is retained after claim deletion, it cannot be bound to another claim. I know this is how it works in PVC/PV model but I think the cloud resources are not as application-specific/low-level as volumes. It seems like fair game to me for another application to claim a managed resource and use it.

We discussed this on a call today, but I want to capture my (somewhat rambling) thoughts here too. I believe the following to be true with or without the change proposed by this PR:

* Persistent volume claims and resource claims can be consumed by multiple pods simultaneously.
* Persistent volume claims and resource claims can be consumed by multiple pods serially.

What changes with this PR is whether a managed resource can be _claimed_ (which is distinct from _used_) multiple times serially, which brings us to philosophising about _what is a claim_? I see managed resources (and classes) as being "offerings" published by an infrastructure operator, and resource claims being just that - "claims" to one of those offerings by a particular entity. So an entity can _claim_ a managed resource in a particular namespace, and that _claim_ can be _used_ by one or more entities within its namespace. At some point the entity who claimed the managed resource will not want it any more. I feel that a claim being deleted means "One or more entities were using this resource for a particular purpose or set of purposes. We no longer need this resource for that purpose or set of purposes." The question then is whether it's safe and sensible to offer the same underlying resource to another claimant, potentially in another namespace. Given that there is no guarantee that those entities cleaned up after themselves, I don't think it is a safe and sensible default to then allow another entity to claim that managed resource for a different purpose or set of purposes, given that by doing so they'll potentially gain access to any state that was left over by the previous claimant.

Put otherwise - "a big database instance" is a managed resource, while "the billing database instance" is a resource claim. By sharing or reusing at the resource claim level you're allowing multiple entities to use "the billing database". By reusing at the managed resource level you're allowing multiple entries to use "a big database instance", without consideration of the fact that the big database used to be the billing database instance.

Keep in mind the reclaim policy is set by the infrastructure operator, not the claimant, so the infrastructure operator is deciding what happens to it when it's done. The cases we support here are:

* `Delete` - No cleanup needed, just delete the thing and be done with it.
* `Retain` - I need or want to manually clean this thing up. Keep it around until I delete it manually. Maybe I want to ensure there's a backup of an SQL database, or want to retain a globally unique bucket name, or want to manually delete LoadBalancer type services from an EKS cluster so their ELBs are not orphaned when the cluster is deleted.

Persistent volumes support a deprecated `Recycle` policy that attempts to clean up a previously claimed PV and let others claim it. I'm guessing this was deprecated because it's difficult to securely erase volumes with arbitrary underlying storage technologies, and because there's just not that much point recycling a volume when you could instead treat it as cattle; delete it and dynamically provision an identical new one. I suspect these are both as or more true for our managed resources.

pkg/resource/resource.go
negz: I force pushed a change that attempts to resolve this by doubling down on the switch statement. ;)

muvaf:
> I don't think it is a safe and sensible default to then allow another entity to claim that managed resource for a different set of purposes, given that by doing so they'll potentially gain access to any state that was left over by the previous claimant.

I agree with this statement. It definitely shouldn't be the default. But I also see the value in making this possible. Opened https://github.com/crossplaneio/crossplane-runtime/issues/88 to track this discussion.

PR #89, opened 2019-11-22
Add integration testing package

hasheddan:

Moves integration testing framework from https://github.com/hasheddan/athodyd. For example implementation see https://github.com/crossplaneio/stack-gcp/pull/96.

### Checklist

I have:


pkg/test/integration/crd.go
```diff
@@ -0,0 +1,101 @@
...
+
+	docs := [][]byte{}
+	reader := k8syaml.NewYAMLReader(bufio.NewReader(bytes.NewReader(b)))
+	for {
+		// Read document
+		doc, err := reader.Read()
+		if err != nil {
+			if err == io.EOF {
+				break
+			}
```
negz: Nitpick: This error case does not need to be nested; it could be before the `err != nil` check.

pkg/test/integration/server.go
```diff
@@ -0,0 +1,202 @@
...
+package integration
+
+import (
+	"time"
+
+	"github.com/pkg/errors"
+	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	"k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+
+	// Allow auth to cloud providers
```
negz: I opened the https://godoc.org/k8s.io/client-go/plugin/pkg/client/auth trying to find more context on what it means to allow auth to cloud providers by importing a package, but could not find any. Perhaps it's worth adding a little more explanation here?

pkg/test/integration/server.go
```diff
@@ -0,0 +1,202 @@
...
+
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
+
+const (
+	syncPeriod = "30s"
+	errCleanup = "failure in athodyd default cleanup"
```
negz: Should this still refer to `athodyd`? There's one or two other references in this file.

pkg/test/integration/server.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	channel := make(chan struct{})
+	return &Manager{mgr, channel, e, client, c}, nil
+}
+
+// Run starts a controller-runtime manager with Athodyd's channel.
+func (m *Manager) Run() {
+	go func() {
+		if err := m.Start(m.channel); err != nil {
```
negz: Do you think it would make sense for `channel` to have a more explicit name? `channel` does not do much to communicate its purpose. Should we embed a stop channel at all, rather than relying on the caller to pass one in? I imagine there is a reason `envtest` was designed such that the caller passes in a channel.

pkg/test/integration/server.go
```diff
@@ -0,0 +1,202 @@
...
+// NewBuilder returns a new no-op Builder
+func NewBuilder() OperationFn {
+	return func(*envtest.Environment, client.Client) error {
+		return nil
+	}
+}
+
+// NewCleaner returns a new Cleaner that deletes all installed CRDs from the API
+// server.
+func NewCleaner() OperationFn {
```
negz: https://github.com/crossplaneio/crossplane/pull/1032#discussion_r346576526

Cross-linking this existing comment from the design doc.

pkg/test/integration/server.go
hasheddan: It just allows you to authenticate to managed k8s offerings on a variety of providers. The GCP plugin has a pretty good doc [here](https://github.com/kubernetes/client-go/blob/8d0e6f1b7b780ef8c8f81800d504509cd1f68e9f/plugin/pkg/client/auth/gcp/gcp.go#L58).

negz: That's good context, thanks! Would you mind capturing it in the comment in case any future readers are also unclear?

pkg/test/integration/server.go
hasheddan: I think calling it `stop` would be appropriate. `envtest` doesn't actually do any manager setup, so there is not channel being passed in. My initial thinking was that since this can only ever be an empty struct channel (a.k.a. a "signal channel") I don't see any enhancement from an end user perspective in letting someone pass in their own channel. However, if a user wanted to test stopping the manager (and thus any registered controllers), then continuing with any tests, it may be useful to have more fine-grained control of the channel. I don't see this as a major use-case (if resources want to be tested without controllers running then a separate test can be created that does not register the controllers), but I could be wrong when we start using this more. If it is all the same to you, I would prefer to implement in a later PR as needed :)

hasheddan:
@negz all changes have been addressed either with changes or comments, thanks for your feedback! RTM if looks good to you :)

pkg/test/integration/server.go
negz: I'd still prefer a clearer explanation here, but will merge to unblock this work.

hasheddan: Oops yes meant to add a more descriptive comment, will follow up!

PR #90, opened 2019-11-25
managed reconciler: add method ref to external client errors

### LINKED ISSUE #61: Managed reconciler should denote which external client function failed

Right now, in order for user to know in which step the error happened is through the error messages _inside_ the ExternalClient functions. However, that can be quite ambiguous. Think of the following case: `Update` call fails but it because of the `Get` call of the provider API in the `Update` function. Usually, developer puts an error message saying `Get`failed but the only way for them to say `Get` failed in the context of `Update` is to wrap a prefix error message. But this is already known by the managed reconciler, so, we can let the developer worry about only the error of the specific call and be sure that enough context is provided for debugging/tracing.

Instead of using `managed.SetConditions(v1alpha1.ReconcileError(err))` in all calls of the ExternalClient, we can use something like `managed.SetConditions(v1alpha1.ReconcileError(errors.Wrap(err, "<Create/Observe/Update/Delete> failed")))`

hasheddan:

Fixes #61 

### Checklist

I have:

### DIFF
--- a/pkg/resource/managed_reconciler.go
+++ b/pkg/resource/managed_reconciler.go
@@ -44,11 +44,11 @@ const (
 // Error strings.
 const (
 	errGetManaged       = "cannot get managed resource"
-	errReconcileConnect = "Connect failed"
-	errReconcileObserve = "Observe failed"
-	errReconcileCreate  = "Create failed"
-	errReconcileUpdate  = "Update failed"
-	errReconcileDelete  = "Delete failed"
+	errReconcileConnect = "connect failed"
+	errReconcileObserve = "observe failed"
+	errReconcileCreate  = "create failed"
+	errReconcileUpdate  = "update failed"
+	errReconcileDelete  = "delete failed"
 )
 
 // ConnectionDetails created or updated during an operation on an external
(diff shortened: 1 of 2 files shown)


pkg/resource/managed_reconciler.go
```diff
@@ -43,7 +43,12 @@ const (
 
 // Error strings.
 const (
-	errGetManaged = "cannot get managed resource"
+	errGetManaged       = "cannot get managed resource"
+	errReconcileConnect = "Connect failed"
+	errReconcileObserve = "Observe failed"
+	errReconcileCreate  = "Create failed"
+	errReconcileUpdate  = "Update failed"
+	errReconcileDelete  = "Delete failed"
```
negz: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

Do you think this satisfies the above rule? If we assume `Connect` etc are proper nouns, i.e. names of the method that failed, it does. If we think of them as generic actions it probably does not.

hasheddan: I would tend to think of them as generic actions, will update :)

hasheddan:
@negz error messages updated!

PR #91, opened 2019-12-03
integration tests: support downloading remote CRDs

hasheddan:

Adds the ability to specify a remote URL to download CRDs for use in integration tests. Example implementation in https://github.com/crossplaneio/stack-gcp/pull/96.

Note: `remote.go` is in a separate file in anticipation of further refinement. It currently has very little functionality on its own.

### Checklist

I have:


pkg/test/integration/server.go
```diff
@@ -33,8 +34,10 @@ import (
 )
 
 const (
-	syncPeriod = "30s"
-	errCleanup = "failure in default cleanup"
+	syncPeriod   = "30s"
+	errCleanup   = "failure in default cleanup"
+	errGetRemote = "unable to download CRDs from remote location"
+	remotePath   = "./tmp-test"
```
negz: Is this a temporary directory used to store downloaded remote CRDs?

Would it make sense to use https://golang.org/pkg/io/ioutil/#TempDir rather than a fixed directory? If not, could you add a comment explaining what this directory is for.

pkg/test/integration/server.go
```diff
@@ -82,6 +86,14 @@ func NewCRDCleaner() OperationFn {
 	}
 }
 
+// NewRemoteDirCleaner cleans up the default directory where remote CRDs were
+// downloaded.
+func NewRemoteDirCleaner() OperationFn {
+	return func(e *envtest.Environment, c client.Client) error {
+		return os.RemoveAll(remotePath)
```
negz: Should we assume we'll always be dealing with a real filesystem here, or would there be value in using something like https://github.com/spf13/afero to let us abstract that away?

pkg/test/integration/server.go
```diff
@@ -121,6 +140,7 @@ func defaultConfig() *Config {
 	}
 
 	return &Config{
+		RemoteCRDPaths:    []string{},
 		CRDDirectoryPaths: []string{},
```
negz: Do we need to make the distinction between remote and local CRD paths in our API? It seems like go-getter supports "getting" from the local filesystem.

hasheddan: We could keep them all the same and then just copy anything local into the tmp dir as well, then pass that path to `envtest`. The reasoning for keeping them separate was that if no remote paths are supplied then we won't make a new tmp dir or do any operations, we will go straight to installing the local CRDs.

pkg/test/integration/server.go
hasheddan: I really like `afero`, but as long as we are using `envtest` to install our CRDs (not entirely certain we will want to forever) we will be reliant on the fact that it uses the actual real filesystem.

pkg/test/integration/server.go
hasheddan: You are correct, and I like the idea! Pending decisions on comments below, I will implement this using `TempDir` instead :)

pkg/test/integration/server.go
negz: Got it, fair enough.

pkg/test/integration/server.go
negz: My preference would be for what I feel is the cleaner UX of only taking one set of directory paths, even if it means we do some extra, unnecessary operations when only local CRDs are used.

pkg/test/integration/remote.go
```diff
@@ -0,0 +1,50 @@
...
+
+// Downloads the given url into a subdirectory in the given path.
+func downloadPath(url, path string) (string, error) {
+	// Pwd is necessary for downloading local files via relative path.
+	pwd, err := os.Getwd()
+	if err != nil {
+		return "", err
+	}
+
+	// Subdirectory is given name of md5 hash of url.
```
negz: Does this _have_ to be an md5 hash of the URL, or is this just a convention we've established? If the latter I'd suggest using https://golang.org/pkg/hash/fnv/ rather than MD5, presuming this does not need to be a cryptographically sound hash.

pkg/test/integration/server.go
```diff
@@ -149,33 +164,54 @@ func New(cfg *rest.Config, o ...Option) (*Manager, error) {
 		op(c)
 	}
 
+	// Create temporary directory for CRDs.
+	dir, err := ioutil.TempDir(".", "")
```
negz: It's more typical in Unix environments to create temp dirs under the system temp dir (e.g. `/tmp`). I believe `TempDir` will do this if the dir argument is an empty string. Is there a reason we prefer to create the temp dir under the current directory?

pkg/test/integration/server.go
```diff
@@ -149,33 +164,54 @@ func New(cfg *rest.Config, o ...Option) (*Manager, error) {
...
+	// Create environment configuration.
 	e := &envtest.Environment{
-		CRDDirectoryPaths:  c.CRDDirectoryPaths,
+		CRDDirectoryPaths:  crdPaths,
 		Config:             cfg,
 		UseExistingCluster: &useExisting,
 	}
 
-	cfg, err := e.Start()
+	// Start environment.
```
negz: Nitpick: Do you think the comments in this `New` function are adding value? Many people (including myself) feel comments are best when they explain _why_ something is happening, not _what_ is happening. In this case the comments appear to be explaining what is happening, and I'm not sure they really need to given how readable the code is.

pkg/test/integration/remote.go
hasheddan: Nope! Was actually converting this to `fnv` when you commented :)

pkg/test/integration/server.go
hasheddan: Nope! Happy to remove them.

pkg/test/integration/server.go
hasheddan: I don't believe so. I was actually doing this originally for testing just to see it be created when the test ran.

PR #92, opened 2019-12-05
secret reconciler: allow for propagation to multiple secrets

hasheddan:

Allows for the secret reconciler to propagate a single `from` secret's data to multiple `to` secrets per https://github.com/crossplaneio/crossplane/pull/1101#discussion_r354505820.

TODO:
- [x] Handle cumulative errors in secret reconciler more elegantly
- [x] Add secret reconciler unit tests that check for multiple error handling

### Checklist

I have:

### DIFF
--- a/pkg/resource/api.go
+++ b/pkg/resource/api.go
@@ -18,7 +18,6 @@ package resource
 
 import (
 	"context"
-	"fmt"
 	"strings"
 
 	"github.com/pkg/errors"
@@ -126,7 +125,9 @@ func (a *APIManagedConnectionPropagator) PropagateConnection(ctx context.Context
 		}
 		cmcs.Data = mgcs.Data
 		meta.AddAnnotations(cmcs, map[string]string{
-			fmt.Sprintf(AnnotationKeyPropagateFromFormat, string(mgcs.GetUID())): strings.Join([]string{mgcs.GetNamespace(), mgcs.GetName()}, "/"),
+			AnnotationKeyPropagateFromNamespace: mgcs.GetNamespace(),
+			AnnotationKeyPropagateFromName:      mgcs.GetName(),
+			AnnotationKeyPropagateFromUID:       string(mgcs.GetUID()),
 		})
 		return nil
 	}); err != nil {
@@ -134,7 +135,7 @@ func (a *APIManagedConnectionPropagator) PropagateConnection(ctx context.Context
 	}
 
 	meta.AddAnnotations(mgcs, map[string]string{
-		fmt.Sprintf(AnnotationKeyPropagateToFormat, string(cmcs.GetUID())): strings.Join([]string{cmcs.GetNamespace(), cmcs.GetName()}, "/"),
+		strings.Join([]string{AnnotationKeyPropagateToPrefix, string(cmcs.GetUID())}, SlashDelimeter): strings.Join([]string{cmcs.GetNamespace(), cmcs.GetName()}, SlashDelimeter),
 	})
 
 	return errors.Wrap(a.client.Update(ctx, mgcs), errUpdateSecret)
--- a/pkg/resource/api.go
+++ b/pkg/resource/api.go
@@ -18,6 +18,8 @@ package resource
 
 import (
 	"context"
+	"fmt"
+	"strings"
 
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
@@ -124,9 +126,7 @@ func (a *APIManagedConnectionPropagator) PropagateConnection(ctx context.Context
 		}
 		cmcs.Data = mgcs.Data
 		meta.AddAnnotations(cmcs, map[string]string{
-			AnnotationKeyPropagateFromNamespace: mgcs.GetNamespace(),
-			AnnotationKeyPropagateFromName:      mgcs.GetName(),
-			AnnotationKeyPropagateFromUID:       string(mgcs.GetUID()),
+			fmt.Sprintf(AnnotationKeyPropagateFromFormat, string(mgcs.GetUID())): strings.Join([]string{mgcs.GetNamespace(), mgcs.GetName()}, "/"),
 		})
 		return nil
 	}); err != nil {
@@ -134,9 +134,7 @@ func (a *APIManagedConnectionPropagator) PropagateConnection(ctx context.Context
 	}
 
 	meta.AddAnnotations(mgcs, map[string]string{
-		AnnotationKeyPropagateToNamespace: cmcs.GetNamespace(),
-		AnnotationKeyPropagateToName:      cmcs.GetName(),
-		AnnotationKeyPropagateToUID:       string(cmcs.GetUID()),
+		fmt.Sprintf(AnnotationKeyPropagateToFormat, string(cmcs.GetUID())): strings.Join([]string{cmcs.GetNamespace(), cmcs.GetName()}, "/"),
 	})
 
 	return errors.Wrap(a.client.Update(ctx, mgcs), errUpdateSecret)
--- a/pkg/resource/enqueue_handlers.go
+++ b/pkg/resource/enqueue_handlers.go
@@ -106,25 +106,23 @@ func addPropagated(obj runtime.Object, queue adder) {
 	a := ao.GetAnnotations()
 
 	for key, val := range a {
-		if strings.HasPrefix(key, AnnotationKeyPropagateTo) {
-			t := strings.Split(val, "/")
-			if len(t) != 2 {
-				continue
-			}
-			toNamespace := t[0]
-			toName := t[1]
-
-			switch {
-			case toName == "":
-				continue
-			case toNamespace == "":
-				continue
-			default:
-				queue.Add(reconcile.Request{NamespacedName: types.NamespacedName{
-					Namespace: toNamespace,
-					Name:      toName,
-				}})
-			}
+		if !strings.HasPrefix(key, AnnotationKeyPropagateToPrefix) {
+			continue
+		}
+		t := strings.Split(val, SlashDelimeter)
+		if len(t) != 2 {
+			continue
+		}
+		switch {
+		case t[0] == "":
+			continue
+		case t[1] == "":
+			continue
+		default:
+			queue.Add(reconcile.Request{NamespacedName: types.NamespacedName{
+				Namespace: t[0],
+				Name:      t[1],
+			}})
 		}
 	}
 }
--- a/pkg/resource/enqueue_handlers.go
+++ b/pkg/resource/enqueue_handlers.go
@@ -17,6 +17,8 @@ limitations under the License.
 package resource
 
 import (
+	"strings"
+
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/workqueue"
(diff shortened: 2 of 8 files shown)


pkg/resource/secret_reconciler.go
```diff
@@ -27,15 +27,16 @@ import (
...
 // Supported resources with all of these annotations will be fully or partially
 // propagated to the named resource of the same kind, assuming it exists and
 // consents to propagation.
 const (
-	AnnotationKeyPropagateToNamespace = "crossplane.io/propagate-to-namespace"
-	AnnotationKeyPropagateToName      = "crossplane.io/propagate-to-name"
-	AnnotationKeyPropagateToUID       = "crossplane.io/propagate-to-uid"
+	AnnotationKeyPropagateToNamespaces = "crossplane.io/propagate-to-namespace"
+	AnnotationKeyPropagateToNames      = "crossplane.io/propagate-to-name"
+	AnnotationKeyPropagateToUIDs       = "crossplane.io/propagate-to-uid"
```
negz: I wonder if instead of a comma separated list we could encode the different UIDs in the annotation keys. The namespace and name are really just there for human readability, so maybe something like:

`crossplane.io/propagate-to/668cd01a-17ab-11ea-8d71-362b9e155667: default/coolsecret`

Or maybe:

`to.propagate.crossplane.io/668cd01a-17ab-11ea-8d71-362b9e155667: default/coolsecret`

hasheddan: I actually experimented with this a little bit and ended up finding it a little cleaner for a few reasons related to [annotation structure](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set):
- Only one slash can be used to separate the `prefix` from the `name` (we could obviously utilize another character for the separations, but it might look a little weird)
- The `name` is limited to 63 characters max (`668cd01a-17ab-11ea-8d71-362b9e155667:default/coolsecret` is already 55 chars so I think we could hit this quickly`)

negz: > I actually experimented with this a little bit and ended up finding it a little cleaner for a few reasons related to annotation structure

Do you mean it's a little clearer to stick with your proposed comma separated pattern?

hasheddan: Yep, that's what I was indicating. Do you feel strongly in favor of the putting the secret in the annotation keys?

pkg/resource/secret_reconciler.go
```diff
@@ -78,43 +80,65 @@ func NewSecretPropagatingReconciler(m manager.Manager) reconcile.Reconciler {
...
-			Namespace: from.GetAnnotations()[AnnotationKeyPropagateToNamespace],
-			Name:      from.GetAnnotations()[AnnotationKeyPropagateToName],
+		fromAnnotations := from.GetAnnotations()
+
+		uids := util.Split(fromAnnotations[AnnotationKeyPropagateToUIDs], ",")
+		names := util.Split(fromAnnotations[AnnotationKeyPropagateToNames], ",")
+		namespaces := util.Split(fromAnnotations[AnnotationKeyPropagateToNamespaces], ",")
+		propagationErrors := []error{}
+
+		if len(uids) == len(names) && len(names) == len(namespaces) {
```
negz: If we do stick with this comma separated annotations pattern I would prefer we returned early in the case the annotations were incongruent, rather than nesting the bulk of the logic of this controller inside this if statement.

negz:
> It's not immediately clear to me what such a solution would look like.

One option may be to have each propagation controller use its own set of annotation keys. This way the controller is still concerned with a 1:1 secret propagation pattern.

As part of this PR we'll also need to consider how these annotations are set. Currently the annotations are added by the `PropagateConnection` method, and we have no concept of annotations being removed.

https://github.com/crossplaneio/crossplane-runtime/blob/e02208e/pkg/resource/api.go#L92

pkg/util/errors.go
```diff
@@ -0,0 +1,39 @@
+/*
```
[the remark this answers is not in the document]
hasheddan: Sure thing, thanks for pointing this out!

hasheddan:
> One option may be to have each propagation controller use its own set of annotation keys. This way the controller is still concerned with a 1:1 secret propagation pattern.

I'm not sure this solves our problem because that single controller could (and likely would) still be propagating a single secret to multiple namespaces. For instance, there will be one of these types of controllers for `GKECluster` and the reason for this PR is that a single `GKECluster` connection secret is desired to be propagated to multiple namespaces.

negz:
> I'm not sure this solves our problem 

Ah, you're right. I forgot that we needed this one to many pattern for multiple `KubernetesTarget` resources pointing to one `GKECluster`, not just because we might need to propagate a secret to a `KubernetesTarget` and a `KubernetesCluster`.

pkg/resource/secret_reconciler.go
negz: So I did a little more reading about this. My read is that that the annotation key is of the format `optional-prefix/name`. The _name_ is limited to 63 characters, but the prefix is limited to 253. So I think something like `to.propagate.crossplane.io` would work.

https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/

I also think that if we switch the controller and predicate around to queue the reconcile for the _destination_ rather than the source secret we can avoid iterating over all of the annotations of the destination and instead:

1. Read the annotations of the destination secret.
1. Grab its source secret, and check whether the destination secret's UUID exists in its annotation keys.
1. If all seems well, copy the source data to the destination.

pkg/resource/api.go
```diff
@@ -124,19 +126,15 @@ func (a *APIManagedConnectionPropagator) PropagateConnection(ctx context.Context
 		}
 		cmcs.Data = mgcs.Data
 		meta.AddAnnotations(cmcs, map[string]string{
-			AnnotationKeyPropagateFromNamespace: mgcs.GetNamespace(),
-			AnnotationKeyPropagateFromName:      mgcs.GetName(),
-			AnnotationKeyPropagateFromUID:       string(mgcs.GetUID()),
+			fmt.Sprintf(AnnotationKeyPropagateFromFormat, string(mgcs.GetUID())): strings.Join([]string{mgcs.GetNamespace(), mgcs.GetName()}, "/"),
```
hasheddan: Note: we do not have any mechanism that restricts a propagated `Secret` to only be propagated from one source. We could theoretically encounter a situation where a `Secret` had multiple `from.propagate.crossplane.io/` annotations and controllers could fight to update the secret's data. We could check here that the `Secret` does not already have a propagate from annotation and refuse to operate on it if so. We could also remove any existing and replace with the new annotation. Otherwise, we could just let this unexpected scenario happen and leave it to a user to fix the conflict.

pkg/resource/secret_reconciler.go
```diff
@@ -68,53 +68,85 @@ func NewSecretPropagatingReconciler(m manager.Manager) reconcile.Reconciler {
...
-			// The propagating secret expected a different propagated secret. We
+		toAnnotations := to.GetAnnotations()
+		var fromUID string
+		var fromName string
+		var fromNamespace string
+
+		for key, val := range toAnnotations {
+			// If this is a 'from' secret we do not want to reconcile and will
+			// instead defer to the enqueued request for the 'to' secret
+			if strings.HasPrefix(key, AnnotationKeyPropagateTo) {
```
hasheddan: Is there a better way to do this?

pkg/resource/api.go
hasheddan: Yes! I was thinking and I believe there are two steps to ensuring this:

1. In `Secret` propagation here we could refuse to add from annotations to a `Secret` that already has them. This is the only place where the annotations are actually set so unless annotations were manually applied by a user, this would restrict from ever having more than one set. Alternatively, we could remove any existing from annotations here and replace them with the desired set. However, this could cause issues if two controllers are both reconciling and removing each others from annotations over and over.
2. We could refuse to operate on `Secrets` that have more than one set of from annotations in the secret propagating reconciler, essentially refusing to fight over updating a `Secret`. This would mean that any `Secret` that somehow obtained multiple sets of from annotations would not have data propagated from either. Because we are restricting the ability to add multiple sets of from annotations via propagation in (1) this situation would only be encountered if a user manually add additional from annotations.

Thoughts?

negz: This sounds good to me. Perhaps given we'll only ever allow one propagate-from annotation we should maintain the existing annotation style in which the from UID is the value, not part of the key? I could go either way - I think the existing pattern would make it more obvious that [there can be only one](https://www.youtube.com/watch?v=_J3VeogFUOs) from annotation, but I it would make the to and from annotations asymmetrical.

pkg/resource/api.go
```diff
@@ -124,19 +126,15 @@ func (a *APIManagedConnectionPropagator) PropagateConnection(ctx context.Context
...
 		return nil
 	}); err != nil {
 		return errors.Wrap(err, errCreateOrUpdateSecret)
 	}
 
 	meta.AddAnnotations(mgcs, map[string]string{
-		AnnotationKeyPropagateToNamespace: cmcs.GetNamespace(),
-		AnnotationKeyPropagateToName:      cmcs.GetName(),
-		AnnotationKeyPropagateToUID:       string(cmcs.GetUID()),
+		fmt.Sprintf(AnnotationKeyPropagateToFormat, string(cmcs.GetUID())): strings.Join([]string{cmcs.GetNamespace(), cmcs.GetName()}, "/"),
```
negz: Perhaps the `/` separator should be a constant, given it's used across at least two files.

pkg/resource/enqueue_handlers.go
```diff
@@ -65,52 +67,64 @@ func addClaim(obj runtime.Object, queue adder) {
...
+			switch {
+			case toName == "":
+				continue
+			case toNamespace == "":
+				continue
+			default:
+				queue.Add(reconcile.Request{NamespacedName: types.NamespacedName{
+					Namespace: toNamespace,
+					Name:      toName,
+				}})
```
negz: Nit: You could probably skip the `toNamespace` and `toName` vars and instead go straight to a `types.NamespacedName`, e.g. `nn := types.NamespacedName{Namespace: t[0], Name: t[1]}`.

pkg/resource/enqueue_handlers.go
```diff
@@ -65,52 +67,64 @@ func addClaim(obj runtime.Object, queue adder) {
...
-	case a[AnnotationKeyPropagateFromName] == "":
-		return
-	default:
-		queue.Add(reconcile.Request{NamespacedName: types.NamespacedName{
-			Namespace: a[AnnotationKeyPropagateFromNamespace],
-			Name:      a[AnnotationKeyPropagateFromName],
-		}})
+
+	for key, val := range a {
+		if strings.HasPrefix(key, AnnotationKeyPropagateTo) {
```
negz: Nit: In cases like this where there's little to no code outside the conditional I like to flip it to avoid the extra layer of nesting in the happy path, e.g.:

```go
for key, val := range a {
    if !strings.HasPrefix(key, AnnotationKeyPropagateTo) {
        continue
    }
    // Logic goes here.
}

File: pkg/resource/predicates.go [hunk 9]
```diff
@@ -135,16 +137,13 @@ func IsPropagator() PredicateFn {
...
-		switch {
-		case a[AnnotationKeyPropagateToNamespace] == "":
-			return false
-		case a[AnnotationKeyPropagateToName] == "":
-			return false
-		case a[AnnotationKeyPropagateToUID] == "":
-			return false
-		default:
-			return true
+		for key := range a {
```
negz: Nit: You could probably iterate over `ao.GetAnnotations()` directly now that this function uses it only once.

pkg/resource/secret_reconciler.go
```diff
@@ -33,33 +35,31 @@ import (
...
-// Supported resources with all of these annotations consent to be fully or
-// partially propagated from the named resource of the same kind.
-const (
-	AnnotationKeyPropagateFromNamespace = "crossplane.io/propagate-from-namespace"
-	AnnotationKeyPropagateFromName      = "crossplane.io/propagate-from-name"
-	AnnotationKeyPropagateFromUID       = "crossplane.io/propagate-from-uid"
+	AnnotationKeyPropagateTo         = "to.propagate.crossplane.io/"
+	AnnotationKeyPropagateToFormat   = "to.propagate.crossplane.io/%s"
+	AnnotationKeyPropagateFrom       = "from.propagate.crossplane.io/"
+	AnnotationKeyPropagateFromFormat = "from.propagate.crossplane.io/%s"
```
negz: I notice you're using `fmt.Sprintf` for annotation keys, and `strings.Join` for annotation values. Perhaps we should settle on one or the other? My vote would be for `strings.Join`, given we already use `AnnotationKeyPropagateTo` and `AnnotationKeyPropagateFrom` to test for prefix matches.

Also, perhaps these should be `AnnotationKeyPrefixPropagateTo` (and `From`) to clarify that they're a prefix?

pkg/resource/api.go
hasheddan: I agree that we should go with the existing pattern. I like that it is asymmetrical with the to annotations because they behave differently :)

pkg/resource/secret_reconciler.go
```diff
@@ -68,53 +68,85 @@ func NewSecretPropagatingReconciler(m manager.Manager) reconcile.Reconciler {
...
+				fromUID = strings.TrimPrefix(key, AnnotationKeyPropagateFrom)
+				f := strings.Split(val, "/")
+				if len(f) != 2 {
+					return reconcile.Result{}, errors.New(errInvalidPropagateFormat)
+				}
+				fromNamespace = f[0]
+				fromName = f[1]
+				break
+			}
+		}
```
negz: I think we could simplify this logic quite a lot if we used something similar to the contemporary `from` annotation, perhaps:

```
from.propagate.crossplane.io/uid: 8aedda70-2e77-11ea-978f-2e728ce88125
from.propagate.crossplane.io/namespace: default
from.propagate.crossplane.io/name: coolsecret 
```

Alternatively if we wanted to keep our annotations symmetrical (even though in practice we'd only support propagating from a single secret) then your suggestion of bailing out early if there is more than one from annotation would also allow us to simplify this block.

pkg/resource/secret_reconciler.go
negz: Do we need to consider the case in which a secret is both propagated and a propagator? I think that would work (by accident) in the contemporary implementation, though you'd have to manually write annotations for it to ever happen.

hasheddan: Hmm I was thinking about this and I believe you are correct that this could only happen with manual annotations. So then I was thinking of whether this is something we should allow or not. The purpose of the secret reconciler is to propagate data from one secret to another when they both consent to the process. Technically a chain of secrets that do this should still be able to be handled by this reconciler as long as each consents to propagation with the next and preceding secrets. While I am not sure of a use-case for this, I am leaning towards allowing it in the context of this reconciler. What do you think?

pkg/resource/api.go
hasheddan: Using the existing pattern of `from` annotations actually solves both (1) and (2) above :)

pkg/resource/secret_reconciler.go
negz: Yeah, given it doesn't seem to a a risk I'd lean toward allowing it, especially if it simplifies the code.

pkg/resource/secret_reconciler.go
```diff
@@ -68,53 +68,51 @@ func NewSecretPropagatingReconciler(m manager.Manager) reconcile.Reconciler {
...
 			// There's no propagation to be done if the secret we propagate to
 			// does not exist. We assume we have a watch on that secret and will
 			// be queued if/when it is created. Otherwise we'll be requeued
 			// implicitly because we return an error.
 			return reconcile.Result{}, errors.Wrap(IgnoreNotFound(err), errGetSecret)
 		}
+		// The 'from' secret is also know as the 'propagating' secret.
+		from := &corev1.Secret{}
+		n := types.NamespacedName{
+			Namespace: to.GetAnnotations()[AnnotationKeyPropagateFromNamespace],
```
hasheddan: I have removed the "invalid propagation format" errors in favor of the existing pattern of just checking whether we can get the from `Secret` using what is provided.

pkg/resource/secret_reconciler.go
```diff
@@ -33,14 +34,9 @@ import (
...
-)
+	AnnotationKeyPropagateToPrefix = "to.propagate.crossplane.io"
+	SlashDelimeter                 = "/"
 
-// Supported resources with all of these annotations consent to be fully or
-// partially propagated from the named resource of the same kind.
-const (
 	AnnotationKeyPropagateFromNamespace = "crossplane.io/propagate-from-namespace"
 	AnnotationKeyPropagateFromName      = "crossplane.io/propagate-from-name"
 	AnnotationKeyPropagateFromUID       = "crossplane.io/propagate-from-uid"
```
negz: Could we make these `propagate.crossplane.io/from-namespace`? This way both ends of our propagation annotations are within the `propagate.crossplane.io` annotation domain.

hasheddan: Ah yes definitely. Given that we have `to.propagate.crossplane.io` should we do `from.propagate.crossplane.io/namespace`, etc.?

negz: Yeah, that seems sensible.

hasheddan: Awesome, done!

PR #93, opened 2019-12-06
Add Provider interface

muvaf:

`Provider` object is used in all calls made to the external providers. Because of that its lifecycle differs from others. Stack who manage Crossplane resources currently have no way of knowing whether an object is a provider unless they import specific infrastructure stack module, which is not practical due to added dependencies.

The case I needed this is that I'm working on a generic configurations stack reconciler and `Provider`s have to be deleted last as opposed to other resources because their deletion fails if `Provider` is gone. However, since it has to be generic, identifying whether a resource is a `Provider` one has to import every infra stack and check the object's `GroupVersionKind`. This interface will make it much easier to identify a `Provider` type resource in a generic way.

### Checklist

I have:

### DIFF
--- a/pkg/resource/interfaces.go
+++ b/pkg/resource/interfaces.go
@@ -82,6 +82,13 @@ type Reclaimer interface {
 	GetReclaimPolicy() v1alpha1.ReclaimPolicy
 }
 
+// A CredentialsSecretReferencer may refer to a credential secret in an arbitrary
+// namespace.
+type CredentialsSecretReferencer interface {
+	GetCredentialsSecretReference() *v1alpha1.SecretKeySelector
+	SetCredentialsSecretReference(r *v1alpha1.SecretKeySelector)
+}
+
 // A Claim is a Kubernetes object representing an abstract resource claim (e.g.
 // an SQL database) that may be bound to a concrete managed resource (e.g. a
 // CloudSQL instance).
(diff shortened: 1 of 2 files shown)


muvaf:
A little bit more context. There will be a repository with generic configuration stack reconciler in which we'll filter out the resources who implements `Provider` interface. The actual configuration stacks that will use the reconciler will supply the `runtime.scheme` object to the reconciler, which will make it possible for the reconciler to construct the correct objects and know the functions of that specific `Provider` type, hence the filtering will work.

pkg/resource/interfaces.go
```diff
@@ -82,6 +82,13 @@ type Reclaimer interface {
 	GetReclaimPolicy() v1alpha1.ReclaimPolicy
 }
 
+// A CredentialsSecretReferencer may refer to a credential secret to an arbitrary
+// namespace.
+type CredentialsSecretReferencer interface {
```
hasheddan: Since these methods are not currently present on existing `Provider` types, do you plan to implement them in each stack? We should probably generate them using `crossplane-tools/angryjet`

muvaf: Yes, after getting a thumbs-up from you guys I plan to implement them in `crossplane-tools/angryjet`

hasheddan: Nice!

PR #99, opened 2019-12-30
make wording in conditions cover resources other than managed ones

muvaf:

Conditions are not useful only to managed resource, but for all types of resources we have. They are already used in other resource types but the wording of the `reason`s confuses people. This PR removes `managed` from all of the conditions and code comments.

### Checklist

I have:

### DIFF
--- a/apis/core/v1alpha1/condition.go
+++ b/apis/core/v1alpha1/condition.go
@@ -28,14 +28,14 @@ type ConditionType string
 
 // Condition types.
 const (
-	// TypeReady managed resources are believed to be ready to handle work.
+	// TypeReady resources are believed to be ready to handle work.
 	TypeReady ConditionType = "Ready"
 
-	// TypeSynced managed resources are believed to be in sync with the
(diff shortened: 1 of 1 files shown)


apis/core/v1alpha1/condition.go
```diff
@@ -262,7 +261,7 @@ func ReconcileError(err error) Condition {
 }
 
 // ReferenceResolutionSuccess returns a condition indicating that Crossplane
-// successfully resolved the references used in the managed resource.
+// successfully resolved the references used in the resource.
```
hasheddan: Do we use CRR in app stacks? I guess it makes sense to make this not specific to managed resources even if we don't.

muvaf: Yes, we will use it here https://github.com/crossplaneio/sample-stack-wordpress/pull/20

But I didn't open this PR just for that, I think `condition`s have become our way of communicating the state now (similar to k8s) and wording needs to be adjusted for that.

PR #102, opened 2020-01-02
managed reconciler: status update has its own context

### LINKED ISSUE #98: Context deadline exceed error does not appear on status
### What happened?

I have a resource whose queries apparently exceeds the context deadline. However, I was only able to find that out through the logs of the controller, there is no information on custom resource status. The pod log:
```
{"level":"error","ts":1577272335.9786623,"logger":"controller-runtime.controller","msg":"Reconciler error","controller":"mysqlservervirtualnetworkrule.database.azure.crossplane.io","request":"/sample-vnet-rule","error":"cannot update managed resource status: context deadline exceeded","errorVerbose":"context deadline exceeded\ncannot update managed resource status\ngithub.com/crossplaneio/crossplane-runtime/pkg/resource.(*ManagedReconciler).Reconcile\n\t/home/upbound/go/pkg/mod/github.com/crossplaneio/crossplane-runtime@v0.2.4-0.20191220203046-1ab588031d3f/pkg/resource/managed_reconciler.go:568\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/home/upbound/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.4.0/pkg/internal/controller/controller.go:256\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/home/upbound/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.4.0/pkg/internal/controller/controller.go:232\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker\n\t/home/upbound/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.4.0/pkg/internal/controller/controller.go:211\nk8s.io/apimachinery/pkg/util/wait.JitterUntil.func1\n\t/home/upbound/go/pkg/mod/k8s.io/apimachinery@v0.0.0-20190913080033-27d36303b655/pkg/util/wait/wait.go:152\nk8s.io/apimachinery/pkg/util/wait.JitterUntil\n\t/home/upbound/go/pkg/mod/k8s.io/apimachinery@v0.0.0-20190913080033-27d36303b655/pkg/util/wait/wait.go:153\nk8s.io/apimachinery/pkg/util/wait.Until\n\t/home/upbound/go/pkg/mod/k8s.io/apimachinery@v0.0.0-20190913080033-27d36303b655/pkg/util/wait/wait.go:88\nruntime.goexit\n\t/usr/local/go/src/runtime/asm_amd64.s:1337","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/home/upbound/go/pkg/mod/github.com/go-logr/zapr@v0.1.1/zapr.go:128\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/home/upbound/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.4.0/pkg/internal/controller/controller.go:258\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/home/upbound/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.4.0/pkg/internal/controller/controller.go:232\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker\n\t/home/upbound/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.4.0/pkg/internal/controller/controller.go:211\nk8s.io/apimachinery/pkg/util/wait.JitterUntil.func1\n\t/home/upbound/go/pkg/mod/k8s.io/apimachinery@v0.0.0-20190913080033-27d36303b655/pkg/util/wait/wait.go:152\nk8s.io/apimachinery/pkg/util/wait.JitterUntil\n\t/home/upbound/go/pkg/mod/k8s.io/apimachinery@v0.0.0-20190913080033-27d36303b655/pkg/util/wait/wait.go:153\nk8s.io/apimachinery/pkg/util/wait.Until\n\t/home/upbound/go/pkg/mod/k8s.io/apimachinery@v0.0.0-20190913080033-27d36303b655/pkg/util/wait/wait.go:88"}
```

I think, `cr.Status().Update()` function calls should use `context.Background()` instead of the current context because the reconciliation context is related to specific operations while status updates report to the user. We could desire not to keep going with the operations after a deadline is met, but reporting to user is not something we may want to skip in any case. However, there is the risk of timout for `cr.Status().Update()` calls but that's fairly lower than the total of all operations that are conducted in one reconciliation pass.

### How can we reproduce it?

In my case, it happened with `MySQLServerVirtualNetworkRule` resource but I can't say it will happen every time. One way could be to somehow throttle your network connection.

Crossplane version: 0.6.0

muvaf:

The status update call should have its own context in order to be able to report error in the case where the main context's deadline is exceeded.

Fixes #98 

### Checklist

I have:

### DIFF
--- a/pkg/reconciler/managed/reconciler.go
+++ b/pkg/reconciler/managed/reconciler.go
@@ -35,8 +35,9 @@ import (
 )
 
 const (
-	managedFinalizerName    = "finalizer.managedresource.crossplane.io"
-	managedReconcileTimeout = 1 * time.Minute
+	managedFinalizerName = "finalizer.managedresource.crossplane.io"
+	reconcileGracePeriod = 30 * time.Second
+	reconcileTimeout     = 1 * time.Minute
 
 	defaultManagedShortWait = 30 * time.Second
 	defaultManagedLongWait  = 1 * time.Minute
@@ -317,6 +318,7 @@ type Reconciler struct {
 
 	shortWait time.Duration
 	longWait  time.Duration
+	timeout   time.Duration
 
 	// The below structs embed the set of interfaces used to implement the
 	// managed resource reconciler. We do this primarily for readability, so
@@ -357,6 +359,16 @@ func defaultMRExternal() mrExternal {
 
 // A ReconcilerOption configures a Reconciler.
 type ReconcilerOption func(*Reconciler)
+
+// WithTimeout specifies the timeout duration cumulatively for all the calls happen
+// in the reconciliation function. In case the deadline exceeds, reconciler will
+// still have some time to make the necessary calls to report the error such as
+// status update.
+func WithTimeout(duration time.Duration) ReconcilerOption {
+	return func(r *Reconciler) {
+		r.timeout = duration
+	}
+}
 
 // WithShortWait specifies how long the Reconciler should wait before queueing a
 // new reconciliation in 'short wait' scenarios. The Reconciler requeues after a
@@ -454,6 +466,7 @@ func NewReconciler(m manager.Manager, of resource.ManagedKind, o ...ReconcilerOp
 		newManaged: nm,
 		shortWait:  defaultManagedShortWait,
 		longWait:   defaultManagedLongWait,
+		timeout:    reconcileTimeout,
 		managed:    defaultMRManaged(m),
 		external:   defaultMRExternal(),
 		log:        logging.NewNopLogger(),
@@ -475,8 +488,13 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error) 
 	log := r.log.WithValues("request", req)
 	log.Debug("Reconciling")
 
-	ctx, cancel := context.WithTimeout(context.Background(), managedReconcileTimeout)
+	ctx, cancel := context.WithTimeout(context.Background(), r.timeout+reconcileGracePeriod)
 	defer cancel()
+
+	// Govet linter has a check for lost cancel funcs but it's a false positive
+	// for child contexts as because parent's cancel is called, so we skip it
+	// for this line.
+	externalCtx, _ := context.WithTimeout(ctx, r.timeout) // nolint:govet
 
 	managed := r.newManaged()
 	if err := r.client.Get(ctx, req.NamespacedName, managed); err != nil {
@@ -493,7 +511,7 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error) 
 		"external-name", meta.GetExternalName(managed),
 	)
 
-	external, err := r.external.Connect(ctx, managed)
+	external, err := r.external.Connect(externalCtx, managed)
 	if err != nil {
 		// We'll usually hit this case if our Provider or its secret are missing
 		// or invalid. If this is first time we encounter this issue we'll be
@@ -543,7 +561,7 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error) 
 		managed.SetConditions(v1alpha1.ReferenceResolutionSuccess())
 	}
 
-	observation, err := external.Observe(ctx, managed)
+	observation, err := external.Observe(externalCtx, managed)
 	if err != nil {
 		// We'll usually hit this case if our Provider credentials are invalid
 		// or insufficient for observing the external resource type we're
@@ -560,7 +578,7 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error) 
 		log = log.WithValues("deletion-timestamp", managed.GetDeletionTimestamp())
 
 		if observation.ResourceExists && managed.GetReclaimPolicy() == v1alpha1.ReclaimDelete {
-			if err := external.Delete(ctx, managed); err != nil {
+			if err := external.Delete(externalCtx, managed); err != nil {
 				// We'll hit this condition if we can't delete our external
 				// resource, for example if our provider credentials don't have
 				// access to delete it. If this is the first time we encounter this
@@ -631,7 +649,7 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error) 
 	}
 
 	if !observation.ResourceExists {
-		creation, err := external.Create(ctx, managed)
+		creation, err := external.Create(externalCtx, managed)
 		if err != nil {
 			// We'll hit this condition if we can't create our external
 			// resource, for example if our provider credentials don't have
@@ -676,7 +694,7 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error) 
 		return reconcile.Result{RequeueAfter: r.longWait}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
 	}
 
-	update, err := external.Update(ctx, managed)
+	update, err := external.Update(externalCtx, managed)
 	if err != nil {
 		// We'll hit this condition if we can't update our external resource,
 		// for example if our provider credentials don't have access to update


pkg/resource/managed_reconciler.go
```diff
@@ -432,6 +432,10 @@ func (r *ManagedReconciler) Reconcile(req reconcile.Request) (reconcile.Result,
 	ctx, cancel := context.WithTimeout(context.Background(), managedReconcileTimeout)
 	defer cancel()
 
+	// The status update call should have its own context in order to be able to
+	// report error in the case where the main context's deadline is exceeded.
+	updateCtx := context.Background()
```
hasheddan: I am not sure that we want to use a context that will never timeout. In my mind it seems like it would make sense to say if our managed context timed out that we should use a new context with a relatively short timeout to try and update that status to reflect that, but it feels like an anti-pattern to use a context with no timeout in response to a previous context timing out.

negz:
This is a tricky one. Currently we have one context, which I think of as the context for the entire reconcile loop. This means its deadline is for both cloud provider and Kubernetes API calls. The downside of this pattern is that a timeout at the cloud provider will prevent us from making the Kubernetes API calls necessary to let a user know we've timed out elsewhere in the system. To be honest I'm not sure we should go to the trouble of keeping a second context for this specific case, but if we do I agree with @hasheddan that it should have its own timeout.

negz:
Another option here (that is probably overkill) would be to create a separate goroutine that handles context timeouts - i.e. waits for the `ctx.Done()` channel to close, and does the needful if `ctx.Err()` is `DeadlineExceeded` - https://golang.org/pkg/context/#Context

muvaf:
> Another option here (that is probably overkill) would be to create a separate goroutine that handles context timeouts.

I think this will make the code harder to read to some extent.

I agree with the need for timeout but it starts to count when the context is created, so we have to create the context right when we start the operation. How about something like this:
```
// The status update calls should have their own context in order to be able to
// report error in the case where the main context's deadline is exceeded.
var updateCtx = func() context.Context {
	ctx, _ := context.WithTimeout(context.Background(), managedReconcileTimeout)
	return ctx
}

func (r *ManagedReconciler) Reconcile(req reconcile.Request) (reconcile.Result, error) { // nolint:gocyclo
....
return reconcile.Result{RequeueAfter: r.shortWait}, errors.Wrap(r.client.Status().Update(updateCtx(), managed), errUpdateManagedStatus)
```

hasheddan:
> I think this will make the code harder to read to some extent.

@muvaf I appreciate the attention to this, but I also think that `crossplane-runtime` is a place where we can handle things elegantly even if it is less readable for someone new to the language. The vast majority of people will be consumers of `crossplane-runtime` who don't necessarily have to understand a detail like this that happens behind the scenes.

All of that being said, if we do not want to introduce new explicit goroutines, your solution seems straightforward to me :)

negz:
What if instead of using a completely separate context we used a child context with a slightly smaller timeout for our cloud provider calls?

I think we should have one context per "request" that our system handles, and in this context I think that means one context per reconcile loop. Currently we set one timeout for that entire loop, meaning we have a certain amount of time (a minute iirc?) to get the resource under reconciliation from the API server, then make 1-2 calls to the cloud provider, then make 1-2 calls to the Kubernetes API server to reflect any changes we made. My feeling is that this PR is probably most useful in the case where we timeout because the cloud provider did not respond quickly enough; if we're timing out because the API server is responding too slowly it's likely that our attempt to update our status (by calling the API server) will also time out.

So if we were to create a context with a timeout of ~60 seconds for the entire reconcile, but use a child context with a timeout of ~45 seconds for the cloud provider API calls, we'd hopefully have 10-15 seconds left of our parent context to update the status if/when we timeout due to the cloud provider being unresponsive.

muvaf:
> So if we were to create a context with a timeout of ~60 seconds for the entire reconcile, but use a child context with a timeout of ~45 seconds for the cloud provider API calls, we'd hopefully have 10-15 seconds left of our parent context to update the status if/when we timeout due to the cloud provider being unresponsive.

@negz Yeah, that's pretty much what we want. I'll look into how I can do that and update this PR. Thanks for the suggestion!

muvaf:
I have tested this by putting `time.Sleep(45*time.Second)` in an `Observe` call of a managed resource and confirmed that `context deadline exceeded` returned from that call appeared as condition.

pkg/reconciler/managed/reconciler.go
```diff
@@ -478,6 +479,9 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
 	ctx, cancel := context.WithTimeout(context.Background(), managedReconcileTimeout)
 	defer cancel()
 
+	externalCtx, cancel := context.WithTimeout(ctx, externalClientTimeout)
+	defer cancel()
```
negz: I wonder whether we should be setting this here, or letting the ExternalConnector and ExternalClient implementations create their own child contexts with appropriate timeouts. What do you think?

muvaf: Letting the developer configure its APIs timeout sounds reasonable. But it should somehow be in sync with our reconciler context because in case they chose 2 minutes and our reconciler timeout is 1 minute, it doesn't make much sense. In cases where the normal response is >1 minute, even though the developer declared that, reconciliation would fail. Since it's a bit harder to figure out the chosen timeout from the context, it makes more sense to me that they just specify a timeout if they need to.

We can possibly allow configuration of the timeout via an `option`, which also affects the reconciler timeout. The defaults would be in place, but if they choose 90 seconds for API timeout, we'd add 30 seconds to that and say total reconciler timeout is 120 seconds while context passed to external client calls have 90 seconds timeout. So, we always make sure we're able to make our own calls. Just updated the PR to show what I mean.

pkg/reconciler/managed/reconciler.go
```diff
@@ -475,7 +488,10 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
 	log := r.log.WithValues("request", req)
 	log.Debug("Reconciling")
 
-	ctx, cancel := context.WithTimeout(context.Background(), managedReconcileTimeout)
+	ctx, cancel := context.WithTimeout(context.Background(), r.timeout+reconcileTimeoutAddition)
+	defer cancel()
```
negz: Do we need to call `cancel()` twice - won't cancelling this parent cancel the child?

pkg/reconciler/managed/reconciler.go
```diff
@@ -35,8 +35,9 @@ import (
 )
 
 const (
-	managedFinalizerName    = "finalizer.managedresource.crossplane.io"
-	managedReconcileTimeout = 1 * time.Minute
+	managedFinalizerName     = "finalizer.managedresource.crossplane.io"
+	reconcileTimeoutAddition = 30 * time.Second
```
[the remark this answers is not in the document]
muvaf: Grace period sounds better.

> Perhaps it should be configurable?

I'm not sure whether it should be configurable by provider stack developer. The commands that will be called in this period are against the api-server and that's not known about by the developer of the provider stack, i.e. external timeout is more relevant to them since that doesn't change, you always query the same entity but they don't know the quality of the k8s you're running their stack. Maybe it'd make sense to get it from an env var or flag to cover the case where api-server is so slow that the person who's running Crossplane wants to make it longer. Though I'm not sure whether it's worth adding a new flag.

pkg/reconciler/managed/reconciler.go
muvaf: No, we don't. It seems that it's sufficient to call parent's cancel to cancel everything. I had to add a `nolint` though because govet gives false positive about not calling cancel of a context, details here https://github.com/golang/go/issues/29587

muvaf:
@negz I think this should be ready to merge.

PR #103, opened 2020-01-03
Implement Target interface and reconciler

hasheddan:

This PR implements that [shared reconciler](https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-consuming-k8s-clusters.md#kubernetestarget) outlined in the design for the `KubernetesTarget` object type. It more broadly introduces a `Target` interface, which is a subset of the `Claim` interface. An object satisfying only the `Target` interface would be intended to only serve as a method of propagating connection secrets and pointing to them in a specific `Namespace`. Initially, only `KubernetesTarget` would implement the `Target` interface, but it is possible the other `Target` types could be implemented in the future for scheduling workloads.

Note: This PR branches off #92 due to the fact that multiple secret propagation will be required for this functionality.

### Checklist

I have:


pkg/resource/interfaces.go
```diff
@@ -137,3 +137,15 @@ type Provider interface {
...
 	CredentialsSecretReferencer
 }
+
+// A Target is a Kubernetes object that refers to credentials to connect
+// to a deployment target. Target is a subset of the Claim interface.
+type Target interface {
+	runtime.Object
+	metav1.Object
+
+	LocalConnectionSecretWriterTo
```
hasheddan: I believe we will have to either modify `angryjet` or manually implement the necessary methods for `KubernetesTarget` to satisfy this interface due to the fact that the field used is `SecretRef` rather than `WriteConnectionSecretToRef`.

apis/core/v1alpha1/resource.go
```diff
@@ -208,3 +208,23 @@ type ProviderSpec struct {
...
 }
+
+// A TargetSpec defines the common fields of objects that satisfy the Target interface.
+type TargetSpec struct {
+	// SecretRef specifies the name of a Secret, in the same namespace as this
+	// target, to which any connection details for this target should be written
+	// or already exist. Connection secrets referenced by a target should
+	// contain information for connecting to a resource that allows for
+	// scheduling of workloads. +optional
+	SecretRef *LocalSecretReference `json:"secretRef,omitempty"`
```
hasheddan: Opted for `SecretRef` instead of `ConnectionSecretRef` because it seems implied that the the secret is will be for connection in this case. However, maybe it would be good to be more verbose?

pkg/resource/target_reconciler.go
```diff
@@ -0,0 +1,124 @@
...
+	defer cancel()
+
+	target := r.newTarget()
+	if err := r.client.Get(ctx, req.NamespacedName, target); err != nil {
+		// There's no need to requeue if we no longer exist. Otherwise we'll be
+		// requeued implicitly because we return an error.
+		return reconcile.Result{}, errors.Wrap(IgnoreNotFound(err), errGetTarget)
+	}
+
+	if target.GetWriteConnectionSecretToReference() == nil {
```
hasheddan: Not sure if we should use predicates to filter out any targets that do not have their `SecretRef` field set? I kind of like how this tells the user why the secret is not being propagated.

pkg/resource/interfaces.go
```diff
@@ -137,3 +137,15 @@ type Provider interface {
...
+	ManagedResourceReferencer
```
negz: Doesn't this interface usually correspond to the getters and setters for `resourceRef`? Are we reusing that for `clusterRef` here?

pkg/resource/secret_reconciler.go
```diff
@@ -57,7 +57,9 @@ const (
 // by propagating their data to another secret. Both secrets must consent to
 // this process by including propagation annotations. The Reconciler assumes it
 // has a watch on both propagating (from) and propagated (to) secrets.
-func NewSecretPropagatingReconciler(m manager.Manager) reconcile.Reconciler {
+func NewSecretPropagatingReconciler(m manager.Manager) reconcile.Reconciler { // nolint:gocyclo
+	// NOTE(hasheddan): This function is over our cyclomatic complexity goal.
+	// Be wary of adding additional complexity.
```
negz: I think you removed this linter comment from the other PR - is it sneaking back in here?

pkg/resource/target_reconciler.go
negz: It seems like we're going with the approach in which we don't automatically populate this reference when a cluster is being referenced? i.e. The user must specify both the target and where to write the secret? If so I think this makes sense.

pkg/resource/target_reconciler.go
```diff
@@ -0,0 +1,124 @@
...
+	client     client.Client
+	newTarget  func() Target
+	newManaged func() Managed
+
+	propagator ManagedConnectionPropagator
+}
+
+// NewTargetReconciler returns a Reconciler that reconciles
+// KubernetesTarget by setting their ConnectionSecretRef to that of the
+// referenced Kubernetes cluster implementation.
```
negz: This reads a little like we literally use the referenced cluster's secret, rather than propagating it.

pkg/resource/interfaces.go
negz: I think we'll need to modify angryjet. In reading this I wonder whether we should use a `LocalConnectionSecretReferencer` variant of this interface. We're only specifying where to write the secret in certain modes of operation (propagation from a cluster) - in others we're specifying where to read it from.

apis/core/v1alpha1/resource.go
hasheddan: Agreed, probably will change this to `ConnectrionSecretRef`.

pkg/resource/target_reconciler.go
hasheddan: Yes, that is where we stand currently. The tricky part here is that `writeConnectionSecretToRef` would make sense in the case that a cluster is referenced, but just `connectionSecretRef` makes sense if the secret is being referenced directly (i.e. you have manually created a secret and are referencing it). We could also have both of these fields and then...
1. If `connectionSecretRef` is set then we ignore `writeConnectionSecretToRef` and `clusterRef` and always assume that you have manually created the secret.
1. If `writeConnectionSecretToRef` is set and `clusterRef` is not then we do reconcile with the shared Target reconciler because the missing `clusterRef` means that the Target will not pass predicates of any controllers.
1. If `clusterRef` is set but `writeConnectionSecretToRef` is not then we hit this situation in the code above where we say "please tell us where to write your secret to" in the Target status. We could also opt to generate the secret name here?

What do you think about this pattern?

apis/core/v1alpha1/condition.go
```diff
@@ -36,7 +36,11 @@ const (
 	TypeSynced ConditionType = "Synced"
 
 	// TypeReferencesResolved resources' references are resolved
-	TypeReferencesResolved = "ReferencesResolved"
+	TypeReferencesResolved ConditionType = "ReferencesResolved"
+
+	// TypeSecretPropagated resources have had connection information propagated
+	// to their secret reference.
+	TypeSecretPropagated ConditionType = "SecretPropagated"
```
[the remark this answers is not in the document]
hasheddan: Sounds good to me!

pkg/resource/interfaces.go
hasheddan: Yes. This one is less contentious than the `LocalConnectionSecretWriterTo` because there is nothing that mandates that the `Target` will need to implement this existing interface. However, if we create another interface it will look very similar and function the exact same. I think the reason I did not add a new one is because though we have no plans for it at the moment, I feel that `Target` could be reused to propagate secrets from other types of resources in the future. Do you think we should add a `ClusterReferencer` interface or `ManagedClusterReferencer` interface?

pkg/resource/secret_reconciler.go
hasheddan: ah yes, I must have messed up the rebase.

pkg/resource/target_reconciler.go
hasheddan: Agree here, will modify.

pkg/resource/target_reconciler.go
negz: Having two different ways to reference a secret seems more complex than it needs to be to me. My preference is to just name this field `connectionSecretRef` and use it for both cases.

pkg/resource/interfaces.go
negz: I'm a little worried about using getter and setter interfaces that correspond to fields with different names - at the very least it goes against the established convention in Crossplane. I don't feel super strongly about this, but it tingles my spider sense.

pkg/resource/interfaces.go
negz: Could we keep using `LocalConnectionSecretWriterTo` for the existing resources that have a `writeConnectionSecretToRef` field, and add a new `LocalConnectionSecretReferencer` interface that we'd use only for targets with `connectionSecretRef` fields? I imagine the expense would be that we couldn't have `Target` be a strict subset of `Claim`?

hasheddan: The primary expense would be that the `Claim` and `Target` interfaces could not use the same managed secret propagation implementation.

hasheddan: @negz i think the options we have here are:
1. Implement a separate connection secret propagator for each of Claim and Target (low effort, duplicate code does not feel great though).
1. Use a common interface with methods that access fields with different names (current implementation).
1. Change the name of the Secret field in Target to writeConnectionSecretToRef (this is a weird name if referencing the Secret directly).

Do you have a preference for any of these or an idea for a path I have defined above?

pkg/resource/target_reconciler.go
```diff
@@ -0,0 +1,124 @@
...
+
+	if target.GetWriteConnectionSecretToReference() == nil {
+		// If the ConnectionSecretRef is not set on this Target, we will not
+		// know where to propagate the secret. We do not explicitly requeue
+		// because we have a watch on updates to the KubernetesTarget.
+		target.SetConditions(v1alpha1.SecretPropagatedError(errors.New(errMissingSecretRef)))
+		return reconcile.Result{Requeue: false}, errors.Wrap(r.client.Status().Update(ctx, target), errUpdateTarget)
+	}
+
+	if meta.WasDeleted(target) {
```
muvaf: Does the secret created by this reconciler have the owner reference of this object for cleanup after it's deleted? I checked the propagator implementation but couldn't catch it.

pkg/resource/api.go
```diff
@@ -91,7 +91,7 @@ func NewAPIManagedConnectionPropagator(c client.Client, t runtime.ObjectTyper) *
 }
 
 // PropagateConnection details from the supplied resource to the supplied claim.
-func (a *APIManagedConnectionPropagator) PropagateConnection(ctx context.Context, cm Claim, mg Managed) error {
+func (a *APIManagedConnectionPropagator) PropagateConnection(ctx context.Context, cm Target, mg Managed) error {
```
muvaf: I guess renaming `cm Target` to something like `tr Target` or `t Target` would read better.

hasheddan: Yep, if we keep this implementation, I definitely think the variable name should be representative of the most narrow accepted interface.

pkg/resource/target_reconciler.go
hasheddan: It is a local secret so we would only need name, but I generally agree that we should generate the name if not provided (per 3 above).

pkg/resource/target_reconciler.go
hasheddan: Yep! The propagator calls `LocalConnectionSecretFor` on the `Target` which sets owner references. Importantly, we also plan to add an owner reference if a `Target` references an existing `Secret` directly (not part of this PR) such that if a `Secret` is manually created in a `Namespace` and a `KubernetesTarget` is created to make it a `KubernetesApplication` scheduling target, it will automatically be cleaned up by garbage collection when the `Target` is deleted. In the case that a `KubernetesTarget` is created that references an existing `Secret` that was claimed in the `Namespace`, we plan to add an owner reference to that `Secret` such that it is not garbage collected until both the `KubernetesCluster` claim and the `KubernetesTarget` are deleted. This would mean that a `KubernetesTarget` could reference a `Secret` that will not allow for successful connection to a k8s cluster (in the case that the `KubernetesCluster` claim was deleted and `Reclaim: Delete`), but this seems fine due to the fact that `KubernetesTarget` makes no claims (apologize for using an overloaded term) about making a successful connection, just rather that a `Secret` with connection info exists in the `Namespace`.

https://github.com/crossplaneio/crossplane-runtime/blob/d5c9dedd2a61108120fb5b3bba32e5a8d43ffeee/pkg/resource/resource.go#L45

pkg/resource/target_reconciler.go
```diff
@@ -0,0 +1,124 @@
...
+		return reconcile.Result{Requeue: false}, nil
+	}
+
+	managed := r.newManaged()
+	if err := r.client.Get(ctx, meta.NamespacedNameOf(target.GetResourceReference()), managed); err != nil {
+		target.SetConditions(v1alpha1.SecretPropagatedError(err))
+		return reconcile.Result{RequeueAfter: aShortWait}, errors.Wrap(r.client.Status().Update(ctx, target), errUpdateTarget)
+	}
+
+	if !IsBound(managed) {
```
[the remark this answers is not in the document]
hasheddan: That is the current thinking. The idea is that we generally think of an unbound resource as one that is not "in use". We would like for a k8s cluster managed resource to be claimed to indicate that it is "in use" before we allow for scheduling to it.

pkg/resource/interfaces.go
hasheddan: Following up on this from today's stand-up: consensus is that for the time being we will use `LocalConnectionSecretWriterTo` in `Target` interface, make the name of the field in the embedded `TargetSpec` struct `WriteConnectionSecretToRef`, but have the JSON tag field name just be `connectionSecretRef`. Will update and add `TODO` to denote that we will likely try to clean this up in the future.

pkg/resource/interfaces.go
hasheddan: The secret propagator actually mandates this sub-interface as well, so I have left it as-is and tried to make `Target` a little more compliant by changing the struct field to `ResourceReference` but keeping the JSON tag as `clusterRef`.

pkg/resource/target_reconciler.go
hasheddan: Was looking at this and am wondering if we have any other examples of updating a `spec` field of a user-created managed resource? Feels like a bit of an anti-pattern.

negz: Any of our managed resource late initializations are an example of this; upstream types like `Service` do it to (e.g. will set an IP if you don't specify one).

apis/core/v1alpha1/condition.go
```diff
@@ -38,9 +38,9 @@ const (
 	// TypeReferencesResolved resources' references are resolved
 	TypeReferencesResolved ConditionType = "ReferencesResolved"
 
-	// TypeSecretPropagated resources have had connection information propagated
-	// to their secret reference.
-	TypeSecretPropagated ConditionType = "SecretPropagated"
+	// TypeConnectionSecretPropagated resources have had connection information
+	// propagated to their secret reference.
+	TypeConnectionSecretPropagated ConditionType = "SecretPropagated"
```
negz: Haha, I think you got the opposite of my intent here, sorry. I was thinking:

```go
TypeSecretPropagated ConditionType = "ConnectionSecretPropagated"
```

i.e. More context for users, less typing for us.

apis/core/v1alpha1/condition.go
```diff
@@ -299,10 +299,10 @@ func ReferenceResolutionBlocked(err error) Condition {
 // successfully propagated connection data to the referenced secret.
 func SecretPropagatedSuccess() Condition {
```
negz: Can we use `Propagation` rather than `Propagated` here too?

apis/core/v1alpha1/resource.go
```diff
@@ -209,19 +209,20 @@ type ProviderSpec struct {
...
-
-	// A ClusterRef specifies an existing managed resource, in any namespace, to
-	// which this target should attempt to propagate a connection secret from.
-	// +optional
-	ClusterRef *corev1.ObjectReference `json:"clusterRef,omitempty"`
+	// WriteConnectionSecretToReference specifies the name of a Secret, in the
+	// same namespace as this target, to which any connection details for this
+	// target should be written or already exist. Connection secrets referenced
+	// by a target should contain information for connecting to a resource that
+	// allows for scheduling of workloads. +optional
```
negz: Did the `+optional` stanzas get wrapped here (and on surrounding types)?

pkg/resource/target_reconciler.go
hasheddan: Good point. Are you in agreement on generating secret name of not provided?

apis/core/v1alpha1/condition.go
hasheddan: Ah! That makes more sense haha

apis/core/v1alpha1/resource.go
hasheddan: Looks like it, will fix!

pkg/resource/target_reconciler.go
negz: I think:

* If a secret is specified (either by the user or by us having propagated one from a specified target) we ignore the target (if any) and use the specified secret.
* If a target is specified _and_ no secret is specified, we create a secret and configure the target's secret to propagate to our newly created secret.

This means that in the case where a user specifies a target _and_ a secret, we ignore the target and just use the secret.

hasheddan: I am assuming you mean Cluster where you use Target here, and if so I mostly agree, but it would mean that the case where both are provided we would not propagate the secret. I think if both the cluster is referenced and a secret name is provided then we should propagate to that secret name.

negz: > I am assuming you mean Cluster where you use Target here

Derp - yeah I mean cluster.

> I think if both the cluster is referenced and a secret name is provided then we should propagate to that secret name.

So: 

1. If only a secret is specified, we presume it exists and use it to connect to the cluster.
2. If only a cluster is specified, we generate a secret name and setup propagation.
3. If both are specified, we setup propagation from the cluster to the secret.

For some reason I was thinking 3 would not work, but now in the future of 10 minutes later I can't recall why. I imagine it won't work if the user specifies an _existing_ secret (which won't have the propagation annotations), but if they specify a non-existent secret we'll be able to create it.

hasheddan: I *think* you are right about the case in which the `Secret` already exists because we will try to get a `Secret` of the supplied name but it will not be owned by the `Target` so I imagine `CreateOrUpdate` will fail. I think this is the desired behavior we want here.

pkg/resource/target_reconciler.go
```diff
@@ -0,0 +1,123 @@
...
+		// There's no need to requeue if we no longer exist. Otherwise we'll be
+		// requeued implicitly because we return an error.
+		return reconcile.Result{}, errors.Wrap(IgnoreNotFound(err), errGetTarget)
+	}
+
+	if target.GetWriteConnectionSecretToReference() == nil {
+		// If the ConnectionSecretRef is not set on this Target, we generate a
+		// Secret name that matches the UID of the Target. We implicitly
+		// requeued because of the Target update.
+		target.SetWriteConnectionSecretToReference(&v1alpha1.LocalSecretReference{Name: string(target.GetUID())})
```
hasheddan: @negz @muvaf  The `Secret` will be created with a `Name` that matches the `Target` UID. Does that feel like the right pattern to y'all?

negz: SGTM!

PR #106, opened 2020-01-14
Break reconcilers out into distinct packages

negz:

This PR break all reconcilers out into distinct packages, rather than the resource package.

```go
// Before
import "github.com/crossplaneio/crossplane-runtime/pkg/resource"

r := resource.NewClaimSchedulingReconciler(...)
```

```go
// After
import "github.com/crossplaneio/crossplane-runtime/pkg/reconciler/claimscheduling"

r := claimscheduling.NewReconciler(...)
```

I'm hoping this will help us pass in loggers and eventers using variadic options without awkward names like WithClaimBindingReconcilerLogger. This is a huge diff but there's no functional changes - only moving things around and renaming them.

I was hoping we'd be able to do this without a breaking API change by using type aliases, but doing so would create cyclic imports, because the new reconciler packages depend on the resource package, which is where we'd need to create the type alias to the reconciler packages.

### Checklist

I have:


negz:
Note: This will require an update to the Services Developer Guide (in the Crossplane repo) to point to the new packages. It will also require a new minor version release of crossplane-runtime, and some (boilerplate) changes to existing stacks.

pkg/reconciler/claimbinding/api.go
```diff
@@ -125,18 +124,18 @@ func (a *APIManagedConnectionPropagator) PropagateConnection(ctx context.Context
...
 	}); err != nil {
 		return errors.Wrap(err, errCreateOrUpdateSecret)
 	}
 
-	meta.AddAnnotations(mgcs, map[string]string{
-		strings.Join([]string{AnnotationKeyPropagateToPrefix, string(cmcs.GetUID())}, SlashDelimeter): strings.Join([]string{cmcs.GetNamespace(), cmcs.GetName()}, SlashDelimeter),
-	})
+	k := strings.Join([]string{resource.AnnotationKeyPropagateToPrefix, string(cmcs.GetUID())}, resource.AnnotationDelimiter)
+	v := strings.Join([]string{cmcs.GetNamespace(), cmcs.GetName()}, resource.AnnotationDelimiter)
+	meta.AddAnnotations(mgcs, map[string]string{k: v})
```
hasheddan: Thanks for cleaning this up :)

pkg/reconciler/claimbinding/configurator.go
```diff
@@ -51,16 +52,16 @@ func NewObjectMetaConfigurator(_ runtime.ObjectTyper) *ObjectMetaConfigurator {
...
 
 // Configure the supplied Managed resource's object metadata.
-func (c *ObjectMetaConfigurator) Configure(ctx context.Context, cm Claim, cs Class, mg Managed) error {
+func (c *ObjectMetaConfigurator) Configure(ctx context.Context, cm resource.Claim, cs resource.Class, mg resource.Managed) error {
 	return ConfigureNames(ctx, cm, cs, mg)
 }
 
 // ConfigureNames configures the name and external name of the supplied managed
-// resource. The managed resource name is derived from the supplied resource
+// resource. The managed resource name is derived from d resource
```
hasheddan: I am thinking this is an unintentional change here.

pkg/resource/enqueue_handlers.go
```diff
@@ -109,7 +111,7 @@ func addPropagated(obj runtime.Object, queue adder) {
 		if !strings.HasPrefix(key, AnnotationKeyPropagateToPrefix) {
 			continue
 		}
-		t := strings.Split(val, SlashDelimeter)
+		t := strings.Split(val, AnnotationDelimiter)
```
hasheddan: +1

pkg/reconciler/claimbinding/configurator.go
negz: It was - thanks!

pkg/reconciler/claimbinding/api.go
negz: Thanks! To be honest I think it was fine as is, but it started to feel pretty long after adding the `resource.` package name in a few places.

negz:
@hasheddan PTAL - I've updated this to factor your `Target` reconciler implementation out into a distinct package like the other reconcilers. Two things of note:

* I updated the `ManagedConnectionPropagator` interface to take `LocalConnectionSecretOwner` as an argument, rather than `Claim` or `Target`. I'd forgotten it existed, and it was pretty much a subset of both `Target` and `Claim`.
* I added a second commit that makes the target reconciler's `ManagedConnectionPropagator` configurable, which lets us stub it out for slightly smaller and more loosely coupled unit tests.

PR #115, opened 2020-02-11
Make ProviderSpec credentialsSecretRef optional

hasheddan:

Makes the `credentialsSecretRef` field optional in `ProviderSpec` such that alternative auth methods can be utilized without having to reference a `Secret`.

Ref: https://github.com/crossplaneio/stack-aws/pull/126#discussion_r376681446

### Checklist

I have:


hasheddan:
@negz that is correct :+1: I am happy to make those updates. We could also not make it a pointer, then we would just get empty strings and fail to get the secret, but I feel like that would go against our conventions so I would prefer to just make the updates if that sounds good

negz:
@hasheddan SGTM - I imagine the updates should be fairly mechanical.

hasheddan:
@negz anything else needed here?

negz:
Nope - just forgot to hit merge. :)

hasheddan:
@negz thanks :)

PR #121, opened 2020-02-24
Add a package for getting and setting unstructured fields by path

negz:

https://github.com/kubernetes-sigs/kustomize/blob/d190e1/api/k8sdeps/kunstruct/helper.go
https://github.com/kubernetes/apimachinery/blob/2373d0/pkg/apis/meta/v1/unstructured/helpers.go

This package is similar to the above two, with some key differences:

* Our fieldpath lexer is a little stricter; it won't allow dangling open braces, unexpected periods, or empty brackets. It also supplies the position of any syntax error if lexing fails.
* We support setting and getting fields within a pkg/json unmarshalled object by fieldpath, but only leaf fields. We do not support getting or setting entire subtrees of an object.

### Checklist

I have:

### DIFF
--- a/pkg/fieldpath/fieldpath.go
+++ b/pkg/fieldpath/fieldpath.go
@@ -1,0 +1,280 @@
+/*
+Copyright 2019 The Crossplane Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package fieldpath provides utilities for working with field paths.
+//
+// Field paths reference a field within a Kubernetes object via a simple string.
+// API conventions describe the syntax as "standard JavaScript syntax for
+// accessing that field, assuming the JSON object was transformed into a
+// JavaScript object, without the leading dot, such as metadata.name".
+//
+// Valid examples:
+//
+// * metadata.name
+// * spec.containers[0].name
+// * data[.config.yml]
+// * metadata.annotations['crossplane.io/external-name']
+// * spec.items[0][8]
+// * apiVersion
+// * [42]
+//
+// Invalid examples:
+//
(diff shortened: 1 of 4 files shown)


pkg/fieldpath/fieldpath.go
```diff
@@ -0,0 +1,280 @@
...
+	return Field(s)
+}
+
+// Field produces a new segment from the supplied string. The segment is always
+// considered to be an object field name.
+func Field(s string) Segment {
+	return Segment{Type: SegmentField, Field: strings.Trim(s, "'\"")}
+}
+
+// Parse the supplied path into a slice of Segments. API conventions describes a
```
hasheddan: Looks like comment got cut off here.

pkg/fieldpath/paved.go
```diff
@@ -0,0 +1,266 @@
...
+
+	b, ok := v.(bool)
+	if !ok {
+		return false, errors.Errorf("%s: not a bool", segments)
+	}
+	return b, nil
+}
+
+// GetNumber value of the supplied field path.
+func (p *Paved) GetNumber(path string) (float64, error) {
```
hasheddan: Do you think helpers like `GetNumberAsInt` could be useful here?

pkg/fieldpath/fieldpath.go
negz: Thanks, fixed!

pkg/fieldpath/paved.go
negz: I thought about that, but given JSON numbers should always unmarshal as a float64 it seemed safer to return the float64 and let the caller handle conversion. This way it will always be obvious to the caller that the field value might have been a float, and thus might be truncated if/when it's asserted to an int.

hasheddan: SGTM

muvaf:
> We support setting and getting fields within a pkg/json unmarshalled object by fieldpath, but only leaf fields. We do not support getting or setting entire subtrees of an object.

Do you mean we can only get primitive types, like `int`, `string` etc.?

negz:
> Do you mean we can only get primitive types, like int, string etc.?

Right. It should be possible to get subtrees (arrays or objects, potentially of other arrays or objects) but I don't have a use case for that right now.

negz:
> I don't have a use case for that right now.

I added some getters for string slice and string map - we can add setters in future if they're common enough. I imagine that mostly we'll be using plain old `GetValue` and `SetValue` (which return and accept `interface{}`) to enable `fromFieldPath/toFieldPath` patch functionality.

PR #122, opened 2020-02-27
Support ApplicationConfiguration controller

negz:

This PR adds two relatively unrelated pieces of functionality:

* Getters and setters for the unstructured content of a `fieldpath.Paved` This helps convert a `*fieldpath.Paved` to and from an `*unstructured.Unstructured`. The API matches that of `*unstructured.Unstructured`
* An `Apply` function that either creates or patches a resource. We've implemented this in at least two controllers, and I've found myself doing it again.

The only theme linking these commits is "stuff I need to implement the OAM ApplicationConfiguration controller". :)

### Checklist

I have:

### DIFF
--- a/pkg/resource/resource.go
+++ b/pkg/resource/resource.go
@@ -18,6 +18,7 @@ package resource
 
 import (
 	"context"
+	"encoding/json"
 
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
@@ -25,6 +26,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
 	"github.com/crossplane/crossplane-runtime/pkg/meta"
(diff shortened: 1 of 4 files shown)


pkg/resource/resource.go
```diff
@@ -227,3 +229,52 @@ func IsBound(b Bindable) bool {
...
+	}
+	if err != nil {
+		return errors.Wrap(err, "cannot get object")
+	}
+
+	if opts.ControllersMustMatch && !meta.HaveSameController(o.(metav1.Object), m) {
+		return errors.New("existing object has a different (or no) controller")
+	}
+
+	return errors.Wrap(c.Patch(ctx, existing, client.MergeFrom(o)), "cannot patch object")
```
muvaf: `client.MergeFrom` creates a merge patch JSON from the difference but that merge patch aims to make the final state of the object exactly like `o`. Meaning if any field is late-inited and you didn't work on the most up-to-date version, produced merge patch will have a `null` for that field and remove it. So, the ideal workflow for `MergeFrom` is that you fetch the object, make the changes and then issue a `MergeFrom` patch. However, here the object fetched and directly supplied for patch, so you'll delete the fields that are populated but you don't know about.

I've used [`ConstantPatch`](https://github.com/crossplane/templating-controller/blob/cb1327d/pkg/controllers/templating_reconciler.go#L182) in templating-controller after realizing that, which solved the problem of deleting the fields that `o` is not aware and are populated.

negz: Aha, thanks! I initially copied your approach but later switched to `MergeFrom` simply because it seemed more succinct. Perhaps this behaviour should be configurable using an `ApplyOption`? If so, I wonder whether the default should be to remove or preserve any unexpected fields. The former seems safer, but the latter is arguably less surprising.

hasheddan: Just a note that `ConstantPatch` is deprecated in favor of `RawPatch`. Would likely make sense to go ahead and use `RawPatch` so that we do not have to fix in coming updates 👍 https://github.com/kubernetes-sigs/controller-runtime/blob/9f8aab6551a3ecc211b6c6078245899495b7e0f5/pkg/client/patch.go#L57

negz: @hasheddan I noticed that, but `RawPatch` doesn't seem to exist yet in the version of controller-runtime we're using.

hasheddan: @negz do you think it would be worth bumping our version or no?

negz: > do you think it would be worth bumping our version or no?

I ended up not using `ConstantPatch`, but instead implementing the `Patch` interface in a fashion similar to `ConstantPatch`.

negz:
> What do you think?

@muvaf made a solid case for only supporting the "constant patch" option. This is effectively what I've done, by implementing the `Patch` interface in a fashion that returns a constant patch based on the original object.

muvaf:
@negz Thanks for writing modified version of `ConstantPatch` where the input is directly the object instead of `[]byte` to bury that conversion in the `Data` function. I'll update templating-controller, too, to use this function once merged.

negz:
> I'll update templating-controller, too, to use this function once merged.

@muvaf I'll hopefully merge today; I haven't actually run this code yet so I want to do that first. :)

negz:
@muvaf @hasheddan I added two commits just now. They’re both fairly small updates that relate to the two existing approved commits, so I’m going to cheekily hit merge but I’d appreciate it if you could take a retroactive look when you come online.

PR #123, opened 2020-02-28
Implement GetExternalTags 

muvaf:

Implement GetExternalTags to return Crossplane tags for managed resource controllers to tag their external resources.

Makes it easier to implement https://github.com/crossplane/crossplane/blob/master/design/one-pager-managed-resource-api-design.md#external-resource-labeling

### Checklist

I have:

### DIFF
--- a/pkg/resource/resource.go
+++ b/pkg/resource/resource.go
@@ -19,6 +19,7 @@ package resource
 import (
 	"context"
 	"encoding/json"
+	"strings"
 
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
@@ -44,6 +45,15 @@ const (
 	AnnotationKeyPropagateFromUID       = "from.propagate.crossplane.io/uid"
 
 	AnnotationDelimiter = "/"
+)
+
+// External resources are tagged/labelled with the following keys in the cloud
+// provider API if the type supports.
+const (
+	ExternalResourceTagKeyKind     = "crossplane-kind"
+	ExternalResourceTagKeyName     = "crossplane-name"
+	ExternalResourceTagKeyClass    = "crossplane-class"
+	ExternalResourceTagKeyProvider = "crossplane-provider"
 )
 
 // A ClaimKind contains the type metadata for a kind of resource claim.
@@ -298,3 +308,19 @@ type patch struct{ from runtime.Object }
 
 func (p *patch) Type() types.PatchType                 { return types.MergePatchType }
 func (p *patch) Data(_ runtime.Object) ([]byte, error) { return json.Marshal(p.from) }
+
+// GetExternalTags returns the identifying tags to be used to tag the external
+// resource in provider API.
+func GetExternalTags(mg Managed) map[string]string {
+	tags := map[string]string{
+		ExternalResourceTagKeyKind: strings.ToLower(mg.GetObjectKind().GroupVersionKind().GroupKind().String()),
+		ExternalResourceTagKeyName: mg.GetName(),
+	}
+	if mg.GetClassReference() != nil {
+		tags[ExternalResourceTagKeyClass] = mg.GetClassReference().Name
+	}
+	if mg.GetProviderReference() != nil {
+		tags[ExternalResourceTagKeyProvider] = mg.GetProviderReference().Name
+	}
+	return tags
+}
--- a/pkg/meta/meta_test.go
+++ b/pkg/meta/meta_test.go
@@ -810,15 +810,13 @@ func TestWasCreated(t *testing.T) {
 }
 
 func TestGetExternalName(t *testing.T) {
-	testName := "my-external-name"
-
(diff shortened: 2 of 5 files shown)


muvaf:
I thought I opened the PR but seems like forgot to click the button :(

@negz @hasheddan I was thinking about adding the following labels, too, to make it possible do more detailed resource filterin in cloud provider:
* `crossplane.io/class`
* `crossplane.io/provider`

What do you guys think?

muvaf:
> Class tag: I am not sure how much longer are current class structure will be relevant so might hold off on that one.

Well, be it `class` or some other thing, I think we'll have something that serves as resource class. So, it might be worth adding it now and change the string if the name changes.

> Provider tag: Assuming this would be the name of the literal provider object? If so I think that could be useful!

Yes, it's literal object name.

hasheddan:
That sounds fine. Another thing that makes class different is that statically provisioned managed resources do not have a class ref while the other 3 tags are values that are always present.

muvaf:
> Another thing that makes class different is that statically provisioned managed resources do not have a class ref while the other 3 tags are values that are always present.

Yeah, I imagine we'd just skip class label in that case.

I'll circle back to API patterns doc and add these if everyone is onboard. @negz what do you think?

pkg/resource/resource.go
```diff
@@ -43,6 +43,15 @@ const (
 	AnnotationDelimiter = "/"
 )
 
+// External resources are tagged/labelled with the following keys in the cloud
+// provider API if the type supports.
+const (
+	ExternalResourceKindTagKey     = "crossplane.io/kind"
+	ExternalResourceNameTagKey     = "crossplane.io/name"
+	ExternalResourceClassTagKey    = "crossplane.io/class"
+	ExternalResourceProviderTagKey = "crossplane.io/provider"
```
negz: Nit: I think we have two formats for key constant names emerging within this package. Above we have `AnnotationKeyPropagateFromName`, and here we have `ExternalResourceKindTagKey`. We should probably pick one of:

1. `AnnotationKeyPropagateFromName` and `ExternalResourceTagKeyKind` (purpose then name)
2. `PropagateFromNameAnnotationKey` and `ExternalResourceKindTagKey` (name then purpose)

I prefer option 1, because it groups all the keys that are an `AnnotationKey` or an `ExternalResourceTag` lexicographically by their prefix.

pkg/resource/resource.go
```diff
@@ -227,3 +236,19 @@ func IsBound(b Bindable) bool {
 func IsConditionTrue(c v1alpha1.Condition) bool {
 	return c.Status == corev1.ConditionTrue
 }
+
+// GetExternalTags returns the identifying tags to be used to tag the external
+// resource in provider API.
+func GetExternalTags(mg Managed) map[string]string {
+	tags := map[string]string{
+		ExternalResourceKindTagKey: mg.GetObjectKind().GroupVersionKind().GroupKind().String(),
```
negz: Can we rely on this field always being set? I recall it was being automatically cleared by some clients, apparently intentionally, which is why most of our crossplane-runtime functions that need a GVK require we pass either a GVK or a scheme to determine one. https://github.com/kubernetes-sigs/controller-runtime/pull/528 indicates that the old behaviour I described may now be gone.

pkg/resource/resource.go
muvaf: I agree. Option 1 sounds good.

pkg/resource/resource.go
muvaf: I will test it and write here. I simply relied on the fact that all those functions returned non-pointer objects.

muvaf: Verified that it does return GVK correctly.

muvaf:
@hasheddan @negz Heads-up, I'm replacing `/` with `-` as it seems `/` in tag key is only accepted by AWS and hyphen is accepted by all three. See the restrictions for each:
* AWS https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html#tag-restrictions
* GCP https://cloud.google.com/compute/docs/labeling-resources#label_format
* Azure https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/tag-resources#limitations

muvaf:
Alright, `.` is not supported by GCP. What do you think about `crossplane-kind`, `crossplane-class` etc ?

pkg/resource/resource.go
negz: Nice! This might allow us to simplify some of our functions in `pkg/meta` (and possibly `pkg/resource`?) that assume they can't rely on `GetObjectKind` to actually return the kind of an object.

negz:
Looks like this needs to be rebased - I'll merge once the branches align.

muvaf:
@negz ready to go! I added a small change to make `kind` lowercase(checked kinds are not case-sensitive) to comply with GCP more easily without breaking anything.

PR #126, opened 2020-03-06
Ignore previously cleaned up CRDs in integration tests

hasheddan:

Updates CRD cleaner to ignore deletion errors that are `Not Found`. Also fixes issue with formatting FNV hash at string.

### Checklist

I have:

### DIFF
--- a/pkg/test/integration/remote.go
+++ b/pkg/test/integration/remote.go
@@ -35,10 +35,8 @@ func downloadPath(url, path string) (string, error) {
 
 	// Subdirectory is given name of fnv hash of url.
 	hasher := fnv.New32a()
-	if _, err = hasher.Write([]byte(url)); err != nil {
-		return "", err
-	}
-	dst := filepath.Join(path, fmt.Sprint(hasher.Sum32()))
+	hasher.Write([]byte(url)) // nolint:errcheck
+	dst := filepath.Join(path, fmt.Sprintf("%x", hasher.Sum32()))
 
 	c := getter.Client{
 		Src:  url,
--- a/pkg/test/integration/remote.go
+++ b/pkg/test/integration/remote.go
@@ -17,6 +17,7 @@ limitations under the License.
 package integration
 
 import (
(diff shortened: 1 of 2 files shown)


pkg/test/integration/server.go
```diff
@@ -28,6 +28,7 @@ import (
 	// Allow auth to cloud providers
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
```
golangcibot: File is not `goimports`-ed with -local github.com/crossplane/crossplane (from `goimports`)
```suggestion

```

hasheddan:
@negz anything blocking here? If not, I think we can merge :)

pkg/test/integration/remote.go
```diff
@@ -37,13 +38,14 @@ func downloadPath(url, path string) (string, error) {
 	if _, err = hasher.Write([]byte(url)); err != nil {
```
negz: https://golang.org/pkg/hash/#Hash

Not introduced in this PR, but `hasher.Write` will never return an error so you could probably just use `//nolint` here.

pkg/test/integration/remote.go
```diff
@@ -37,13 +38,14 @@ func downloadPath(url, path string) (string, error) {
...
 		return "", err
 	}
-	dst := filepath.Join(path, string(hasher.Sum32()))
+	dst := filepath.Join(path, fmt.Sprint(hasher.Sum32()))
```
negz:
```suggestion
	dst := filepath.Join(path, fmt.Sprintf("%x", hasher.Sum32()))
```

I usually like to represent hashes as hexadecimal strings. That said, the only advantage of doing so is probably a shorter string in some cases, so feel free to ignore this suggestion.

hasheddan:
@negz I updated here per your suggestions and tested with the GCP integration tests :+1:

PR #129, opened 2020-03-16
Pass apply checks as functions to Applicators

hasheddan:

Updates `Applicator` interface to take functions which can check conditions before executing patch. Allows for package consumer to pass in custom checks (ref: https://github.com/crossplane/crossplane/pull/1341/files#diff-ba7cd8f84dd4227171ca0d3ed2bf46baR275). Also allows for any transformations to be applied to object before patch (ref: https://github.com/crossplane/addon-oam-kubernetes-remote/pull/1/files#diff-c4486ee3d6e8e3a194c5722f78c15955R73).

@negz I'll go back and update any places we are already using `Apply` after this is merged :+1: 

### Checklist

I have:

### DIFF
--- a/pkg/resource/resource.go
+++ b/pkg/resource/resource.go
@@ -254,34 +254,24 @@ func (fn ApplyFn) Apply(ctx context.Context, c client.Client, o runtime.Object, 
 	return fn(ctx, c, o, ao...)
 }
 
-// ApplyOptions configure how changes are applied to an object.
-type ApplyOptions struct {
-	// ControllersMustMatch requires any existing object to have a controller
-	// reference, and for that controller reference to match the controller
-	// reference of the supplied object.
-	ControllersMustMatch bool
-}
-
-// An ApplyOption configures how changes are applied to an object.
-type ApplyOption func(a *ApplyOptions)
+// An ApplyOption is a function that checks for a condition before patch.
+type ApplyOption func(ctx context.Context, current, desired runtime.Object) error
 
 // ControllersMustMatch requires any existing object to have a controller
 // reference, and for that controller reference to match the controller
 // reference of the supplied object.
 func ControllersMustMatch() ApplyOption {
-	return func(a *ApplyOptions) {
-		a.ControllersMustMatch = true
+	return func(_ context.Context, current, desired runtime.Object) error {
+		if !meta.HaveSameController(current.(metav1.Object), desired.(metav1.Object)) {
+			return errors.New("existing object has a different (or no) controller")
+		}
+		return nil
 	}
 }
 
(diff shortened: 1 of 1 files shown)


pkg/resource/resource.go
```diff
@@ -243,45 +243,33 @@ func IsConditionTrue(c v1alpha1.Condition) bool {
...
-	// ControllersMustMatch requires any existing object to have a controller
-	// reference, and for that controller reference to match the controller
-	// reference of the supplied object.
-	ControllersMustMatch bool
-}
-
-// An ApplyOption configures how changes are applied to an object.
-type ApplyOption func(a *ApplyOptions)
+// An ApplyOptFn is a function that checks for a condition before patch.
+type ApplyOptFn func(ctx context.Context, current, desired runtime.Object) error
```
negz: I'd prefer to keep the name `ApplyOption` here, despite our common pattern of pluggable functions ending in `Fn`.

pkg/resource/resource.go
```diff
@@ -243,45 +243,33 @@ func IsConditionTrue(c v1alpha1.Condition) bool {
...
 // reference of the supplied object.
-func ControllersMustMatch() ApplyOption {
-	return func(a *ApplyOptions) {
-		a.ControllersMustMatch = true
+func ControllersMustMatch(ctx context.Context, current, desired runtime.Object) error {
+	if !meta.HaveSameController(current.(metav1.Object), desired.(metav1.Object)) {
+		return errors.New("existing object has a different (or no) controller")
 	}
+	return nil
 }
```
negz: While this pattern works for `ControllersMustMatch` I think we should prefer to have our options _return_ a function with signature `func(ctx context.Context, current, desired runtime.Object) error` rather than _being_ a function with that signature. This would allow us to be consistent if we instead wanted an option like:

```go
func ControllerMustMatch(c corev1.ObjectReference) ApplyOption {
    return func(_ context.Context, current, desired runtime.Object) error {
       // stuff
       return nil
    }
}
```
```

pkg/resource/resource.go
negz: Also, how likely do you think we are to need `ctx` here?

hasheddan: I would think not very likely, but why not pass it in case?

hasheddan:
@negz updated :+1:

pkg/resource/resource.go
negz: Haha okay I'm sold. Can you change it to `_`in `ControllersMustMatch` to reflect that it's unused?

hasheddan: Done :)

PR #130, opened 2020-03-18
Add OAM workload and trait reconcilers

hasheddan:

This PR moves the OAM `Workload` and `Trait` reconcilers from https://github.com/crossplane/addon-oam-kubernetes-remote. We should not merge until we figure out the `crossplane` circular dependency this introduces, which is a result of defining the `WorkloadReferencer` interface for `Trait` here.

### Checklist

I have:


pkg/reconciler/oam/workload/translate.go
```diff
@@ -0,0 +1,151 @@
...
+// NoopWrapper does not wrap the workload translation and does not return error.
+func NoopWrapper(ctx context.Context, w resource.Workload, objs []Object) ([]Object, error) {
+	return objs, nil
+}
+
+var _ TranslationWrapper = ServiceInjector
+
+// ServiceInjector adds a Service object for the first Port on the first
+// Container for the first Deployment observed in a workload translation.
+func ServiceInjector(ctx context.Context, w resource.Workload, objs []Object) ([]Object, error) {
```
hasheddan: @negz I kept this for now since it only relies on k8s types and seems like it could be reused across remote and local OAM addons. Happy to remove if you prefer.

pkg/reconciler/oam/trait/modify.go
```diff
@@ -0,0 +1,84 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+const (
+	errNotKubeApp           = "object passed to KubernetesApplication accessor is not KubernetesApplication"
+	errNoDeploymentForTrait = "no deployment found for trait in KubernetesApplication"
+)
+
+var (
+	deploymentKind       = reflect.TypeOf(appsv1.Deployment{}).Name()
```
golangcibot: var `deploymentKind` is unused (from `unused`)

pkg/reconciler/oam/workload/translate.go
```diff
@@ -0,0 +1,151 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+const (
+	errInjectService = "unable to inject Service in objects"
+)
+
+var (
+	deploymentGroupVersionKind = appsv1.SchemeGroupVersion.WithKind(reflect.TypeOf(appsv1.Deployment{}).Name())
```
golangcibot: var `deploymentGroupVersionKind` is unused (from `unused`)

pkg/reconciler/oam/workload/translate_test.go
```diff
@@ -0,0 +1,267 @@
...
+			})
+		}
+		d.Spec.Template.Spec.Containers = append(d.Spec.Template.Spec.Containers, corev1.Container{
+			Name:  containerName,
+			Ports: p,
+		})
+	}
+}
+
+func dmWithReplicas(r *int32) deploymentModifier {
```
golangcibot: func `dmWithReplicas` is unused (from `unused`)

pkg/reconciler/oam/workload/translate_test.go
```diff
@@ -0,0 +1,267 @@
...
+
+var (
+	workloadName      = "test-workload"
+	workloadNamespace = "test-namespace"
+	workloadUID       = "a-very-unique-identifier"
+
+	containerName = "test-container"
+	portName      = "test-port"
+
+	trueVal = true
```
golangcibot: var `trueVal` is unused (from `unused`)

pkg/reconciler/oam/trait/modify.go
```diff
@@ -0,0 +1,84 @@
...
+	deploymentAPIVersion = appsv1.SchemeGroupVersion.String()
```
golangcibot: var `deploymentAPIVersion` is unused (from `unused`)

pkg/reconciler/oam/trait/modify.go
```diff
@@ -0,0 +1,84 @@
...
+	"reflect"
+
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+const (
+	errNotKubeApp           = "object passed to KubernetesApplication accessor is not KubernetesApplication"
+	errNoDeploymentForTrait = "no deployment found for trait in KubernetesApplication"
```
golangcibot: const `errNoDeploymentForTrait` is unused (from `unused`)

pkg/reconciler/oam/trait/modify.go
```diff
@@ -0,0 +1,84 @@
...
+	"context"
+	"reflect"
+
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+const (
+	errNotKubeApp           = "object passed to KubernetesApplication accessor is not KubernetesApplication"
```
golangcibot: const `errNotKubeApp` is unused (from `unused`)

pkg/reconciler/oam/workload/translate.go
```diff
@@ -0,0 +1,151 @@
...
+
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+const (
+	errInjectService = "unable to inject Service in objects"
```
golangcibot: const `errInjectService` is unused (from `unused`)

pkg/reconciler/oam/trait/reconciler.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+
+	r.record.Event(trait, event.Normal(reasonTraitModify, "Successfully modifed workload translation"))
+	log.Debug("Successfully modified referenced workload", "kind", trait.GetObjectKind().GroupVersionKind().String())
+
+	trait.SetConditions(v1alpha1.ReconcileSuccess())
+	return reconcile.Result{RequeueAfter: longWait}, errors.Wrap(r.client.Status().Update(ctx, trait), errUpdateTraitStatus)
+}
+
+func workloadGroupKind(w oamv1alpha2.WorkloadReference) string {
```
golangcibot: func `workloadGroupKind` is unused (from `unused`)

pkg/reconciler/oam/trait/modify.go
```diff
@@ -0,0 +1,84 @@
...
+limitations under the License.
+*/
+
+package trait
+
+import (
+	"context"
+	"reflect"
+
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
```
golangcibot: File is not `goimports`-ed with -local github.com/crossplane/crossplane (from `goimports`)
```suggestion

```

pkg/reconciler/oam/workload/translate.go
```diff
@@ -0,0 +1,151 @@
...
+limitations under the License.
+*/
+
+package workload
+
+import (
+	"context"
+	"reflect"
+
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
```
golangcibot: File is not `goimports`-ed with -local github.com/crossplane/crossplane (from `goimports`)
```suggestion

```

pkg/reconciler/oam/workload/translate_test.go
```diff
@@ -0,0 +1,267 @@
...
+func service(mod ...serviceModifier) *corev1.Service {
+	s := &corev1.Service{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       serviceKind,
+			APIVersion: serviceAPIVersion,
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: workloadName,
+			Labels: map[string]string{
+				labelKey: string(workloadUID),
```
golangcibot: unnecessary conversion (from `unconvert`)

pkg/resource/interfaces.go
```diff
@@ -139,20 +144,39 @@ type Managed interface {
...
+	Object
 
 	LocalConnectionSecretWriterTo
 	ManagedResourceReferencer
 
 	Conditioned
 }
+
+// A WorkloadReferencer is a type that has a Workload reference.
+type WorkloadReferencer interface {
```
hasheddan: This is what is causing our circular dependency. The immediate options I can think of for solving this problem are:

- Create a separate `crossplane-oam-runtime` repo for the reconcilers and interfaces added in this PR that is not imported by `crossplane`
- Move the OAM types out of core `crossplane` and let both `crossplane` and `crossplane-runtime` import them

negz: https://github.com/crossplane/crossplane/blob/206995/apis/oam/v1alpha2/core_types.go#L227

WorkloadReference, TraitReference, and ScopeReference are currently all identical, apart from their Godoc strings. It was pointed out during the original review that they could be a single type.

https://github.com/crossplane/crossplane-runtime/blob/c09220/apis/core/v1alpha1/resource.go#L60

We already have a few subtypes of ObjectReference in crossplane-runtime. I suggest we combine our OAM reference types into one and move it into crossplane-runtime. There's plenty of non-OAM places in the Crossplane codebase where we could use a reference type that requires a GVK and name and has an optional UID. `TypedReference` or something?

hasheddan: Great idea! Will update :)

pkg/reconciler/oam/workload/translate.go
negz: I'm not exactly sure where these should live, but I suspect crossplane-runtime is not the right place. Do we have other examples of functions in this Go module that operate on core Kubernetes resource kinds?

hasheddan: I don't think we do.. let's just keep them in `addon-oam-kubernetes-remote` for now :+1:

pkg/reconciler/oam/workload/translate.go
```diff
@@ -0,0 +1,93 @@
...
+	"reflect"
+
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
+)
+
+var (
+	serviceKind       = reflect.TypeOf(corev1.Service{}).Name()
+	serviceAPIVersion = corev1.SchemeGroupVersion.String()
```
golangcibot: `serviceAPIVersion` is unused (from `varcheck`)

pkg/reconciler/oam/workload/translate.go
```diff
@@ -0,0 +1,93 @@
...
+	"context"
+	"reflect"
+
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
+)
+
+var (
+	serviceKind       = reflect.TypeOf(corev1.Service{}).Name()
```
golangcibot: `serviceKind` is unused (from `varcheck`)

pkg/reconciler/oam/workload/translate.go
```diff
@@ -0,0 +1,93 @@
...
+	serviceAPIVersion = corev1.SchemeGroupVersion.String()
+)
+
+var labelKey = "workload.oam.crossplane.io"
```
golangcibot: `labelKey` is unused (from `varcheck`)

hasheddan:
@negz I have updated with the implementation of `TypedObjectReference` there are quite a few places where this turned out to be appropriate. This is a fairly significant (breaking, depending on how you define it) change, and will require updates to method generation in `crossplane-tools` and is changing APIs (though not in a way that should have a real effect on end users) in numerous places.

negz:
@hasheddan Would it be easier to proceed with this PR if we only updated OAM to use `TypedReference` for the time being? We could replace our many instances of `*ObjectReference` in a future PR.

hasheddan:
@negz I thought about that, but I just looked through `provider-gcp` and I actually feel as though this would be pretty mechanical with the updates. I am usually one to push for doing something early if we think it is inevitable, but would be happy to wait if you feel strongly :+1:

negz:
@hasheddan If you're signing up to push through all the changes I'm happy either way!

negz:
> @hasheddan If you're signing up to push through all the changes I'm happy either way!

Actually, on second thought I'd feel more comfortable reviewing and moving forward this if we broke it up into two changes. I'm not sure about some of the changes we're introducing here (e.g. variants like `TypedNamespaceNameOf`), and I don't think there's an urgency to replace all existing instances of *ObjectReference.

hasheddan:
@negz updated :+1:

negz:
Thanks @hasheddan!

Issue #138, opened 2020-03-30
Allow resource claims to be backed up and restored


negz:
### What problem are you facing?

When a managed resource and its resource claim are backed up and subsequently restored using a tool like https://velero.io they do not remain bound:

```
$ kubectl get mysqlinstances.database.crossplane.io -A
 NAMESPACE   NAME         STATUS   CLASS-KIND              CLASS-NAME                          RESOURCE-KIND      RESOURCE-NAME            AGE
 hasan       testme-sql            CloudSQLInstanceClass   hasan-cloudsqlinstanceclass-mysql   CloudSQLInstance   hasan-testme-sql-ngzkq   27m
 $ kubectl get CloudSQLInstance
 NAME                     STATUS    STATE      CLASS                               VERSION     AGE
 hasan-testme-sql-ngzkq   Unbound   RUNNABLE   hasan-cloudsqlinstanceclass-mysql   MYSQL_5_7   27m
```

### How could Crossplane help solve your problem?

It should be possible to backup a managed resource and its claim, then recreate them (in any order) and have them remain bound and functional.

jbw976:
Might be worth poking in velero issues or slack to get some advice on how this should be handled.  not restoring owner references seems a bit broken to me too but i'm sure it's something they thought through and they may have insight to share on how to handle this as a consumer of velero.  This is the first issue i found on the topic, wonder if there are more: https://github.com/vmware-tanzu/velero/issues/965

negz:
> not restoring owner references seems a bit broken to me too

@jbw976 I'm not sure I feel Velero's current behaviour is broken. I could see a case that we could be using controller references better.

Currently Crossplane relies on controller references to:

1. Provide a human readable (i.e. via `kubectl describe`) indication of a child->parent relationship.
1. Enable automatic garbage collection of orphaned children.
1. Avoid accidentally commandeering and overwriting resources that collide with desired children.

1 and 2 seem like uncontroversial uses for controller references based on my read of the [controller reference design document](https://github.com/kubernetes/community/blob/c529f24/contributors/design-proposals/api-machinery/controller-ref.md), but I'm not so sure about 3. 

In the third case we're using the controller reference to prevent the case in which:

1. A secret named `coolsecret` exists. It has nothing to do with Crossplane.
1. Someone creates a `MySQLInstance` with `spec.writeConnectionSecretToRef.name: coolsecret`. 
[...]

PR #139, opened 2020-04-02
Introduce a distinct secret type for connection secrets

negz:

Fixes https://github.com/crossplane/crossplane-runtime/issues/138

This PR introduces a new type of secret: `connection.crossplane.io/v1alpha`. Using a distinct type for connection secrets allows Crossplane to adopt orphaned connection secrets (i.e. those without controller references) without needing to be concerned about adopting an existing secret that has nothing to do with Crossplane. Refer to https://github.com/crossplane/crossplane-runtime/issues/138#issuecomment-607009508 for more context.

### Checklist

I have:

### DIFF
--- a/pkg/reconciler/oam/workload/reconciler.go
+++ b/pkg/reconciler/oam/workload/reconciler.go
@@ -120,7 +120,7 @@ func NewReconciler(m ctrl.Manager, workload resource.WorkloadKind, o ...Reconcil
 		client:      m.GetClient(),
 		newWorkload: nw,
 		workload:    TranslateFn(NoopTranslate),
-		applicator:  resource.NewAPIApplicator(m.GetClient()),
+		applicator:  resource.NewAPIPatchingApplicator(m.GetClient()),
 		// TODO(negz): Migrate to resource.ControlledBy
 		applyOpts: []resource.ApplyOption{resource.ControllersMustMatch()}, // nolint:staticcheck
 		log:       logging.NewNopLogger(),
--- a/pkg/reconciler/oam/workload/reconciler.go
+++ b/pkg/reconciler/oam/workload/reconciler.go
@@ -121,9 +121,10 @@ func NewReconciler(m ctrl.Manager, workload resource.WorkloadKind, o ...Reconcil
 		newWorkload: nw,
 		workload:    TranslateFn(NoopTranslate),
 		applicator:  resource.NewAPIApplicator(m.GetClient()),
-		applyOpts:   []resource.ApplyOption{resource.ControllersMustMatch()},
-		log:         logging.NewNopLogger(),
-		record:      event.NewNopRecorder(),
+		// TODO(negz): Migrate to resource.ControlledBy
+		applyOpts: []resource.ApplyOption{resource.ControllersMustMatch()}, // nolint:staticcheck
+		log:       logging.NewNopLogger(),
+		record:    event.NewNopRecorder(),
 	}
 
 	for _, ro := range o {
--- a/pkg/reconciler/oam/workload/reconciler.go
+++ b/pkg/reconciler/oam/workload/reconciler.go
@@ -120,7 +120,7 @@ func NewReconciler(m ctrl.Manager, workload resource.WorkloadKind, o ...Reconcil
 		client:      m.GetClient(),
 		newWorkload: nw,
 		workload:    TranslateFn(NoopTranslate),
-		applicator:  resource.ApplyFn(resource.Apply),
+		applicator:  resource.NewAPIApplicator(m.GetClient()),
 		applyOpts:   []resource.ApplyOption{resource.ControllersMustMatch()},
 		log:         logging.NewNopLogger(),
 		record:      event.NewNopRecorder(),
@@ -179,7 +179,7 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error) 
 		// so this label is not being used.
 		meta.AddLabels(o, map[string]string{lowerGroupKind(workload.GetObjectKind()): string(workload.GetUID())})
 
-		if err := r.applicator.Apply(ctx, r.client, o, r.applyOpts...); err != nil {
+		if err := r.applicator.Apply(ctx, o, r.applyOpts...); err != nil {
 			log.Debug("Cannot apply workload translation", "error", err, "requeue-after", time.Now().Add(shortWait))
 			r.record.Event(workload, event.Warning(reasonCannotApplyWorkloadTranslation, err))
 			workload.SetConditions(v1alpha1.ReconcileError(errors.Wrap(err, errApplyWorkloadTranslation)))
--- a/pkg/resource/api.go
+++ b/pkg/resource/api.go
@@ -18,10 +18,12 @@ package resource
 
 import (
 	"context"
+	"encoding/json"
 	"strings"
 
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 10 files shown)


negz:
Leaving this in draft for the time being as I still need to verify that it works as expected in practice. I may also tackle secret propagation (via `pkg/reconciler/secret`) in this PR.

pkg/resource/api.go
```diff
@@ -63,41 +67,119 @@ func (a *APIManagedConnectionPropagator) PropagateConnection(ctx context.Context
...
+	v := strings.Join([]string{to.GetNamespace(), to.GetName()}, AnnotationDelimiter)
+	meta.AddAnnotations(from, map[string]string{k: v})
+
+	return errors.Wrap(a.client.Update(ctx, from), errUpdateSecret)
+}
+
+// An APIPatchingApplicator applies changes to an object by either creating or
+// patching it in a Kubernetes API server.
+type APIPatchingApplicator struct {
+	client client.Client
```
muvaf, line 105: How about having a field with `client.Patch` interface to allow people use their own patch object so that they can choose patch type? `NewAPIPatchingApplicator` would initialize with the default `patch` struct we have.

pkg/resource/api.go
```diff
@@ -63,41 +67,119 @@ func (a *APIManagedConnectionPropagator) PropagateConnection(ctx context.Context
...
+	m, ok := o.(metav1.Object)
+	if !ok {
+		return errors.New("cannot access object metadata")
+	}
+
+	current := o.DeepCopyObject()
+
+	err := a.client.Get(ctx, types.NamespacedName{Name: m.GetName(), Namespace: m.GetNamespace()}, current)
+	if kerrors.IsNotFound(err) {
+		// TODO(negz): Apply ApplyOptions here too?
```
muvaf, line 171: I don't have a use case but I'd expect `ApplyOption`s that I supply to be called for all successful calls, which includes `Create`.

pkg/resource/api.go
```diff
@@ -63,41 +67,119 @@ func (a *APIManagedConnectionPropagator) PropagateConnection(ctx context.Context
...
-		}
-		cmcs.Data = mgcs.Data
-		meta.AddAnnotations(cmcs, map[string]string{
-			AnnotationKeyPropagateFromNamespace: mgcs.GetNamespace(),
-			AnnotationKeyPropagateFromName:      mgcs.GetName(),
-			AnnotationKeyPropagateFromUID:       string(mgcs.GetUID()),
-		})
-		return nil
-	}); err != nil {
+	to := LocalConnectionSecretFor(o, MustGetKind(o, a.typer))
```
muvaf, line 83: I think the returned secret from `LocalConnectionSecretFor` and `ConnectionSecretFor` should be of type `connection.crossplane.io/v1alpha1` for all this to work.

negz: Ah, good catch. I did that but it seems to have been lost somewhere along the way.

pkg/resource/api.go
negz: Me too. This is used in a few places though so I don't want to risk altering the default logic too much in this PR, just in case.

pkg/resource/api.go
negz: I like the idea, but have just added a TODO for the moment. I suspect we could introduce this in a future PR without any breaking API change via a variadic option to `NewAPIPatchingApplicator`.

negz:
I've now tested this by following the steps outlined in https://github.com/crossplane/crossplane-runtime/issues/138. i.e.:

1. Create a `PostgreSQLInstance` claim using a `CloudSQLInstanceClass`.
1. Wait for the `CloudSQLInstance` to provision and bind.
1. Backup all `PostgreSQLInstance`, `CloudSQLInstance`, and `Secret` resources using Velero.
1. Stop `provider-gcp` so that deleting things in Crossplane won't delete things in the cloud.
1. Manually delete the `PostgreSQLInstance` and `CloudSQLInstance` (i.e. remove their finalizers)
1. Start `provider-gcp` again.
1. Restore the backup made in step 3.
1. Confirm the `PostgreSQLInstance` and `CloudSQLInstance` adopt their existing external CloudSQL instance and reclaim ownership of their orphaned connection secrets.
1. Confirm that updating the `CloudSQLInstance` updates the external instance

```
$ kubectl get postgresqlinstance app-postgresql -o json|jq '.metadata.annotations["crossplane.io/external-name"]'
"default-app-postgresql-sm9rb"

$ gcloud --project crossplane-playground sql instances describe default-app-postgresql-sm9rb|head -n5
backendType: SECOND_GEN
connectionName: crossplane-playground:us-west1:default-app-postgresql-sm9rb

$ kubectl get postgresqlinstance app-postgresql -o json|jq .metadata.uid
"1a502c92-2b5d-4024-a398-da0af604dc40"
$ kubectl get secret postgresqlconn -o json|jq '.metadata.ownerReferences[0].uid'
"1a502c92-2b5d-4024-a398-da0af604dc40"

$ kubectl get cloudsqlinstance default-app-postgresql-sm9rb -o json|jq .metadata.uid
"bdf9883e-0d92-46ac-a8a0-b99052d1751e"
$ kubectl -n crossplane-system get secret c50999fd-7606-4768-ace6-e57424f0960c -o json|jq '.metadata.ownerReferences[0].uid'
"bdf9883e-0d92-46ac-a8a0-b99052d1751e"

$ kubectl edit cloudsqlinstance
# Change dataDiskSizeGb from 10 to 12 in vim...

$ kubectl describe cloudsqlinstance
# ...
Events:
  Type    Reason                   Age   From                                                 Message
  ----    ------                   ----  ----                                                 -------
  Normal  UpdatedExternalResource  9s    managed/cloudsqlinstance.database.gcp.crossplane.io  Successfully requested update of external resource


$ gcloud --project crossplane-playground sql instances describe default-app-postgresql-sm9rb|grep DiskSize
  dataDiskSizeGb: '12'
```

negz:
> I may also tackle secret propagation (via `pkg/reconciler/secret`) in this PR.

I decided I'd prefer to tackle that in a separate PR, so this one should be ready to review.

pkg/reconciler/oam/workload/reconciler.go
```diff
@@ -120,10 +120,11 @@ func NewReconciler(m ctrl.Manager, workload resource.WorkloadKind, o ...Reconcil
 		client:      m.GetClient(),
 		newWorkload: nw,
 		workload:    TranslateFn(NoopTranslate),
-		applicator:  resource.ApplyFn(resource.Apply),
-		applyOpts:   []resource.ApplyOption{resource.ControllersMustMatch()},
-		log:         logging.NewNopLogger(),
-		record:      event.NewNopRecorder(),
+		applicator:  resource.NewAPIPatchingApplicator(m.GetClient()),
+		// TODO(negz): Migrate to resource.ControlledBy
```
hasheddan, line 124: I can circle back around and make these updates :+1:

hasheddan:
One follow-up, am I correct in assuming that we should not actually use this until the secret reconciler is updated as well?

negz:
> am I correct in assuming that we should not actually use this until the secret reconciler is updated as well?

I think this is safe to use now. There's two known issues with regard to connection secrets:

1. Claims and resources won't adopt controllerless secrets after a backup and restore.
1. Secrets won't be constantly propagated due to differing UIDs after a backup and restore.

If we rolled out this change now (i.e. updated providers to use this code) the first issue would be fixed, and the second would be no worse than it currently is. That said, I don't think there's any urgency to go update all the providers right away; we may as well wait until both issues are fixed.

PR #144, opened 2020-04-09
Update OAM trait and workload reconcilers to use MustBeControllableBy

hasheddan:

In order to enable backup and restore of OAM traits and workloads, resources they own should be able to be regained control of if they are not already controlled by a different owner.

*Note: Keeping in Draft until desired behavior is verified in `addon-oam-kubernetes-remote`.*

### Checklist

I have:

### DIFF
--- a/pkg/reconciler/oam/trait/reconciler.go
+++ b/pkg/reconciler/oam/trait/reconciler.go
@@ -178,12 +178,11 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error) 
 	}
 
 	// The trait's referenced workload should always be translated in an
-	// object(s) that is controlled by the workload. In the case where an
-	// object(s) already exists in the same namespace and with the same name
-	// before it is created, this wll guard against modifying it.
-
-	// TODO(negz): Migrate to resource.ControlledBy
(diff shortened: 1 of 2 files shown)


pkg/resource/resource.go
```diff
@@ -283,6 +283,26 @@ func MustBeControllableBy(u types.UID) ApplyOption {
 	}
 }
 
+// MustBeControllableByDesiredController requires that the current object is
+// controllable by the desired object's controller. An object is controllable if
+// its controller reference matches desired object's controller's UID, or it has
+// no controller reference.
+func MustBeControllableByDesiredController() ApplyOption {
```
negz, line 290: I was hoping to remove this pattern in which our required controller is implicitly inferred from the desired object when I deprecated `ControllersMustMatch`. Is it possible to use `ControllabeBy` instead?

hasheddan: @negz It could be removed, but it would result in the user never being to override this in the `Workload` reconciler. So all `ApplyOptions` provided would be in addition to `MustBeControllableBy` (which I am assuming is what you are referring to as `ControllableBy`). Do you prefer that?

negz: I prefer that, at least for the time being.

PR #147, opened 2020-04-13
Make APIFinalizer interface compatible with all Kubernetes resources

muvaf:

We use the same functionality in number of places in crossplane-runtime and now I need it in Crossplane core, too. Since all implementations differ only by the function signature, having one and using it everywhere makes more sense.

### Checklist

I have:


muvaf:
> it's only used internally by claimbinding.Reconciler today

Yes, claim binding and managed reconciler had their own implemenations of the same thing. But since the function signature enforced claim and managed types, I don't think they were used somewhere else. Now both reconcilers will use this common one and I'll use it in Crossplane core, too. Since it's generic we'd want to treat this as public interface.

PR #149, opened 2020-04-14
apply should work for objects with only generateName

muvaf:

It's a valid creation call but `Apply` doesn't currently support it. The `Get` call returns error when there is no name.

### Checklist

I have:

### DIFF
--- a/pkg/resource/api.go
+++ b/pkg/resource/api.go
@@ -111,6 +111,10 @@ func (a *APIPatchingApplicator) Apply(ctx context.Context, o runtime.Object, ao 
 		return errors.New("cannot access object metadata")
 	}
 
+	if m.GetName() == "" && m.GetGenerateName() != "" {
+		return errors.Wrap(a.client.Create(ctx, o), "cannot create object")
+	}
+
 	desired := o.DeepCopyObject()
 
 	err := a.client.Get(ctx, types.NamespacedName{Name: m.GetName(), Namespace: m.GetNamespace()}, o)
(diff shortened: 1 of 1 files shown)


pkg/resource/api.go
```diff
@@ -111,6 +111,10 @@ func (a *APIPatchingApplicator) Apply(ctx context.Context, o runtime.Object, ao
 		return errors.New("cannot access object metadata")
 	}
 
+	if m.GetName() == "" && m.GetGenerateName() != "" {
+		return errors.Wrap(a.client.Create(ctx, o), "cannot create object")
+	}
```
hasheddan, line 116: Nitpick: Could we add this to the `kerrors.IsNotFound(err)` check below?

```
if kerrors.IsNotFound(err) || (err != nil && m.GetGenerateName() != "") {
}
```

muvaf: But the error is `name is required` rather than `resource not found`

hasheddan: Right, this would check if the error is `IsNotFound`, and if not, would check if `m.GetGenerateName()` was populated. If either were true it would attempt to create the object (the same thing you do above). If this is a specific error type then you could also replace the `err != nil` with something like `kerrrors.IsNameRequred(err)`

muvaf: Yeah. But it feels awkward to make the `Get` call when you can be 100% sure that it's going to fail by checking `m.GetName() == ""`.

hasheddan: Fair. I am fine either way, just seemed like some duplication of code :+1:

negz: Is it worth making this check a function in `pkg/meta`? `meta.PendingName` or something?

muvaf: @negz Do you have a another use-case in mind?

negz: Not really; happy to just leave this as it is. We can always move it out if we find other use cases.

muvaf: Yea, it's just I'm trying to be a bit wary of generic things with one consumer.

PR #150, opened 2020-04-14
Allow Paved objects to satify resource.Conditioned

negz:

I started working on this as part of an unstructured implementation of our cross resource reference resolution code. I presume it could be useful for our composition work too.

I was a little on the fence around whether `fieldpath.Paved` itself should expose these methods and thus satisfy the `resource.Conditioned` interface, or whether it should be wrapped in other type that uses `fieldpath.Paved` behind the scenes to satisfy the interface (and presumably eventually others).

Example usage:

```go
u := &unstructured.Unstructured{}
u.SetAPIVersion(v1alpha3.MySQLInstanceAPIVersion)
u.SetKind(v1alpha3.MySQLInstanceKind)
_ = client.Get(ctx, nn, u)

fieldpath.Pave(u.Object).SetConditions(runtimev1alpha1.Available())
_ = client.Update(ctx, u)
```

### Checklist

I have:

### DIFF
--- a/pkg/fieldpath/paved.go
+++ b/pkg/fieldpath/paved.go
@@ -20,6 +20,9 @@ import (
 	"encoding/json"
 
 	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
 )
 
 // A Paved JSON object supports getting and setting values by their field path.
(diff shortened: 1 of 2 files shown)


negz:
> I would typically think of trying to assert `resource.Conditioned` then call the methods.

Can you give an example of what you mean?

hasheddan:
@negz just something like this
```
c, ok := o.(resource.Conditioned)
if !ok {
	return errors.New("not a conditioned object")
}
c.SetConditions(v1alpha1.Condition{Type: v1alpha1.TypeReady})
```

negz:
@hasheddan That's only possible when `o` is an interface type that contains a type that satisfies the `resource.Conditioned` interface (e.g. has `GetCondition` and `SetConditions` methods).

hasheddan:
@negz gotcha, I was thinking in the case of unstructured we would attempt to unmarshal to an object then operate. Is the idea here to not have to do so?

negz:
> I was thinking in the case of unstructured we would attempt to unmarshal to an object then operate.

@hasheddan That would work too, presuming we had a compatible type to unmarshal to. This seems to be the approach @muvaf is taking with `InfraInstance` in https://github.com/crossplane/crossplane/pull/1412. In the case I wrote this for I didn't really want to have to plumb the scheme down to the caller to allow it to convert to a registered type. I'm not sure I want to pursue that case though, so I'll close this for now.

PR #146, opened 2020-04-13
Add a simpler cross resource reference API, with support for reference selection

negz:
### Description of your changes

Fixes https://github.com/crossplane/crossplane-runtime/issues/148

This commit deprecates (but still supports) the contemporary cross resource reference API, in which API types must satisfy the resource.AttributeReferencer type. It introduces a new implementation that supports both references (which explicitly name a managed resource to derive a field value from) and selectors (which set a referencer based on simple rules).

### Checklist

I have:
- [x] Run `make reviewable` to ensure this PR is ready for review.
- [x] Ensured this PR contains a neat, self documenting set of commits.
- [ ] Updated any relevant [documentation], [examples], or [release notes].
- [ ] Updated the RBAC permissions in [`clusterrole.yaml`] to include any new types.

[documentation]: https://github.com/crossplane/crossplane/tree/master/docs
[examples]: https://github.com/crossplane/crossplane/tree/master/cluster/examples
[release notes]: https://github.com/crossplane/crossplane/tree/master/PendingReleaseNotes.md
[`clusterrole.yaml`]: https://github.com/crossplane/crossplane/blob/master/cluster/charts/crossplane/templates/clusterrole.yaml

PR #151, opened 2020-04-15
Simplify resource references, and introduce reference selection

### LINKED ISSUE #76: Make AttributeReferencer a no-op when its target field is already set?

Crossplane supports [cross-resource-referencing](https://github.com/crossplaneio/crossplane/blob/29ea87447ce362c03f1383a93e969f84e37e0961/design/one-pager-cross-resource-referencing.md), such that:

```yaml
spec:
  # This reference field
  fooRef:
    name: someOtherResource
  # Can be used to set this target field
  foo: someString
```

Initially references were resolved only once. Resolving references successfully set the "references resolved" status condition to true, and resources would only be resolved when said condition was untrue. #66 removed this gate, instead resolving references on every reconcile. We quickly noticed that this caused #74 and raised #75 to address that. At the time of writing references are resolved on create and update attempt, but not on deletes.

This means:

* Reference fields take precedence over their target field; if both are set the reference field will overwrite the target field.
* We resolve references a bunch when we don't really need to. This will typically involve reading each referenced resource twice per reconcile, but the reads should be from cache.
* References are never resolved at delete time, even if they're stale.

I wouldn't call these problems as such, but I want to explore a different user experience.

If we instead:

* Invoked `ResolveReferences` on every reconcile (even deletes).
* Maintained the existing behaviour in which a "references access error" blocked the managed resource reconciler from proceeding to observe, create, update, or delete the external resource.
* Did not try to resolve `fooRef` at all when `foo` was set.

This would give the user finer control over if and when reference resolution happened. e.g.

* If you only want reference resolution to happen once set the reference field and don't set the target field. The target field will be set for you once, and not touched until it is deleted.
* If you want to force a refresh of the reference resolution for a specific field, edit the managed resource and unset its target value.

Going down this path may force us to use a "monolithic" alternative to the current `AttributeReferencer`, which would push more of the contract on the implementor e.g.:

```go
func (r *MonolithicReferencer) Resolve(ctx context.Context, c client.Reader, obj runtime.Object) ([]resource.ReferenceStatus, error) {
  // Referencer is a managed resource that references a Referenced managed resource.
  referencer, ok := obj.(*Referencer)
  if !ok {
    return nil, errNotReferencer
  }

  if referencer.Spec.TargetField != "" {
    return nil, nil
  }

  // Referenced is a managed resource referenced by a Referencer
  referenced := &Referenced{}
  if err := c.Get(ctx, types.NamespacedName{Name: r.Name}, referenced); err != nil {
    if kerrors.IsNotReady(err) {
      return []resource.ReferenceStatus{{Name: r.Name, Status: resource.ReferenceNotFound}}, nil
    }
    return nil, err
  }

  if !resource.IsConditionTrue(referenced.GetCondition(runtimev1alpha1.TypeReady)) {
    return []resource.ReferenceStatus{{Name: r.Name, Status: resource.ReferenceNotReady}}, nil
  }
  
  referencer.Spec.TargetField = referenced.Spec.ReferencedField
  return []resource.ReferenceStatus{{Name: r.Name, Status: resource.ReferenceReady}}, nil
}
```

### LINKED ISSUE #78: CRR: `ReferencesResolved` condition is still `False` even though the `Ready` condition is `True`
### What happened?

CRR: `ReferencesResolved` condition is still `False` even though the `Ready` condition is `True`.

### How can we reproduce it?

1. Install crossplane@0.4.1
1. Install stack-gcp@0.2.1
1. Setup gcp account provider
1. Run 
```bash
kubectl apply -k github.com/crossplaneio/crossplane//cluster/examples/workloads/kubernetes/wordpress/gcp/network-config?ref=release-0.4
```

*Observation*:
Run `kubectl get globaladdress.compute.gcp.crossplane.io/sample-globaladdress -o yaml` to observe the behavior:

```yaml
apiVersion: compute.gcp.crossplane.io/v1alpha3
kind: GlobalAddress
metadata:
  annotations:
    crossplane.io/external-name: sample-globaladdress
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"compute.gcp.crossplane.io/v1alpha3","kind":"GlobalAddress","metadata":{"annotations":{},"name":"sample-globaladdress"},"spec":{"addressType":"INTERNAL","name":"my-cool-globaladdress","networkRef":{"name":"sample-network"},"prefixLength":16,"providerRef":{"name":"gcp-provider"},"purpose":"VPC_PEERING","reclaimPolicy":"Delete"}}
  creationTimestamp: "2019-11-14T06:06:03Z"
  finalizers:
  - finalizer.managedresource.crossplane.io
  generation: 3
  name: sample-globaladdress
  resourceVersion: "9686"
  selfLink: /apis/compute.gcp.crossplane.io/v1alpha3/globaladdresses/sample-globaladdress
  uid: 3274f21b-96e6-477e-9ded-85c4178f6a02
spec:
  address: 10.29.0.0
  addressType: INTERNAL
  name: my-cool-globaladdress
  network: projects/crossplane-playground/global/networks/my-cool-network
  networkRef:
    name: sample-network
  prefixLength: 16
  providerRef:
    name: gcp-provider
  purpose: VPC_PEERING
  reclaimPolicy: Delete
status:
  conditions:
  - lastTransitionTime: "2019-11-14T06:06:03Z"
    message: '[{reference:sample-network status:NotFound}]'
    reason: One or more referenced resources do not exist, or are not yet Ready
    status: "False"
    type: ReferencesResolved
  - lastTransitionTime: "2019-11-14T06:06:34Z"
    reason: Managed resource is available for use
    status: "True"
    type: Ready
  - lastTransitionTime: "2019-11-14T06:06:34Z"
    reason: Successfully reconciled managed resource
    status: "True"
    type: Synced
  creationTimestamp: "2019-11-13T22:06:34.318-08:00"
  id: 8409971861024231429
  selfLink: https://www.googleapis.com/compute/v1/projects/crossplane-playground/global/addresses/my-cool-globaladdress
  status: RESERVED
```

crossplane-runtime version: v0.2.1

### LINKED ISSUE #148: Implement selectors for cross resource references
Currently Crossplane supports "Cross resource referencing" in which one resource may reference another in order to derive the value of a field. For example a VPC Subnet managed resource may reference a VPC Network managed resource in order to specify which network it should be created in. This only works when the referenced resource already exists, or will be created with a predictable, deterministic name. In some cases we'd like to create two resources at the same time and have one reference the other, even if we don't know its name before creation time. See https://github.com/crossplane/crossplane/pull/1163 for more context.

Crossplane could add a selector for each existing cross resource reference, allowing the reference to be selected and set using labels or other criteria as appropriate.

negz:

Fixes #148
Fixes #78
Fixes #76

This commit introduces `managed.APISimpleReferenceResolver`, which satisfies the `managed.ReferenceResolver` interface. This variant requires much less plumbing and reflection because it expects that managed resources expose a single `ResolveReferences` method that will optionally select and then resolve any resource references. It also adds a new pkg/reference which contains a library that managed resource authors _may_ choose to use to avoid reimplementing common reference selection and resolution logic.

The existing `managed.APIReferenceResolver` implementation remains the default, but is marked deprecated and will be removed once all managed resources use `APISimpleReferenceResolver`. Notably, the "reference resolution" condition is no longer set by `managed.Reconciler` - managed resources will report reference resolution issues via the `Synced` condition (i.e. `ReconcileError`).

See #146 for a prior draft approach, and discussion.

### Checklist

I have:


negz:
Below is an example of a managed resource that exposes a `ResolveReferences` method compatible with `APISimpleReferenceResolver`. This method uses `reference.NewAPIResolver` to avoid reimplementing common reference selection and resolution logic, but could supply its own bespoke implementation (or an improved common implementation) if appropriate.

```go
// ResolveReferences of this MySQLServerVirtualNetworkRule.
func (mg *MySQLServerVirtualNetworkRule) ResolveReferences(ctx context.Context, c client.Reader) error {
	r := reference.NewAPIResolver(c, mg)

	// Resolve spec.resourceGroupName
	rsp, err := r.Resolve(ctx, reference.ResolutionRequest{
		CurrentValue: mg.Spec.ResourceGroupName,
		Reference:    mg.Spec.ResourceGroupNameRef,
		Selector:     mg.Spec.ResourceGroupNameSelector,
		To:           reference.To{Managed: &v1alpha3.ResourceGroup{}, List: &v1alpha3.ResourceGroupList{}},
		Extract:      reference.ExternalName(),
	})
	if err != nil {
		return err
	}
	mg.Spec.ResourceGroupName = rsp.ResolvedValue
	mg.Spec.ResourceGroupNameRef = rsp.ResolvedReference

	// Resolve spec.virtualNetworkSubnetID
	rsp, err = r.Resolve(ctx, reference.ResolutionRequest{
		CurrentValue: mg.Spec.VirtualNetworkSubnetID,
		Reference:    mg.Spec.VirtualNetworkSubnetIDRef,
		Selector:     mg.Spec.VirtualNetworkSubnetIDSelector,
		To:           reference.To{Managed: &networkv1alpha3.Subnet{}, List: &networkv1alpha3.SubnetList{}},
		Extract:      networkv1alpha3.SubnetID(),
	})
	if err != nil {
		return err
	}
	mg.Spec.VirtualNetworkSubnetID = rsp.ResolvedValue
	mg.Spec.VirtualNetworkSubnetIDRef = rsp.ResolvedReference

	// Resolve spec.serverName.
	rsp, err = r.Resolve(ctx, reference.ResolutionRequest{
		CurrentValue: mg.Spec.ServerName,
		Reference:    mg.Spec.ServerNameRef,
		Selector:     mg.Spec.ServerNameSelector,
		To:           reference.To{Managed: &v1beta1.MySQLServer{}, List: &v1beta1.MySQLServerList{}},
		Extract:      reference.ExternalName(),
	})
	if err != nil {
		return err
	}
	mg.Spec.ServerName = rsp.ResolvedValue
	mg.Spec.ServerNameRef = rsp.ResolvedReference

	return nil
}
```

negz:
Below is an example of a managed resource that cannot resolve its references. References are resolved serially, so we'll only report the first unresolvable reference (as opposed to the previous behaviour of reporting all unresolvable references at any point in time). 

```bash
$ kubectl describe mysqlserver
# ...
Status:
  Conditions:
    Last Transition Time:  2020-04-15T06:20:39Z
    Message:               cannot resolve references: cannot get managed resource: ResourceGroup.azure.crossplane.io "mysql-example" not found
    Reason:                Encountered an error during resource reconciliation
    Status:                False
    Type:                  Synced
Events:
  Type     Reason                           Age                     From                                              Message
  ----     ------                           ----                    ----                                              -------
  Warning  CannotResolveResourceReferences  7s                      managed/mysqlserver.database.azure.crossplane.io  cannot resolve references: cannot get managed resource: ResourceGroup.azure.crossplane.io "mysql-example" not found
```

apis/core/v1alpha1/condition.go
```diff
@@ -285,6 +287,8 @@ func ReferenceResolutionSuccess() Condition {
 // unable to resolve the references used in the resource. This could
 // mean that one or more of referred resources do not yet exist, or are not yet
 // Ready.
+//
+// Deprecated: Use ReconcileError.
```
hasheddan, line 291: Since we are expecting for some resources that they may be in a state of waiting on other resources for a non-trivial period of time, does it make sense to make blocked on reference resolution a reconcile error condition? It feels like it is still successful reconciliation.

negz: My feeling is that there's less value in having a distinct condition for references now that we also emit a `CannotResolveResourceReferences` event.

It's also worth noting that `reference.NewAPIResolver` doesn't check the `Ready` condition - so the only reason we'd report a reconcile error would be if you referenced a managed resource that did not actually exist. If this persists for a non-trivial amount of time I'd suspect it was actually a configuration error (i.e. you typoed your reference).

negz:
> I actually enjoyed the UX of seeing them get resolved one by one in the condition of the resource, but I may be in the minority here.

@hasheddan One thing I struggled with about the contemporary UX was that the error/condition only gave the names of the references, not what references they were or what kind of resource they pointed to. I experimented with including this but the messages started to feel very long. I ended up opting for exposing only the first failed reference using a more verbose error that includes it kind. I like that this simplifies the implementation a little.

hasheddan:
@negz that seems reasonable, what about the usage of a separate condition?

negz:
@hasheddan what about the usage of a separate condition?

Could you illustrate what you mean?

hasheddan:
@negz I am just talking about using `ReconcileError` when references are not resolved:
https://github.com/crossplane/crossplane-runtime/pull/151#discussion_r409140092

PR #152, opened 2020-04-16
Connection secret publishers are made consumable by all resources with secret

muvaf:

This is a non-breaking change that simplifies the interface that the connection publishers require. It's necessary for composition instances to be able to use the existing connection publishers.

### Checklist

I have:

### DIFF
--- a/pkg/reconciler/managed/api.go
+++ b/pkg/reconciler/managed/api.go
@@ -72,21 +72,21 @@ func NewAPISecretPublisher(c client.Client, ot runtime.ObjectTyper) *APISecretPu
 // PublishConnection publishes the supplied ConnectionDetails to a Secret in the
 // same namespace as the supplied Managed resource. It is a no-op if the secret
 // already exists with the supplied ConnectionDetails.
-func (a *APISecretPublisher) PublishConnection(ctx context.Context, mg resource.Managed, c ConnectionDetails) error {
+func (a *APISecretPublisher) PublishConnection(ctx context.Context, owner resource.ConnectionSecretOwner, c ConnectionDetails) error {
 	// This resource does not want to expose a connection secret.
-	if mg.GetWriteConnectionSecretToReference() == nil {
+	if owner.GetWriteConnectionSecretToReference() == nil {
 		return nil
 	}
 
-	s := resource.ConnectionSecretFor(mg, resource.MustGetKind(mg, a.typer))
+	s := resource.ConnectionSecretFor(owner, resource.MustGetKind(owner, a.typer))
 	s.Data = c
-	return errors.Wrap(a.secret.Apply(ctx, s, resource.ConnectionSecretMustBeControllableBy(mg.GetUID())), errCreateOrUpdateSecret)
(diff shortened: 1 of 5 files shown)


pkg/reconciler/managed/api.go
```diff
@@ -72,21 +72,21 @@ func NewAPISecretPublisher(c client.Client, ot runtime.ObjectTyper) *APISecretPu
 // PublishConnection publishes the supplied ConnectionDetails to a Secret in the
 // same namespace as the supplied Managed resource. It is a no-op if the secret
 // already exists with the supplied ConnectionDetails.
-func (a *APISecretPublisher) PublishConnection(ctx context.Context, mg resource.Managed, c ConnectionDetails) error {
+func (a *APISecretPublisher) PublishConnection(ctx context.Context, owner resource.ConnectionSecretOwner, c ConnectionDetails) error {
```
hasheddan, line 75: Would it make sense to move `APISecretPublisher` and `PublisherChain` out of `reconciler/managed` if they are to potentially be consumed by other types of reconcilers? It would be more of a breaking change, but most controllers would not be affected as they are getting it from the default: https://github.com/crossplane/crossplane-runtime/blob/55a4bb6323bc74a61f89a2537d74f3d1cfcc682f/pkg/reconciler/managed/reconciler.go#L317

Per usual, if this is something we may want in the future then I would prefer we break now :)

muvaf: Yeah, I wanted to move it to somewhere else but couldn't find a better place. Happy to take suggestions. Maybe `pkg/resource/api.go`?

By the way, the resource that I'll be using this for is the cluster-scoped instance of the composition (in Crossplane core) which has similar traits to managed resource but that's a very loose argument.

hasheddan: `pkg/resource` sounds good to me :+1:

pkg/reconciler/managed/publisher.go
```diff
@@ -27,9 +27,9 @@ type PublisherChain []ConnectionPublisher
 
 // PublishConnection calls each ConnectionPublisher.PublishConnection serially. It returns the first error it
 // encounters, if any.
-func (pc PublisherChain) PublishConnection(ctx context.Context, mg resource.Managed, c ConnectionDetails) error {
+func (pc PublisherChain) PublishConnection(ctx context.Context, owner resource.ConnectionSecretOwner, c ConnectionDetails) error {
```
negz, line 30: Nit: I think `o resource.ConnectionSecretOwner` or `so resource.ConnectionSecretOwner` would be plenty for these relatively small methods given that the type clearly indicates what the variable is.

pkg/reconciler/managed/api.go
negz: Yeah, let's move these out of `pkg/reconciler/managed` if they will be used by other reconcilers. I suggest:

* Leave interface definitions in the package that uses them, and duplicate them elsewhere if necessary.
* Any implementation with API in the name goes in `pkg/resource/api.go`
* We avoid breaking changes by marking `managed.NewAPISecretPublisher` deprecated and having it call the new `resource.NewAPISecretPublisher`.

muvaf:
I have implemented a different publisher in crossplane core because it has to do some filtering of the keys, so, I don't need these implementations anymore. Closing the PR for now.

PR #153, opened 2020-04-17
Additional reference selection and resolution helpers

negz:

A small handful of our existing reference resolvers support resolving an array of references to an array of values. This PR adds a convenience method to handle this case (alongside a few trivial updates to `pkg/reference` in separate commits). I opted to closely match the existing `Resolve` method's logic in this new implementation, specifically:

* The selector is only considered if no references have been set.
* The references are only considered if no values have been set.

This means, like `Resolve`, `ResolveMultiple` will only resolve references once. The CR author must clear any existing values to trigger re-resolution of references, and must clear any values _and_ references to trigger re-selection of references.

### Checklist

I have:

### DIFF
--- a/pkg/reference/reference.go
+++ b/pkg/reference/reference.go
@@ -36,6 +36,26 @@ const (
 	errNoValue     = "referenced field was empty (reference may not yet be ready)"
 )
 
+// NOTE(negz): There are many equivalents of FromPtrValue and ToPtrValue
+// throughout the Crossplane codebase. We duplicate them here to reduce the
+// number of packages our API types have to import to support references.
+
+// FromPtrValue adapts a string pointer field for use as a CurrentValue.
+func FromPtrValue(v *string) string {
+	if v == nil {
+		return ""
+	}
+	return *v
+}
+
+// ToPtrValue adapts a ResolvedValue for use as a string pointer field.
+func ToPtrValue(v string) *string {
+	if v == "" {
+		return nil
+	}
+	return &v
+}
+
 // To indicates the kind of managed resource a reference is to.
 type To struct {
 	Managed resource.Managed
--- a/pkg/reference/reference.go
+++ b/pkg/reference/reference.go
@@ -115,13 +115,6 @@ func (rr ResolutionResponse) Validate() error {
 	return nil
 }
 
-// A Resolver selects and resolves references to managed resources.
-type Resolver interface {
-	// Resolve the supplied ResolutionRequest. The returned ResolutionResponse
-	// always contains valid values unless an error was returned.
-	Resolve(ctx context.Context, req ResolutionRequest) (ResolutionResponse, error)
-}
-
 // An APIResolver selects and resolves references to managed resources in the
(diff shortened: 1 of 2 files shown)


pkg/reference/reference.go
```diff
@@ -95,11 +115,54 @@ func (rr ResolutionResponse) Validate() error {
...
+// not be processed.
+func (rr MultiResolutionRequest) IsNoOp() bool {
+	// We don't resolve values that are already set; we effectively cache
+	// resolved values. The CR author can invalidate the cache and trigger a new
+	// resolution by explicitly clearing the resolved values. This is a little
+	// unintuitive for the APIMultiResolver but mimics the UX of the APIResolver
+	// and simplifies the overall mental model.
+	if len(rr.CurrentValues) > 0 {
+		return true
+	}
```
hasheddan, line 138: Just making sure I am understanding what is happening here: if somehow some of the values were set for a multi-resolution reference (this should not be possible I don't think using `ResolveMultiple` below), then we would not try to set the remainder of them. The way to get out of this state would be to set clear the set values on the CR and let resolution try to set all of them again. Am I getting that right?

pkg/reference/reference.go
```diff
@@ -36,6 +36,26 @@ const (
 	errNoValue     = "referenced field was empty (reference may not yet be ready)"
 )
 
+// NOTE(negz): There are many equivalents of FromPtrValue and ToPtrValue
+// throughout the Crossplane codebase. We duplicate them here to reduce the
+// number of packages our API types have to import to support references.
+
+// FromPtrValue adapts a string pointer field for use as a CurrentValue.
+func FromPtrValue(v *string) string {
```
muvaf, line 44: Nit: `FromPointer` and `ToPointer` maybe? I assume there won't be int64, bool etc. equivalents of these; but if that's the case I'd name them `FromStringPtr` and `ToStringPtr`.

negz: Yeah I imagine we probably won't be referencing any other types - the identifiers that references resolve to seem to be strings in all cases so far.

pkg/reference/reference.go
negz: Right.

I wondered whether this was the right approach, because I could imagine it being unintuitive to someone trying to add or update values by modifying the array of references. Some things I considered:

* I _think_ that in many cases the resolved values are immutable, even when they're an array. e.g. A Resource that accepts an array of subnet names often can't have have that array modified after the fact.
* Presuming we don't want to resolve all references on each reconcile it's tricky to know when we should cache them. Should we re-resolve when the array of references is longer than the array of values? Would this lead to an incorrect expectation that re-resolution would happen when a user edited an existing reference?

In the end it seemed conceptually simpler overall to maintain the same model as the single reference resolver; selector flows down to refs which flow down to values. Each is resolved once, and you need to "clear the cache" by removing the values, or the values and the refs, in order to trigger re-resolution of references.

negz:
I've tested this using the `provider-gcp` `Connection` resource.

PR #154, opened 2020-04-21
Add composition interfaces and unstructured composition structs.

muvaf:

As part of composition implementation, the interfaces and the unstructured structs that implement those interfaces are moved to crossplane-runtime for re-use by others.

### Checklist

I have:

### DIFF
--- a/pkg/fieldpath/paved.go
+++ b/pkg/fieldpath/paved.go
@@ -103,6 +103,19 @@ func (p *Paved) GetValue(path string) (interface{}, error) {
 	return p.getValue(segments)
 }
 
+// GetValueInto unmarshals the retrieved value of the path into given object.
+func (p *Paved) GetValueInto(path string, out interface{}) error {
+	val, err := p.GetValue(path)
+	if err != nil {
+		return err
+	}
+	js, err := json.Marshal(val)
+	if err != nil {
+		return err
+	}
+	return json.Unmarshal(js, out)
+}
+
 // GetString value of the supplied field path.
 func (p *Paved) GetString(path string) (string, error) {
 	v, err := p.GetValue(path)
--- a/pkg/resource/interfaces.go
+++ b/pkg/resource/interfaces.go
@@ -21,6 +21,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 
 	"github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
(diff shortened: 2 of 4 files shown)


pkg/resource/interfaces.go
```diff
@@ -21,6 +21,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
negz, line 24: This looks like a dupe of the above `metav1` import.

pkg/fieldpath/paved.go
```diff
@@ -103,6 +103,19 @@ func (p *Paved) GetValue(path string) (interface{}, error) {
...
+	val, err := p.GetValue(path)
+	if err != nil {
+		return err
+	}
+	js, err := json.Marshal(val)
+	if err != nil {
+		return err
+	}
+	return json.Unmarshal(js, out)
+}
```
negz, line 117: It would be good to add a test for this - I can do that in a followup PR though.

pkg/resource/interfaces.go
```diff
@@ -210,3 +211,29 @@ type Workload interface {
...
+	ConnectionSecretWriterTo
+
+	SetCompositionSelector(*v1.LabelSelector)
+	GetCompositionSelector() *v1.LabelSelector
+
+	SetCompositionReference(*corev1.ObjectReference)
+	GetCompositionReference() *corev1.ObjectReference
+
+	SetResourceReferences([]corev1.ObjectReference)
+	GetResourceReferences() []corev1.ObjectReference
```
negz, line 230: Should we follow the existing pattern of declaring these last three groups of getters and setters as smaller interfaces and composing them into the larger one? Maybe `CompositionSelector`, `CompositionReferencer`, and `ComposedResourcesReferencer`?

pkg/resource/unstructured/composed.go
```diff
@@ -0,0 +1,91 @@
...
+	for _, f := range opts {
+		f(cr)
+	}
+	return cr
+}
+
+// Composed is used to operate on the composable resources whose schema
+// is not known beforehand.
+type Composed struct {
+	unstructured.Unstructured
```
negz, line 52: Would it be worth embedding the `runtime.Unstructured` interface here instead?

https://godoc.org/k8s.io/apimachinery/pkg/runtime#Unstructured

negz: Ah nevermind - I forgot that interface doesn't satisfy `metav1.Object`. We could create a new interface that composes `metav1.Object` and `runtime.Unstructured`, but there doesn't seem to be much benefit in doing so.

pkg/resource/unstructured/composed.go
```diff
@@ -0,0 +1,91 @@
...
+// GetUnstructured returns the underlying *unstructured.Unstructured.
+func (cr *Composed) GetUnstructured() *unstructured.Unstructured {
+	return &cr.Unstructured
+}
+
+// GetCondition of this Composed.
+func (cr *Composed) GetCondition(ct v1alpha1.ConditionType) v1alpha1.Condition {
+	conditioned := v1alpha1.ConditionedStatus{}
+	// The path is directly `status` because conditions are inline.
+	if err := fieldpath.Pave(cr.Object).GetValueInto("status", &conditioned); err != nil {
```
muvaf, line 64: @negz FWIW, I think having the functions accept a `map[string]interface{}` would be result in a simpler procedure IMO, similar to Kubernetes `unstructured` helpers like `SetNestedField` and others. `Pave` function doesn't do anything meaningful. Or maybe it'd simply extend `unstructured.Unstructured` with the existing functions and the wrapper structs would embed `fieldpath.Paved` instead of `unstructured.Unstructured`. But I'd prefer the former where functions can be used anywhere directly if you're going to change it.

muvaf: It's also more clear that the set operations will be done in-place when the map is supplied as argument instead of constructing a new object using the map.

PR #155, opened 2020-04-21
Remove OAM code, which has moved to crossplane/oam-runtime

negz:
### Checklist

I have:


negz:
@hasheddan I'm thinking we don't need to worry too much about deprecating this code before removing it, as I suspect our unreleased OAM controllers were the only consumers. What do you think?

PR #156, opened 2020-04-21
Tweaks to the new composite and composed types

negz:

Small followups from https://github.com/crossplane/crossplane-runtime/pull/154.

### Checklist

I have:

### DIFF
--- a/pkg/resource/interfaces.go
+++ b/pkg/resource/interfaces.go
@@ -21,7 +21,6 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


pkg/resource/interfaces.go
```diff
@@ -216,18 +233,12 @@ type Workload interface {
 type Composite interface {
 	Object
 
-	Conditioned
-	Bindable
```
muvaf, line 220: Isn't `Bindable` necessary for the relation between namespaced publication and composite resource?

muvaf: Alright, just checked the design and remembered it was under `spec`. I feel like we should change the implementation instead of the interface since the function signatures will be the same but not so sure. Anyway, you'd add it when you implement the publication.

negz: @muvaf I thought we agreed to remove `bindingPhase` altogether?

PR #158, opened 2020-04-22
Add a fieldpath.PaveObject convenience method

negz:
### Checklist

I have:

### DIFF
--- a/pkg/fieldpath/paved.go
+++ b/pkg/fieldpath/paved.go
@@ -20,11 +20,19 @@ import (
 	"encoding/json"
 
 	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/runtime"
 )
 
 // A Paved JSON object supports getting and setting values by their field path.
 type Paved struct {
 	object map[string]interface{}
+}
+
+// PaveObject paves a runtime.Object, making it possible to get and set values
+// by field path.
+func PaveObject(o runtime.Object) (*Paved, error) {
+	u, err := runtime.DefaultUnstructuredConverter.ToUnstructured(o)
+	return Pave(u), errors.Wrap(err, "cannot convert object to unstructured data")
 }
 
 // Pave a JSON object, making it possible to get and set values by field path.


pkg/fieldpath/paved.go
```diff
@@ -20,13 +20,21 @@ import (
...
 
 // A Paved JSON object supports getting and setting values by their field path.
 type Paved struct {
 	object map[string]interface{}
 }
 
+// PaveObject paves a runtime.Object, making it possible to get and set values
+// by field path.
+func PaveObject(o runtime.Object) (*Paved, error) {
+	u, err := runtime.DefaultUnstructuredConverter.ToUnstructured(o)
```
muvaf, line 34: It's worth to check whether the object already has `UnstructuredContent() map[string]interface{}` so that no conversion is needed.

muvaf: Also, did you see my comment on https://github.com/crossplane/crossplane-runtime/pull/154#discussion_r412208408

hasheddan: @muvaf is it less expensive to assert the `runtime.Object` to `runtime.Unstructured` than to just always do conversion? Seems like someone would use `Pave(object map[string]interface{})` if they were dealing with unstructured, so we would likely end up always checking if the `runtime.Object` here was `runtime.Unstructured`, then having to do conversion anyway.

muvaf: I just saw [this comment](https://github.com/crossplane/crossplane/pull/1412#discussion_r412600807), apparently `ToUnstructured` already does that check.

> is it less expensive to assert the runtime.Object to runtime.Unstructured than to just always do conversion?

I think it is. Conversion does marshal/unmarshal json serialization operations while assertion is just a look-up.

> Seems like someone would use Pave(object map[string]interface{}) if they were dealing with unstructured, so we would likely end up always checking if the runtime.Object here was runtime.Unstructured, then having to do conversion anyway.

I doubt that. I can see users of `Paved` passing in strong-typed `runtime.Object`s when the path is given as string.

hasheddan: > I think it is. Conversion does marshal/unmarshal json serialization operations while assertion is just a look-up.

What I mean is that it is not less expensive if we are *also* going to be converting it most of the time anyway (i.e. we would be asserting and serializing).

> I doubt that. I can see users of Paved passing in strong-typed runtime.Objects when the path is given as string.

So they would not be dealing in unstructured in this case right? So they would use the `PaveObject` function, we would check to see if it had `UnstructuredContent()`, it would not because it was not unstructured, then we would convert.

muvaf: > What I mean is that it is not less expensive if we are also going to be converting it most of the time anyway (i.e. we would be asserting and serializing).

I don't follow. If it already has the `map[string]interface{}`, no serialization would take place. So, it's asserted that it has the map, then we don't convert. If it doesn't, then we will. In either case we pay the cost of look-up but in one case we don't pay for serialization. If you look at serialization code it has many assertions, too. So, one assertion is pretty negligible.

> So they would not be dealing in unstructured in this case right? So they would use the PaveObject function, we would check to see if it had UnstructuredContent(), it would not because it was not unstructured, then we would convert.

Yes. But it actually doesn't make much sense now that I think about it again. If it's a strong-typed object, there has to be a conversion back to the object. However, you can't get back the `map[string]interface{}` from the paved object so you actually have to do something like [this](https://github.com/crossplane/crossplane/pull/1412/files#diff-24bc60be8cb21948248f48b589bd3735R133).

PR #159, opened 2020-04-22
AddOwnerRef should update

muvaf:

AddOwnerReference should always set the owner ref array to be able to modify the existing reference. Otherwise it never updates the existing owner ref.

### Checklist

I have:

### DIFF
--- a/pkg/meta/meta.go
+++ b/pkg/meta/meta.go
@@ -109,6 +109,7 @@ func AddOwnerReference(o metav1.Object, r metav1.OwnerReference) {
 	for i := range refs {
 		if refs[i].UID == r.UID {
 			refs[i] = r
+			o.SetOwnerReferences(refs)
 			return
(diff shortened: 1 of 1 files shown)


negz:
Are you sure this is necessary? Typically I'd expect this to work per https://play.golang.org/p/KXo6vizZol2.

muvaf:
> Typically I'd expect this to work per https://play.golang.org/p/KXo6vizZol2.

I'm not able to access this as Go playground seems to be down.

> Are you sure this is necessary?

Since `GetOwnerReferences() []OwnerReference` returns a copy of the owner references `refs[i] = ref` does pretty much nothing. I've run into this in an edge case where the object was owner but not controller and the owner ref didn't get updated because set never called. It'd be more correct if we actually updated the owner ref if it already exists or skip the case where it already exists altogether.

PR #161, opened 2020-04-23
Expand support for unstructured types

negz:

This PR switches names around from `unstructured.Composite` (for example) to `composite.Unstructured`, mostly to allow several unstructured types to use identically named options like `WithGroupVersionKind`. It also adds a few getters and setters required for resource publications, and introduces the `resource.Requirement` type that represents an application's requirement for a published composite resource.

The PR also adds a new `resource.ConnectionPropagator`, which is effectively identical to the existing `resource.ManagedConnectionPropagator`, except that it propagates from any `ConnectionSecretOwner`, rather than requiring the much larger and more specific `Managed` interface. This allows this propagator to be used to propagate from `resource.Managed` to `resource.Claim` and also from `resource.Composite` to `resource.Requirement`.

### Checklist

I have:


pkg/resource/unstructured/composite/composite.go
```diff
@@ -0,0 +1,165 @@
+/*
```
negz, line 1: This should have been `git mv`-ed from `unstructured/composite.go`, sorry. :( I've become too trusting in git's ability to detect this automatically.

pkg/resource/interfaces.go
```diff
@@ -208,22 +220,36 @@ type Target interface {
...
+	Reclaimer
 	ConnectionSecretWriterTo
 
 	Conditioned
 }
 
-// Composable resources can be a resource in a composition.
-type Composable interface {
+// Composed resources can be a composed into a Composite resource.
+type Composed interface {
```
negz, line 238: Named this to match its unstructured equivalent, `composed.Unstructured`.

pkg/resource/unstructured/composite/composite.go
```diff
@@ -0,0 +1,165 @@
...
+	if err := fieldpath.Pave(c.Object).GetValueInto("spec.requirementRef", out); err != nil {
+		return nil
+	}
+	return out
+}
+
+// SetRequirementReference of this Composite resource.
+func (c *Unstructured) SetRequirementReference(ref *corev1.ObjectReference) {
+	_ = fieldpath.Pave(c.Object).SetValue("spec.requirementRef", ref)
+}
```
negz, line 100: The `requirementRef` getters and setters are new.

pkg/resource/unstructured/composite/composite.go
```diff
@@ -0,0 +1,165 @@
...
+// GetReclaimPolicy of this Composite resource.
+func (c *Unstructured) GetReclaimPolicy() v1alpha1.ReclaimPolicy {
+	s, _ := fieldpath.Pave(c.Object).GetString("spec.reclaimPolicy")
+	return v1alpha1.ReclaimPolicy(s)
+}
+
+// SetReclaimPolicy of this Composite resource.
+func (c *Unstructured) SetReclaimPolicy(p v1alpha1.ReclaimPolicy) {
+	_ = fieldpath.Pave(c.Object).SetValue("spec.reclaimPolicy", p)
+}
```
negz, line 146: The `reclaimPolicy` getters and setters are new.

pkg/resource/unstructured/unstructured.go
```diff
@@ -0,0 +1,36 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// An ObjectTyper returns the GroupVersionKind that a runtime.Object considers
+// itself to have.
+type ObjectTyper struct{}
+
+// ObjectKinds returns one GroupVersionKind of the supplied object. It always
+// return false and nil.
```
negz, line 29: We have a few types that use a `runtime.ObjectTyper` (i.e. a `scheme.Scheme`) to determine the GVK of an object, because not all `runtime.Objects` have useful `GetObjectKind()` methods. This allows those types to accept an `ObjectTyper` that just returns `GetObjectKind` under the hood.

muvaf: Actually `scheme.Scheme` already handles this: https://github.com/kubernetes/apimachinery/blob/master/pkg/runtime/scheme.go#L251

negz: Oh nice - will remove.

negz:
> I don't think composed.Unstructured is any better than unstructured.Composed really. Is there other kinds of benefits for having packages for each type here?

@muvaf Yeah, the rename isn't because I think it reads better but rather because reads about the same and it allows for namespaced option types. e.g. I wanted the new unstructured resource requirement type to have a `FromGroupVersionKind` option. I think that reads better with the tighter namespaces, e.g.:

```go
// If everything is in one package
r := unstructured.NewRequirement(unstructured.RequirementWithGroupVersionKind(gvk))
c := unstructured.NewComposite(unstructured.CompositeWithGroupVersionKind(gvk))

// If they're in different packages
r := requirement.New(requirement.WithGroupVersionKind(gvk))
c := composite.New(composite.WithGroupVersionKind(gvk))
```

pkg/resource/api.go
```diff
@@ -44,53 +44,70 @@ const (
 // An APIManagedConnectionPropagator propagates connection details by reading
 // them from and writing them to a Kubernetes API server.
 type APIManagedConnectionPropagator struct {
+	Propagator ConnectionPropagator
+}
+
+// NewAPIManagedConnectionPropagator returns a new APIConnectionPropagator.
+//
+// Deprecated: Use NewAPIConnectionPropagator.
+func NewAPIManagedConnectionPropagator(c client.Client, t runtime.ObjectTyper) *APIManagedConnectionPropagator {
```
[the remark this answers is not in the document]
negz: The interface does change - it now takes `ConnectionSecretOwner` instead of `Managed`.

pkg/resource/fake/mocks.go
```diff
@@ -157,18 +149,61 @@ func (m *CredentialsSecretReferencer) GetCredentialsSecretReference() v1alpha1.S
 	return m.Ref
 }
 
+// TODO(negz): Remove WorkloadReferencer? Is it still used anywhere?
```
hasheddan, line 152: @negz confirmed we can remove, would you mind going ahead and taking it out in this PR? ref: https://github.com/crossplane/oam-kubernetes-runtime/blob/0be8044c8384876133cb48fa33be4f46a1296694/pkg/oam/oam.go#L44

pkg/reconciler/claimbinding/reconciler.go
```diff
@@ -168,7 +168,11 @@ func defaultCRManaged(m manager.Manager) crManaged {
 			ManagedConfiguratorFn(ConfigureNames),
 			ManagedConfiguratorFn(ConfigureReclaimPolicy),
 		},
-		ManagedCreator:              NewAPIManagedCreator(m.GetClient(), m.GetScheme()),
+		ManagedCreator: NewAPIManagedCreator(m.GetClient(), m.GetScheme()),
+
+		// TODO(negz): Switch to ConnectionPropagator once this has been
+		// deprecated for a release or two.
+		//nolint:staticcheck
 		ManagedConnectionPropagator: resource.NewAPIManagedConnectionPropagator(m.GetClient(), m.GetScheme()),
```
muvaf, line 176: Why not use the new one? This is called by `Reconcile` and it's under our control.

hasheddan:
Looks like this is good to go now @negz :+1:

pkg/reconciler/claimbinding/reconciler.go
negz: I was confusing this `managed.ConnectionPublisher`, which is used in a few places outside crossplane-runtime (e.g. in older controllers or controllers that disable publishing via `managed.WithConnectionPublishers()`. I agree that this one is not likely to be used by anyone outside crossplane-runtime.

That said, I don't think it hurts to mark stuff as deprecated for at least one release before we delete it. It's been possible for some time to supply `WithManagedConnectionPropagator` when creating a claim binding controller. I know it's unlikely, but if someone _had_ done this it would be a better experience for this to be deprecated for a release or so before it straight up disappeared. Switching to `ConnectionPropagator` here would mean `WithManagedConnectionPropagator` would no longer work.

muvaf: Yeah I agree with you about marking stuff deprecated before deletion. What I meant here was this line is a usage of the old propagator that we can just replace with the new one. People would import the claim binding reconciler but they don't call its `r.ManagedConnectionPropagator.PropagateConnection()` function, it's the reconciler itself who is calling it. So, the public interface of the reconciler object doesn't change. Same for target reconciler.

PR #166, opened 2020-04-28
Add getters and setters for resource counts

negz:

Adding status.composedResources and status.readyResources will allow us to determine whether a composite resource (and thus its corresponding resource requirement) is ready by diffing the composed from the ready resources.

### Checklist

I have:

### DIFF
--- a/pkg/resource/unstructured/composite/composite.go
+++ b/pkg/resource/unstructured/composite/composite.go
@@ -154,6 +154,28 @@ func (c *Unstructured) SetReclaimPolicy(p v1alpha1.ReclaimPolicy) {
 	_ = fieldpath.Pave(c.Object).SetValue("spec.reclaimPolicy", p)
 }
 
+// SetComposedResourceCount of this Composite resource.
+func (c *Unstructured) SetComposedResourceCount(i int) {
+	_ = fieldpath.Pave(c.Object).SetNumber("status.composedResources", float64(i))
+}
+
+// GetComposedResourceCount of this Composite resource.
+func (c *Unstructured) GetComposedResourceCount() int {
+	v, _ := fieldpath.Pave(c.Object).GetNumber("status.composedResources")
+	return int(v)
+}
+
+// SetReadyResourceCount of this Composite resource.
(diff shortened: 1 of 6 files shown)


pkg/resource/unstructured/composite/composite.go
```diff
@@ -154,6 +154,28 @@ func (c *Unstructured) SetReclaimPolicy(p v1alpha1.ReclaimPolicy) {
 	_ = fieldpath.Pave(c.Object).SetValue("spec.reclaimPolicy", p)
 }
 
+// SetComposedResourceCount of this Composite resource.
+func (c *Unstructured) SetComposedResourceCount(i int) {
+	_ = fieldpath.Pave(c.Object).SetNumber("status.composedResources", float64(i))
+}
```
hasheddan, line 160: This is a broader question, but is the reason why we have to treat these new object types as unstructured rather than a type that satisfies an interface because we start controllers for them without having representations of their actual types? For instance, in the managed reconciler we have a go struct that we pass into the `NewReconciler` function, but here we are starting controllers on the fly for arbitrary object types. However, we still know they can satisfy these interfaces because they were produced by other controllers that we write. Do I have that correct?

muvaf:
I see the similarity between `Deployment`'s `replicas` and `readyReplicas` count but I don't think this would work well in this scenario. In `Deployment`, all the pods, i.e. composed resources, are identical, hence, counting is probably good enough. However, in our case, resources are usually of different kinds and not nearly similar. So, for the user seeing that the numbers are not equal doesn't really give much idea for the places they need to look.

As an alternative, I'd recommend expanding the `composedRef` reference object to include a readiness field so that the place to look for problems is immediately obvious.

negz:
I'm reverting this to a draft - I agree that the approach @muvaf suggests in https://github.com/crossplane/crossplane-runtime/pull/166#issuecomment-620623820 would be better than this approach (though adding readiness to `composedRefs` would make me feel even weirder about `composedRefs` being in `spec` and not `status`). When I opened this I thought it was the quickest path to being able to calculate whether a composite resource was ready, but it turned out it wasn't necessary for that at all.

PR #168, opened 2020-04-28
Update controller-runtime dependency to the latest release

muvaf:

It seems like there has been a release that includes https://github.com/kubernetes-sigs/controller-runtime/pull/863

### Checklist

I have:


muvaf:
@hasheddan I've just reverted updating `controller-tools` as it seems to change the CRDs by adding new fields like `x-kubernetes-int-or-string` to some CRDs, which I know we had some problems regarding backwards compatibility of these validation fields. I don't want to take the risk this close to the release so the PR now only updates the controller-runtime to 0.6.0.

hasheddan:
@muvaf sounds good, I think we need to start thinking about managing transition to v1 CRDs pretty soon so maybe we can circle back to this then 👍

PR #174, opened 2020-05-06
Don't wait for IsBindable before calling AddFinalizer

kasey:
Fixes https://github.com/crossplane/crossplane-runtime/issues/173

@muvaf please take a look. @negz intuition was that the issue reported in #173 is caused by the Finalizer never being added before deletion occurs. Walking through the code, that seems to add up, as we only call `AddFinalizer()` after `IsBindable()` evaluates true.

### Checklist

I have:

### DIFF
--- a/pkg/reconciler/claimbinding/reconciler.go
+++ b/pkg/reconciler/claimbinding/reconciler.go
@@ -384,6 +384,15 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error) 
 		return reconcile.Result{Requeue: false}, nil
 	}
 
+	if err := r.claim.AddFinalizer(ctx, claim); err != nil {
+		// If we didn't hit this error last time we'll be requeued
+		// implicitly due to the status update. Otherwise we want to retry
+		// after a brief wait, in case this was a transient error.
+		log.Debug("Cannot add resource claim finalizer", "error", err, "requeue-after", time.Now().Add(aShortWait))
+		claim.SetConditions(v1alpha1.Creating(), v1alpha1.ReconcileError(err))
+		return reconcile.Result{RequeueAfter: aShortWait}, errors.Wrap(r.client.Status().Update(ctx, claim), errUpdateClaimStatus)
+	}
+
 	// Claim reconcilers (should) watch for either claims with a resource ref,
 	// claims with a class ref, or managed resources with a claim ref. In the
 	// first case the managed resource always exists by the time we get here. In
@@ -470,15 +479,6 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error) 
 			log.Debug("Cannot propagate connection details from managed resource to claim", "error", err, "requeue-after", time.Now().Add(aShortWait))
 			record.Event(claim, event.Warning(reasonCannotPropagate, err))
 			claim.SetConditions(Binding(), v1alpha1.ReconcileError(err))
-			return reconcile.Result{RequeueAfter: aShortWait}, errors.Wrap(r.client.Status().Update(ctx, claim), errUpdateClaimStatus)
-		}
-
-		if err := r.claim.AddFinalizer(ctx, claim); err != nil {
-			// If we didn't hit this error last time we'll be requeued
-			// implicitly due to the status update. Otherwise we want to retry
-			// after a brief wait, in case this was a transient error.
-			log.Debug("Cannot add resource claim finalizer", "error", err, "requeue-after", time.Now().Add(aShortWait))
-			claim.SetConditions(v1alpha1.Creating(), v1alpha1.ReconcileError(err))
 			return reconcile.Result{RequeueAfter: aShortWait}, errors.Wrap(r.client.Status().Update(ctx, claim), errUpdateClaimStatus)
 		}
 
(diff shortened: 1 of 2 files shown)


negz:
I think this looks good based on my read of the logic.

Previously we only added the finalizer at bind time. This would ensure a finalizer was added to the claim whether it was bound to a statically or a dynamically provisioned managed resource. We'd never add a finalizer if we never bound. Most managed resources do not mark themselves bindable until they're successfully provisioned, so the reconcile loop would never be invoked to cleanup dynamically provisioned resources that never provisioned successfully.

We now add the finalizer immediately after the dynamic provisioning block. I believe this will ensure we'll either:

* Immediately add a finalizer after we've created the managed resource. (Dynamic case)
* Immediately add a finalizer after we've confirmed a managed resource exists. (Static case)

We often prefer to add the finalizer immediately before rather than immediately after the action that will require cleanup - i.e. we'd add the finalizer before creating the managed resource. Doing so covers the edge case in which we:

1. Successfully create the managed resource.
1. Fail to add the finalizer.
1. Are deleted, but never cleanup the managed resource because we never added the finalizer.

In this instance I feel okay about adding the finalizer immediately after the create call. There is no logic that could fail between the create call and the addition of the finalizer. If we fail to add the finalizer we'll try again on the next reconcile (i.e. in 30 seconds), so there's a relatively small window in which we risk leaking a managed resource. We _could_ add the finalizer right before we create a dynamically provisioned managed resource, but we'd have to add a second `AddFinalizer` call somewhere else to cover the static provisioning case. I prefer to go with the less complex option (the current implementation). We can always tweak it again if it becomes an issue in practice.

negz:
@kasey One thing I didn't consider while we worked through this was that we may want to ensure the managed resource appears to intend to bind to the claim when we call unbind, i.e at:

* https://github.com/crossplane/crossplane-runtime/blob/8207c4/pkg/reconciler/claimbinding/api.go#L111
* https://github.com/crossplane/crossplane-runtime/blob/8207c4/pkg/reconciler/claimbinding/api.go#L173

It's also quite the edge case, but my thinking is that in the current implementation a malicious party who had RBAC access to create and delete claims but not managed resources could delete (or unbind) managed resources they did not own using a claim, for example"

1. Managed resource A is happily bound to claim B.
1. Malicious party creates claim M and explicitly sets it `spec.resourceRef` to resource A.
1. Malicious party deletes claim M.

I believe in the above scenario Crossplane would happily call Unbind against resource A, which would at best unbind it from its legitimate claim and at worst delete it. I think we could work around this by adding a check at the beginning of the `Unbind` methods that ensured the `mg.GetClaimReference()` of the supplied managed resource was to the supplied claim.

negz:
> Any idea why @negz?

@hasheddan I suspect this is because @kasey initially accidentally pushed to a branch of crossplane/crossplane-runtime, rather than his fork.

hasheddan:
@negz your point about the edge case also makes me realize that I believe the current implementation allows for a dynamically provisioned resource to have another claim reference it (other than the one that triggered it being provisioned) before it becomes bound and then I presume they would race to bind to it. The one that did not provision it could win because we just overwrite the claim reference on the managed resource when we `Bind`:
https://github.com/crossplane/crossplane-runtime/blob/21d9cb6ea9d68c03d61d0d9c928036b79f173c47/pkg/reconciler/claimbinding/api.go#L92

This is once again quite the edge case, but I think it may make sense to do something like we do with connection secrets where both sides have to consent. That being said, not sure how much longer claims will be around so maybe it does not make sense to do too big of an architectural change.

negz:
> they would race to bind to it

It may make sense to add a check in the `Bind` methods too. We'd allow binding to a managed resource with a matching claim reference or no claim reference, but not a managed resource with a mismatched claim reference. I'm okay with making that change (and the Unbind) change in a subsequent PR.

muvaf:
> In this instance I feel okay about adding the finalizer immediately after the create call. There is no logic that could fail between the create call and the addition of the finalizer. If we fail to add the finalizer we'll try again on the next reconcile (i.e. in 30 seconds), so there's a relatively small window in which we risk leaking a managed resource. We could add the finalizer right before we create a dynamically provisioned managed resource, but we'd have to add a second AddFinalizer call somewhere else to cover the static provisioning case. 

The `AddFinalizer` call itself could fail. As you said it's not a common case but it'd be more robust if we can make sure we add the finalizer before any changes on the state of things other than the claim itself. Generally speaking, while the controller may leak some things if finalizer is added late, it never hurts to add it early as reconciliation loop and `Unbind` is idempotent. So, maybe we just move it up to line 386, before `if !meta.WasCreated(managed) && claim.GetClassReference() != nil {` statement to cover both creation and update cases. If anything fails at any point, our cleanup route is not affected and it will have a chance to reconcile and remove the finalizer.

negz:
> So, maybe we just move it up to line 386,

@muvaf That sounds like a better place for it. I like that line 386 means:

* We won't try to add a finalizer if we're being deleted.
* We'll add a finalizer before we attempt to dynamically provision a managed resource.
* We'll add a finalizer before we attempt to bind to a statically provisioned managed resource.
* We don't further increase the complexity of this method by calling `AddFinalizer` in two places.

@kasey Would you mind updating this?

kasey:
Updated @negz -- would you mind rerunning the smoke checks you mentioned?

negz:
I think this is low risk given the prior round of testing, so I'm going to optimistically merge it and test it together with https://github.com/crossplane/crossplane-runtime/pull/175.

PR #175, opened 2020-05-06
Bind and unbind managed resources conservatively

### LINKED ISSUE #177: Composed resources are still claimable

Claims bind to resources whose binding status is `Unbound` and composed resources never become `Bound`

Composed resources shouldn't be claimed by others.

negz:

Fixes #177

The two commits in this PR gate against a handful of possible issues:

1. Composite resource C composes managed resource M. Claim A explicitly sets its resource reference to managed resource M, thereby making M doubly owned by a claim and a composition.
1. Claim A dynamically provisions managed resource M. Claim B runs and wins a race against Claim A to bind to the newly provisioned and bindable M.
1. Claim A dynamically provisions managed resource M. Claim B explicitly sets its resource reference to M, and is then deleted, thereby deleting M which it was never bound to and did not dynamically provision.

We noticed both of these issues during review of https://github.com/crossplane/crossplane-runtime/pull/174 (which did not introduce them). I'd like to wait until that PR is merged before I manually test this.

Note (cc @prasek) that one implication of this change is that a resource claim may not claim a managed resource that was statically provisioned by an infrastructure stack. i.e. You could not author an infra stack that created a `CloudSQLInstance`, then later author a `MySQLInstance` that claimed that `CloudSQLInstance`. To my knowledge we don't do this today - infra stacks only provision "unclaimable" supporting infrastructure like VPC networks.

### Checklist

I have:

### DIFF
--- a/pkg/reconciler/claimbinding/api.go
+++ b/pkg/reconciler/claimbinding/api.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -38,6 +39,7 @@ const (
 	errDeleteManaged       = "cannot delete managed resource"
 	errBindMismatch        = "refusing to bind to managed resource that does not reference resource claim"
 	errUnbindMismatch      = "refusing to 'unbind' from managed resource that does not reference resource claim"
+	errBindControlled      = "refusing to bind to managed resource that is controlled by another resource"
 )
 
 // An APIManagedCreator creates resources by submitting them to a Kubernetes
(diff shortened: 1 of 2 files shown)


pkg/reconciler/claimbinding/api.go
```diff
@@ -190,3 +217,25 @@ func (a *APIStatusBinder) Unbind(ctx context.Context, _ resource.Claim, mg resou
...
+
+// equal returns true if the supplied object references are considered equal. We
+// consider two otherwise matching references with different UIDs to be equal,
+// presuming that they are both references to a particular object that has been
+// deleted and recreated, e.g. due to being restored from a backup.
+//
+// TODO(negz): If we switch to a reference that only has the fields we care
+// about (GVK, namespace, and name) we can just use struct equality.
+// https://github.com/crossplane/crossplane-runtime/issues/49
+func equal(a, b *corev1.ObjectReference) bool {
```
hasheddan, line 229: Do you think this could potentially be reused elsewhere? May make sense to export it and move to a more general pkg

negz: This seems simple enough that I'd rather do that if and when we find ourselves wanting to write it elsewhere, rather than ahead of time.

negz:
@muvaf @hasheddan Can you sanity check the commit I just added that guards against binding to managed resources with controller refs. Do we have any cases where a resource (e.g. a stack) creates and is the controller reference of a managed resource that is intended to be claimable?

negz:
I've rebased this on https://github.com/crossplane/crossplane-runtime/pull/174 (which is now merged) and tested that:

* Dynamic provisioning still works as expected.
* Static provisioning still works as expected.
* Deleting a claim that is trying but failing to dynamically provision a managed resource deletes said managed resource.
* I get an error when I try to nefariously 'unbind' a managed resource that is already bound to a different claim.
* I get an error when I nefariously try to race the claim that dynamically provisioned a managed resource to bind said managed resource.
* I get an error when I try to claim a managed resource that is part of a composition.

pkg/reconciler/claimbinding/api.go
```diff
@@ -141,6 +166,22 @@ func NewAPIStatusBinder(c client.Client, t runtime.ObjectTyper) *APIStatusBinder
 
 // Bind the supplied resource to the supplied claim.
 func (a *APIStatusBinder) Bind(ctx context.Context, cm resource.Claim, mg resource.Managed) error {
+	// A managed resource that was directly statically provisioned by an
+	// infrastructure operator should have a controller reference. We assume a
```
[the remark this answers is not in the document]
negz: Good catch - fixed the comments.

PR #176, opened 2020-05-07
Update with object that is gotten in APIUpdatingApplicator

hasheddan:

This changes APIUpdatingApplicator to make its Update call with the object it gets rather than the one passed so that the default behavior is a no-op rather than a guaranteed error on mismatched resource versions.

This adds an apply option that sets the resource version of the desired object to that of the current object such that it will force operations on update and patch calls.

~Keeping in draft until I can do some testing with Crossplane controllers.~

Please see https://github.com/crossplane/crossplane/pull/1492 for testing details.

### Checklist

I have:

### DIFF
--- a/pkg/resource/api.go
+++ b/pkg/resource/api.go
@@ -123,7 +123,8 @@ func NewAPIPatchingApplicator(c client.Client) *APIPatchingApplicator {
 }
 
 // Apply changes to the supplied object. The object will be created if it does
-// not exist, or patched if it does.
+// not exist, or patched if it does. If the object does it exist it will always
+// be patched, regardless of resource version.
 func (a *APIPatchingApplicator) Apply(ctx context.Context, o runtime.Object, ao ...ApplyOption) error {
 	m, ok := o.(metav1.Object)
 	if !ok {
@@ -173,7 +174,8 @@ func NewAPIUpdatingApplicator(c client.Client) *APIUpdatingApplicator {
 }
 
 // Apply changes to the supplied object. The object will be created if it does
-// not exist, or updated if it does.
+// not exist, or updated if it does. If the object does exist and no
+// ApplyOptions are passed, the update will be a no-op.
 func (a *APIUpdatingApplicator) Apply(ctx context.Context, o runtime.Object, ao ...ApplyOption) error {
 	m, ok := o.(metav1.Object)
 	if !ok {
@@ -201,7 +203,7 @@ func (a *APIUpdatingApplicator) Apply(ctx context.Context, o runtime.Object, ao 
 		}
 	}
 
-	return errors.Wrap(a.client.Update(ctx, o), "cannot update object")
+	return errors.Wrap(a.client.Update(ctx, current), "cannot update object")
 }
 
 // An APIFinalizer adds and removes finalizers to and from a resource.
--- a/pkg/resource/resource.go
+++ b/pkg/resource/resource.go
@@ -273,6 +273,15 @@ func (fn ApplyFn) Apply(ctx context.Context, o runtime.Object, ao ...ApplyOption
 // desired object. ApplyOptions are not called if no current object exists.
 type ApplyOption func(ctx context.Context, current, desired runtime.Object) error
 
+// UpdateFn returns an ApplyOption that is used to modify the current object to
+// match fields of the desired.
+func UpdateFn(fn func(current, desired runtime.Object)) ApplyOption {
+	return func(_ context.Context, c, d runtime.Object) error {
+		fn(c, d)
+		return nil
+	}
+}
+
 // MustBeControllableBy requires that the current object is controllable by an
 // object with the supplied UID. An object is controllable if its controller
 // reference matches the supplied UID, or it has no controller reference.
--- a/pkg/resource/api_test.go
+++ b/pkg/resource/api_test.go
@@ -18,6 +18,7 @@ package resource
 
 import (
 	"context"
+	"fmt"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
@@ -230,8 +231,8 @@ func TestPropagateConnection(t *testing.T) {
 
 func TestAPIPatchingApplicator(t *testing.T) {
 	errBoom := errors.New("boom")
-	named := &object{}
-	named.SetName("barry")
+	desired := &object{}
+	desired.SetName("desired")
 
 	type args struct {
 		ctx context.Context
@@ -317,15 +318,15 @@ func TestAPIPatchingApplicator(t *testing.T) {
 			c: &test.MockClient{
 				MockGet: test.NewMockGetFn(kerrors.NewNotFound(schema.GroupResource{}, "")),
 				MockCreate: test.NewMockCreateFn(nil, func(o runtime.Object) error {
-					*o.(*object) = *named
+					*o.(*object) = *desired
 					return nil
 				}),
 			},
 			args: args{
-				o: &object{},
+				o: desired,
 			},
 			want: want{
-				o: named,
+				o: desired,
 			},
 		},
 		"Patched": {
@@ -333,15 +334,15 @@ func TestAPIPatchingApplicator(t *testing.T) {
 			c: &test.MockClient{
 				MockGet: test.NewMockGetFn(nil),
 				MockPatch: test.NewMockPatchFn(nil, func(o runtime.Object) error {
-					*o.(*object) = *named
+					*o.(*object) = *desired
 					return nil
 				}),
 			},
 			args: args{
-				o: &object{},
+				o: desired,
 			},
 			want: want{
-				o: named,
+				o: desired,
 			},
 		},
 	}
@@ -362,8 +363,10 @@ func TestAPIPatchingApplicator(t *testing.T) {
 
 func TestAPIUpdatingApplicator(t *testing.T) {
 	errBoom := errors.New("boom")
-	named := &object{}
-	named.SetName("barry")
+	desired := &object{}
+	desired.SetName("desired")
+	current := &object{}
+	current.SetName("current")
 
 	type args struct {
 		ctx context.Context
@@ -449,31 +452,39 @@ func TestAPIUpdatingApplicator(t *testing.T) {
 			c: &test.MockClient{
 				MockGet: test.NewMockGetFn(kerrors.NewNotFound(schema.GroupResource{}, "")),
 				MockCreate: test.NewMockCreateFn(nil, func(o runtime.Object) error {
-					*o.(*object) = *named
+					*o.(*object) = *desired
 					return nil
 				}),
 			},
 			args: args{
-				o: &object{},
+				o: desired,
 			},
 			want: want{
-				o: named,
+				o: desired,
 			},
 		},
 		"Updated": {
-			reason: "No error should be returned if we successfully update an existing object",
+			reason: "No error should be returned if we successfully update an existing object. If no ApplyOption is passed the existing should not be modified",
 			c: &test.MockClient{
-				MockGet: test.NewMockGetFn(nil),
+				MockGet: test.NewMockGetFn(nil, func(o runtime.Object) error {
+					*o.(*object) = *current
+					fmt.Println(current.Name)
+					return nil
+				}),
 				MockUpdate: test.NewMockUpdateFn(nil, func(o runtime.Object) error {
-					*o.(*object) = *named
+					fmt.Println(current.Name)
+					fmt.Println(o.(*object).Name)
+					if diff := cmp.Diff(*current, *o.(*object)); diff != "" {
+						t.Errorf("r: -want, +got:\n%s", diff)
+					}
 					return nil
 				}),
 			},
 			args: args{
-				o: &object{},
+				o: desired,
 			},
 			want: want{
-				o: named,
+				o: desired,
 			},
 		},
 	}


pkg/resource/api.go
```diff
@@ -173,7 +174,8 @@ func NewAPIUpdatingApplicator(c client.Client) *APIUpdatingApplicator {
 }
 
 // Apply changes to the supplied object. The object will be created if it does
-// not exist, or updated if it does.
+// not exist, or updated if it does. If the object does exist it will always be
+// updated without respecting the resource version of the object that is passed.
```
hasheddan, line 178: @negz will you make sure these updated doc strings look okay to you? I was having some trouble choosing how to word this :thinking:

pkg/resource/api.go
```diff
@@ -184,24 +186,26 @@ func (a *APIUpdatingApplicator) Apply(ctx context.Context, o runtime.Object, ao
...
 
 	for _, fn := range ao {
-		if err := fn(ctx, current, o); err != nil {
+		if err := fn(ctx, o, desired); err != nil {
 			return err
 		}
 	}
 
-	return errors.Wrap(a.client.Update(ctx, o), "cannot update object")
+	desired.(metav1.Object).SetResourceVersion(o.(metav1.Object).GetResourceVersion())
```
muvaf, line 206: Could you talk about why we need to do this? From my reading of the code, this will wipe out everything in the remote resource, including `status`, and change it with the template we have. Then the controller in the remote cluster will reconcile and fill it up again; continuous battle of controllers. In fact, since we erase `spec`, there could be leak of actual resources.

In `Service` case, the IP is gone in the next reconcile; the LB controller might think that this resource is new and create a new LB resource without even deleting the old one. In case of our managed resource, `crossplane.io/external-name` will be wiped out if it's not given in template and the controller will recreate a new external resource in each reconcile.

IMO, `Update` should usually be the first choice over `Patch` if you have the `struct` of the target resource and can operate on it instead of dumping a static template. Since here we dump a static template, I think it should always use `Patch`. Even though `Patch` has its own problems, regarding lists, it's better than to encounter late-init problems only in arrays compared to whole metadata, spec, status.

pkg/resource/api.go
negz: LGTM

pkg/resource/api.go
negz: I think there is a case for two kinds of applicator:

1. Patching. We overlay our desired state on actual state, preserving any actual state that wasn't included in our desired state.
1. Updating. We replace all actual state with our desired state.

I agree that it wouldn't make sense for to use this updating applicator in many places. KARs for example, should not use this to update their remote resources for the reasons you point out. However I do think there are some cases where we want this:

1. Between KA and KAR. On each reconcile of a KA we want to render out all of its KARs using the templates in the KA. Nothing else ever writes to the spec or metadata of those KARs - it's safe to blow that all away on every reconcile. (We may need to consider KAR status though - I don't think we want to blow that away on each reconcile).
1. Propagated connection secrets. We don't care what the secret we're propagating to looks like. We actually _don't want to_ preserve fields that aren't in the desired state, we want to straight up overwrite the "to" secret with the "from" secret every time.

negz: > We may need to consider KAR status though

@muvaf @hasheddan I forget - does calling `client.Update(ctx, o)` with an empty status overwrite the status of the resource in the API server even if it's using the status subresource? If that's the case then this APIUpdatingApplicator is probably a lot less useful.

muvaf: > Between KA and KAR. On each reconcile of a KA we want to render out all of its KARs using the templates in the KA. Nothing else ever writes to the spec or metadata of those KARs - it's safe to blow that all away on every reconcile. (We may need to consider KAR status though - I don't think we want to blow that away on each reconcile).

I am not sure. It's the way things are right now that KAR is a resource whose metadata and spec don't change but that's only an assumption of today. For example, we might want to store a reference of the resource in the remote cluster in `KAR.spec`. It'll be gone in the next reconcile.

> Propagated connection secrets. We don't care what the secret we're propagating to looks like. We actually don't want to preserve fields that aren't in the desired state, we want to straight up overwrite the "to" secret with the "from" secret every time.

No, there are creation time information in those secrets that should be preserved because there is no way to fetch them from the provider again. I remember we were doing get->fill->update but now we rely on `Patch` behavior to keep the old keys intact.

I don't know, it really sounds like a bold assumption to make. It also renders any other controller who'd like to make changes on the CR useless.

I'd say that we should just never set `resourceVersion` if the change is not done on the retrieved object. If necessary, we can investigate doing the patch in-place through iterating over template and setting the values on the retrieved resource.

> does calling client.Update(ctx, o) with an empty status overwrite the status of the resource in the API server even if it's using the status subresource?

I don't think it does. I included `status` in my argument because first I though this will be used to update the resources in the remote cluster, which we don't know whether they have `status` as subresource.

negz: > I am not sure. It's the way things are right now that KAR is a resource whose metadata and spec don't change but that's only an assumption of today. For example, we might want to store a reference of the resource in the remote cluster in KAR.spec. It'll be gone in the next reconcile.

Shouldn't we worry about that if and when that becomes a use case?

> No, there are creation time information in those secrets that should be preserved because there is no way to fetch them from the provider again.

I think we're talking about different things. I'm discussing `pkg/reconciler/secret` and the managed resource -> claim secret propagation code. I think you're talking about the code that writes the "source of truth" managed resource secrets, not the various places we replicate those to.

negz: In my mind what we're doing here with update is not dramatically more dangerous than what we do in a patch. In either case we take desired state and apply it without concern as to the resource version of an existing resource. This seems reasonable to me as an option _as long as the caller knows what they're doing_. I'm not suggesting this replace Patch everywhere, or become the new default. Just that there seem to be cases where it could be useful.

hasheddan: I feel like doing this makes more sense than the current implementation as the default with the current will always fail unless you have already gotten the current, in which case you might as well call Update yourself. If we think this new behavior is too dangerous I would prefer to still change it to at least call Update with the object that we get so the default is a no-op rather than an error. That would make it more resemble CreateOrUpdate and you would be expected to pass in some sort of mutation in an ApplyOption. However, I think with appropriate warnings the behavior this PR implements is potentially useful.

muvaf: > Shouldn't we worry about that if and when that becomes a use case?

In general yes, but I think this is a case where we break something until a use case comes up rather than over-abstraction for future.

> I think we're talking about different things. I'm discussing pkg/reconciler/secret and the managed resource -> claim secret propagation code. I think you're talking about the code that writes the "source of truth" managed resource secrets, not the various places we replicate those to.

Yeah, I was thinking about managed -> claim.

> In my mind what we're doing here with update is not dramatically more dangerous than what we do in a patch. In either case we take desired state and apply it without concern as to the resource version of an existing resource. This seems reasonable to me as an option as long as the caller knows what they're doing.

> That would make it more resemble CreateOrUpdate and you would be expected to pass in some sort of mutation in an ApplyOption.

If they want that specifically, I think it's dangerous enough that it should be an `ApplyOption` for them to opt-in explicitly rather than doing it in all cases. By default, it will simply be like `CreateOrUpdate`, i.e. useful in cases where they first `Get` -> if NotFound, then dump from template, if Found then do some changes -> `Apply`. Maybe an option like `ForceUpdate()` or `OverrideResourceVersion()`.

hasheddan: Yeah we had talked about that pattern as well. If we go that route I think I would prefer us still change to Update with the object that we Get and then in the force override just set that object to the one we passed in but with the resource version. That way the default with no apply option is the no-op.

hasheddan: Any new thoughts here @negz @muvaf? I think I would be fine with the most recent proposal here.

negz: @hasheddan we (@muvaf and I) just discussed this out of band and we're okay moving forward with the most recent proposal. Thanks for bearing with us!

hasheddan:
@negz @muvaf this is now updated :) I'll run the test with Crossplane again to verify behavior, but should be mostly the same :+1:

pkg/resource/resource.go
```diff
@@ -319,6 +319,19 @@ func ConnectionSecretMustBeControllableBy(u types.UID) ApplyOption {
...
+// version that was set when the desired object was passed. It should only be
+// used in cases where it is desirable for the current object to be modified to
+// look exactly like the desired object.
+func OverrideResourceVersion() ApplyOption {
+	return func(_ context.Context, current, desired runtime.Object) error {
+		desired.(metav1.Object).SetResourceVersion(current.(metav1.Object).GetResourceVersion())
+		return nil
+	}
+}
+
```
hasheddan, line 334: @negz @muvaf  I was testing this out and realized that now that we make the Update with the resource that we Get, we no longer need to override the resource version. I think we should update this to be called `ForceUpdate` and have it set `current = desired`, then set the resource version back to what is was in `current` before. Or maybe we just don't need this right now as I am currently mimicking the existing behavior in KA:

```
func (c *applicationResourceClient) sync(ctx context.Context, template *v1alpha1.KubernetesApplicationResource) (bool, error) {
	submitted := false

	// ApplyOptions are executed in the order that they are passed, so by the
	// time we reach the anonymous function we will already know that this KAR
	// is controllable by the KA.
	err := resource.NewAPIUpdatingApplicator(c.kube).Apply(ctx, template, resource.MustBeControllableBy(metav1.GetControllerOf(template).UID), func(_ context.Context, current, _ runtime.Object) error {
		c, ok := current.(*v1alpha1.KubernetesApplicationResource)
		if !ok {
			return errors.New("current object is not a KubernetesApplicationResource")
		}

		if c.Status.State == v1alpha1.KubernetesApplicationResourceStateSubmitted {
			submitted = true
		}

		c.SetLabels(template.GetLabels())
		c.SetAnnotations(template.GetAnnotations())
		c.Spec = *template.Spec.DeepCopy()

		// By the time we get here Apply will have already checked that this KAR
		// is controllable by the KA in the MustBeControllableBy ApplyOption, so
		// it is safe to overwrite the owner references with the template's.
		c.SetOwnerReferences(template.OwnerReferences)

		return nil
	})
	return submitted, errors.Wrapf(err, "cannot sync %s", v1alpha1.KubernetesApplicationResourceKind)
}
```

hasheddan: Proposed:
```
// not exist, or updated if it does. If the object does exist and no
// ApplyOptions are passed, the update will be a no-op.
func (a *APIUpdatingApplicator) Apply(ctx context.Context, o runtime.Object, ao ...ApplyOption) error {
	m, ok := o.(metav1.Object)
	if !ok {
	}

	if m.GetName() == "" && m.GetGenerateName() != "" {
	}

	current := o.DeepCopyObject()

	err := a.client.Get(ctx, types.NamespacedName{Name: m.GetName(), Namespace: m.GetNamespace()}, current)
	if kerrors.IsNotFound(err) {
	}
	if err != nil {
	}

	for _, fn := range ao {
		if err := fn(ctx, current, o); err != nil {
			return err
		}
	}

	return errors.Wrap(a.client.Update(ctx, current), "cannot update object")
}
```

hasheddan: The only different from the existing implementation is the last line:
`return errors.Wrap(a.client.Update(ctx, o), "cannot update object")`
becomes -->
`return errors.Wrap(a.client.Update(ctx, current), "cannot update object")`

negz: @hasheddan So if I follow correctly:

* If the supplied object does not exist we'll create it (without calling our apply options).
* If the supplied object does exist, the caller _must_ supply an `ApplyOption` that, given `current, desired runtime.Object` must transform `current` based on the state in `desired`.

This feels similar to the UX of `controllerutil.CreateOrUpdate`, but has a few nice properties. Firstly it satisfies our `Applicator` interface so we can more easily mock it out. Secondly it offers what feels like an easier to understand control flow than `CreateOrUpdate`. i.e. This proposed function requires the caller to:

1. Tell me what the desired object is. (The supplied object.)
1. Tell me how to transform a current object into the desired object.

Whereas `CreateOrUpdate` requires the caller to pass a function that tells it how to transform an object in any state into the desired object.

This makes the UX a little different from the patching applicator (which doesn't require an ApplyOption in the case the applied object already exists), but I think it's a good approach. 

I wonder if an `UpdateFn` would help with readability and brevity? I suspect we won't need `ctx` or to `return error` in most cases.

```go
o := &corev1.Pod{
    // Form the desired object.
}

_ := c.Apply(ctx, o, resource.UpdateFn(func(current, desired runtime.Object) {
    // Update current to match desired.
}))
```

I imagine the `UpdateFn` would be something like:

```go
func UpdateFn(fn func(current, desired runtime.Object)) ApplyOption {
    return func(_ context.Context, c, d runtime.Object) error {
        fn(c, d)
        return nil
    }
}
```

hasheddan: @negz is the idea just to make it explicit that the body of the function is meant for the update operation? One could do this same thing with a normal `ApplyOption` and may choose to do additional modifications with additional `ApplyOptions`, which could be confusing. If we did introduce an `UpdateFn` I feel like it might be useful to call if after executing all `ApplyOptions` or something like that.

negz: > is the idea just to make it explicit that the body of the function is meant for the update operation?

Yes, and also to abbreviate the signature. I feel like this change will increase the places in which folks are supplying an anonymous function to `Apply` and I suspect in 99% of cases they won't need to supply a context or return an error. So:

```go
_ := c.Apply(ctx, o,
    resource.MustBeControllableBy(someUID),
    resource.UpdateFn(func(current, desired runtime.Object) {
        // Update current to match desired.
    }),
)
```

As opposed to:

```go
_ := c.Apply(ctx, o,
    resource.MustBeControllableBy(someUID),
    func(_ context.Context, current, desired runtime.Object) error {
        // Update current to match desired.
        return nil
    }),
)
```

I could be convinced that this doesn't add much value - just a thought.

hasheddan: @negz You've sold me :) I believe this is ready to go now. Here is the `crossplane` usage of it for KA if helpful:

```
err := resource.NewAPIUpdatingApplicator(c.kube).Apply(ctx, template, resource.MustBeControllableBy(metav1.GetControllerOf(template).UID), resource.UpdateFn(func(current, desired runtime.Object) {
	c := current.(*v1alpha1.KubernetesApplicationResource)
	if c.Status.State == v1alpha1.KubernetesApplicationResourceStateSubmitted {
		submitted = true
	}

	c.SetLabels(template.GetLabels())
	c.SetAnnotations(template.GetAnnotations())
	c.Spec = *template.Spec.DeepCopy()

	// By the time we get here Apply will have already checked that this KAR
	// is controllable by the KA in the MustBeControllableBy ApplyOption, so
	// it is safe to overwrite the owner references with the template's.
	c.SetOwnerReferences(template.OwnerReferences)
}))
```

Issue #49, opened 2019-10-13
Avoid corev1.ObjectReference


negz:
### What problem are you facing?

Most inter-resource references in Crossplane are either an [`ObjectReference`](https://godoc.org/k8s.io/api/core/v1#ObjectReference) or a [`LocalObjectReference`](https://godoc.org/k8s.io/api/core/v1#ObjectReference). `ObjectReference` contains many optional fields (e.g. `UID`, `FieldPath`, etc) that Crossplane never uses. This could be misleading to users, since some fields (i.e. `Name`) are always optional, and it's not obvious that the other fields are frequently not required when reading the API documentation or CRD specs.

### How could Crossplane help solve your problem?

Crossplane could use purpose-specific references that represent a subset of `ObjectReference`.

negz:
A related issue: https://github.com/kubernetes-sigs/service-catalog/issues/1411

muvaf:
I agree. A PR that implements all meta methods we have that are built on this object for other reference types we have could be helpful for adoption.

muvaf:
The refactor should include the interfaces where `corev1.ObjectReference` is accepted as parameter, too, like `SetCompositionReference(*corev1.ObjectReference)` etc.

PR #182, opened 2020-06-25
Use Reference type for provider references

hasheddan:

We are currently using corev1.ObjectReference for provider resources.
This includes more information than needed and encourages
using helper methods that may depend on other fields in the type that we
do not intend to be utilized. This updates provider references
fields to use the Reference type, which only has a name field.

Partial solution for #49 

Note that this will also require an update to `crossplane-tools` as mentioned in the issue.

An example of why this change is necessary is that we fail to get a provider when we use a helper function like `meta.NamespacedNameOf()` and the user has provided an extraneous `namespace` in the `providerRef`.

Slack thread: https://crossplane.slack.com/archives/CEG3T90A1/p1593112864471900

### Checklist

I have:


hasheddan:
I need to update interfaces here also

hasheddan:
Done :heavy_check_mark:

apis/core/v1alpha1/resource.go
```diff
@@ -197,7 +197,7 @@ type ResourceSpec struct {
 
 	// ProviderReference specifies the provider that will be used to create,
 	// observe, update, and delete this managed resource.
-	ProviderReference *corev1.ObjectReference `json:"providerRef"`
+	ProviderReference *Reference `json:"providerRef"`
```
muvaf, line 200: Shouldn't this field be required actually?

```suggestion
	ProviderReference Reference `json:"providerRef"`
```

If it wasn't a pointer type, then we'd not risk having [nil pointer dereference](https://github.com/crossplane/provider-aws/pull/261) when we use it, like here in your other PR https://github.com/crossplane/provider-aws/blob/87a255e5ee6d2b2d58711f20e5d8e199101c39e2/pkg/controller/cache/cachesubnetgroup/controller.go#L86

hasheddan: Great point! I'll update here :)

hasheddan:
@muvaf updated :+1:

PR #183, opened 2020-07-17
applicators: run supplied ApplyOptions for creations, too

muvaf:

While I was working on Crossplane Agent, I needed to copy a resource from another cluster. Since it has its own UID, resource version etc. I wanted to use an `ApplyOption` to get rid of those but realized they are not run on creation case.

### Checklist

I have:

### DIFF
--- a/pkg/resource/api.go
+++ b/pkg/resource/api.go
@@ -139,7 +139,11 @@ func (a *APIPatchingApplicator) Apply(ctx context.Context, o runtime.Object, ao 
 
 	err := a.client.Get(ctx, types.NamespacedName{Name: m.GetName(), Namespace: m.GetNamespace()}, o)
 	if kerrors.IsNotFound(err) {
-		// TODO(negz): Apply ApplyOptions here too?
+		for _, fn := range ao {
+			if err := fn(ctx, o, desired); err != nil {
+				return err
+			}
+		}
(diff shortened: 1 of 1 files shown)


pkg/resource/api.go
```diff
@@ -139,7 +139,11 @@ func (a *APIPatchingApplicator) Apply(ctx context.Context, o runtime.Object, ao
 
 	err := a.client.Get(ctx, types.NamespacedName{Name: m.GetName(), Namespace: m.GetNamespace()}, o)
 	if kerrors.IsNotFound(err) {
-		// TODO(negz): Apply ApplyOptions here too?
+		for _, fn := range ao {
```
hasheddan, line 142: @muvaf do you anticipate the set of options to always be the same for the `Create` and the `Patch`? It seems like a use case could arise for an option that is applicable for create but not for patch.

muvaf: Sort of, yes. But I converted the PR back to draft because in case of `NotFound` the object isn't changed, meaning it's still a deep copy of the initial one, hence, wouldn't work for my case where I'd need an empty object if `Create` and full object if `Update`.

muvaf:
Superseded by https://github.com/crossplane/crossplane-runtime/pull/184

PR #184, opened 2020-08-11
applicators: Run ApplyOptions against empty instance in case of creation

muvaf:

While I was working on Crossplane Agent, I needed to copy a resource from another cluster. Since it has its own UID, resource version etc. I wanted to use an `ApplyOption` to get rid of those but realized they are not run on creation case.

This PR makes sure options are called in all cases; in case the object doesn't exist, then `current` is the empty form of that object.

Unit tests are working and I tested it manually with the agent code.

### Checklist

I have:

### DIFF
--- a/pkg/resource/api.go
+++ b/pkg/resource/api.go
@@ -125,22 +125,38 @@ func NewAPIPatchingApplicator(c client.Client) *APIPatchingApplicator {
 // Apply changes to the supplied object. The object will be created if it does
 // not exist, or patched if it does. If the object does it exist it will always
 // be patched, regardless of resource version.
-func (a *APIPatchingApplicator) Apply(ctx context.Context, o runtime.Object, ao ...ApplyOption) error {
+func (a *APIPatchingApplicator) Apply(ctx context.Context, o runtime.Object, ao ...ApplyOption) error { //nolint:gocyclo
 	m, ok := o.(metav1.Object)
 	if !ok {
 		return errors.New("cannot access object metadata")
 	}
 
-	if m.GetName() == "" && m.GetGenerateName() != "" {
-		return errors.Wrap(a.client.Create(ctx, o), "cannot create object")
-	}
-
+	empty := NewEmptyInstance(o)
 	desired := o.DeepCopyObject()
 
+	if m.GetName() == "" && m.GetGenerateName() != "" {
+		// This for loop is written twice but run only once for an object. If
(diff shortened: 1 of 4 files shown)


pkg/resource/api.go
```diff
@@ -125,22 +125,38 @@ func NewAPIPatchingApplicator(c client.Client) *APIPatchingApplicator {
...
+	// form of the object that exists in the api-server. Hence, we stop using
+	// "empty" object the moment we realize the object already exists.
 	err := a.client.Get(ctx, types.NamespacedName{Name: m.GetName(), Namespace: m.GetNamespace()}, o)
 	if kerrors.IsNotFound(err) {
-		// TODO(negz): Apply ApplyOptions here too?
-		return errors.Wrap(a.client.Create(ctx, o), "cannot create object")
+		for _, fn := range ao {
+			if err := fn(ctx, empty, desired); err != nil {
+				return err
+			}
```
hasheddan, line 157: I mentioned this on the PR that was closed in favor of this one, but it feels weird to me to apply the same functions in both the `create` and `patch` cases. If a function is passed that overwrites the desired with some field on the existing, it appears that would be overwritten with an empty value in the create case. I guess a consumer could check for this in the function that they pass, is that how you envision this would be addressed @muvaf?

muvaf: > I mentioned this on the PR that was closed in favor of this one, but it feels weird to me to apply the same functions in both the create and patch cases. 

> I guess a consumer could check for this in the function that they pass, is that how you envision this would be addressed @muvaf

Yes, it's a little bit weird from the first look. But when you think about it, `Apply` concept itself inherently requires you to think about both `Create` and `Patch` at the same time, i.e. the object you're passing in has to work with both of them because `Apply` could do both of them. So, the implication of having to think about both cases is already there IMO. In fact, when I was coding the agent I was a little bit surprised that I pass `ApplyOption`s but they are called only in case of `Patch`.

> If a function is passed that overwrites the desired with some field on the existing, it appears that would be overwritten with an empty value in the create case.

That is actually one of the reasons I need this functionality; I fetch the resource from another cluster and run a `OverrideGeneratedMetadata` which takes UID, creation timestamp etc from the existing resource and puts it to metadata of the fetched resource, in case the resource doesn't exist I want those to be overridden with those empty values.

muvaf:
Not needed anymore.

Issue #179, opened 2020-05-15
Rename reclaimPolicy in managed resource to retentionPolicy


muvaf:
### What problem are you facing?

I have encountered this again while writing the 0.11 docs about managed resources.

`reclaimPolicy` on `PersistentVolume`s refer to what happens to the `PersistentVolume` when the claim(old claims or contemporary requirements in our case) is deleted. A different name for the current `reclaimPolicy` would make more sense since it decides on what happens to the actual cloud resource.

### How could Crossplane help solve your problem?

`retentionPolicy` or `deletionPolicy` would make more sense. Honestly, I'd be OK if we removed the field altogether and assumed managed resource is tightly coupled with external resource.

negz:
The field exists in two places today:

* Between requirement and composite. Determines whether the composite is deleted or retained when the requirement is deleted. Has no effect on whether external resources are deleted or kept.
* Between managed resource an external resource. Determines whether the external resource is kept when the managed resource is deleted.

I don't think we should keep the name `reclaimPolicy` for either, because the semantics and behavior have diverged from the upstream (PV) field of that name.

> Honestly, I'd be OK if we removed the field altogether and assumed managed resource is tightly coupled with external resource.

Do we need to support removing an external resource from Crossplane management? i.e. Crossplane has a bug, or the consumer just doesn't like it and wants to back out without losing their infrastructure. I'm okay with considering the lifecycle coupled as long as we have some story for this; perhaps just remove the Crossplane/provider deployments, remove the finalizer, then delete the CR?

muvaf:
> Between requirement and composite. Determines whether the composite is deleted or retained when the requirement is deleted. Has no effect on whether external resources are deleted or kept.

I think that matches the current behavior between PV and PVC. I'm not sure about whether we need to keep the a name that has `claim` in it. But the behavior is same so I'd be OK if we kept it for users to know what it is immediately.

> Do we need to support removing an external resource from Crossplane management? i.e. Crossplane has a bug, or the consumer just doesn't like it and wants to back out without losing their infrastructure.

I think we do, the use case you're describing makes sense to me. I remember a case where a user reached out saying cloudsql controller was sending requests to GCP API aggresively, we were able to provide a mitigation by fixing a field but that was just a nice coincidence. I can see the case where we promise a fix with a patch release and user doesn't want the controller to reconcile the resource until their setup is patched.

[...]

PR #185, opened 2020-08-12
Introduce a deletion policy, and deprecate the reclaim policy.

negz:

Partial fix for https://github.com/crossplane/crossplane-runtime/issues/179.
Adds code that will need to be cleaned up as part of https://github.com/crossplane/crossplane/issues/1670.

This PR introduces a `deletionPolicy` that is intended to replace the contemporary `reclaimPolicy` once it is removed. Corresponding changes will be required to angryjet.

### Checklist

I have:

### DIFF
--- a/apis/core/v1alpha1/policies.go
+++ b/apis/core/v1alpha1/policies.go
@@ -30,3 +30,17 @@ const (
 	// will be retained when the managed resource is deleted.
 	ReclaimRetain ReclaimPolicy = "Retain"
 )
+
+// A DeletionPolicy determines what should happen to the underlying external
+// resource when a managed resource is deleted.
+type DeletionPolicy string
+
+const (
+	// DeletionOrphan means the external resource will orphaned when its managed
+	// resource is deleted.
+	DeletionOrphan DeletionPolicy = "Orphan"
+
+	// DeletionDelete means both the  external resource will be deleted when its
+	// managed resource is deleted.
+	DeletionDelete DeletionPolicy = "Delete"
+)
--- a/pkg/reconciler/managed/reconciler.go
+++ b/pkg/reconciler/managed/reconciler.go
@@ -545,7 +545,7 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error) 
 	if meta.WasDeleted(managed) {
 		log = log.WithValues("deletion-timestamp", managed.GetDeletionTimestamp())
 
-		if observation.ResourceExists && managed.GetReclaimPolicy() == v1alpha1.ReclaimDelete {
+		if observation.ResourceExists && shouldDelete(managed) {
 			if err := external.Delete(externalCtx, managed); err != nil {
 				// We'll hit this condition if we can't delete our external
 				// resource, for example if our provider credentials don't have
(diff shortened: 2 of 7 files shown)


apis/core/v1alpha1/policies.go
```diff
@@ -30,3 +30,17 @@ const (
...
 )
+
+// A DeletionPolicy determines what should happen to the underlying external
+// resource when a managed resource is deleted.
+type DeletionPolicy string
+
+const (
+	// DeletionOrphan means the external resource will orphaned when its managed
+	// resource is deleted.
+	DeletionOrphan DeletionPolicy = "Orphan"
```
muvaf, line 41: Why not reuse `Retain`?

pkg/reconciler/managed/reconciler.go
```diff
@@ -695,3 +695,16 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
...
+// should be inlined when claims (and thus reclaim policies) are removed.
+func shouldDelete(mg resource.Managed) bool {
+	switch {
+	case mg.GetDeletionPolicy() == v1alpha1.DeletionDelete:
+		return true
+	case mg.GetDeletionPolicy() == v1alpha1.DeletionOrphan:
+		return false
+	}
+
+	return mg.GetReclaimPolicy() == v1alpha1.ReclaimDelete
```
muvaf, line 709: Should we rethink the default here? I feel like on this level, the expected behavior is to delete the external resource if I don't say otherwise because that's like lowest level. Also when you delete a `PV`, it doesn't have any policy or whatsoever to manage external deletion; it deletes the external volume immediately. So, deleting by default could feel familiar to people.

pkg/resource/interfaces.go
```diff
@@ -84,6 +84,12 @@ type Reclaimer interface {
 	GetReclaimPolicy() v1alpha1.ReclaimPolicy
 }
 
+// A Deletable may specify a DeletionPolicy.
+type Deletable interface {
```
muvaf, line 88: `Deletable` kind of implies that if the resource doesn't satisfy this interface, it can't be deleted. I don't really have great suggestions though. Maybe `HasDeletionPolicy`?

apis/core/v1alpha1/policies.go
negz: I felt 'Orphan' was more descriptive in this specific context. 'Retain it' could be interpreted as somehow retaining Crossplane control of the resource, while 'orphan it' seems clearer that the resource will be kept without Crossplane controlling it.

pkg/reconciler/managed/reconciler.go
negz: Whether a PV is deleted or not by default depends on whether or not it was dynamically provisioned, I believe. Orphaning the volume seems like the safest approach, but I agree that given these underlying resources will effectively always be dynamically provisioned (as a product of the composite resource). I could go either way on this one - it could be a little confusing for the deletion policy to have a different default from the reclaim policy during the transition period.

pkg/resource/interfaces.go
negz: I'm open to other suggestions, but I prefer `Deletable` to `HasDeletionPolicy`. Would `Orphanable` be preferable?

pkg/reconciler/managed/reconciler.go
negz: One thing to note here is that this would be a behaviour change during the transition period. Currently the default if you omit the reclaim policy is to retain the resource. During the transition if you omitted both policies the default would be to delete the resource.

pkg/resource/interfaces.go
negz: I went with `Orphanable`.

negz:
@hasheddan @muvaf I pushed some new commits in response to @muvaf's feedback. What do you think?

pkg/reconciler/managed/reconciler.go
muvaf: > Whether a PV is deleted or not by default depends on whether or not it was dynamically provisioned, I believe. 

I think we're talking about the EBS volume deletion here rather than `PersistentVolume` deletion here, i.e. the latter policy in PVC<>PV<>external resource. If you manually delete a `PersistentVolume`, it will always delete the external resource in the cloud, AFAIK there is no policy to control that behavior. So, calling `external.Delete` in case of deletion is a must in Kubernetes and we'd be making that behavior default.

> One thing to note here is that this would be a behaviour change during the transition period.

That's true. We're making behavior changes to both policy mechanisms, I guess we have to bite that anyway.

apis/core/v1alpha1/policies.go
muvaf: Yeah, makes sense.

PR #187, opened 2020-08-19
Rename 'requirement' to 'claim'

negz:

Crossplane composite resources are cluster scoped, but they can be 'published' to create a namespaced proxy resource. We called this resource a 'requirement', despite it being conceptually quite similar to our existing (and deprecated) 'resource claim' concept. We've found that the 'publish a requirement' concept has not resonated with the community and have decided to switch our terminology.

Under this new approach platform builders may choose to enable platform operators to 'offer' (not publish) a composite resource to their platform consumers. The namespaced interface to these composite resources will be known as a 'claim' or 'composite resource claim'. Note that we think platform builders and operators are the key audience for these concepts; platform consumers will simply think of themselves as using the resource as its kind indicates - e.g. 'a Kubernetes cluster' or 'an SQL instance', not 'an SQL instance claim'.

In some cases our existing but deprecated resource claim concept has name conflicts with this new take on the claim concept - i.e. the resource.Claim interface. In those cases I've named the new type CompositeClaim to distinguish it.

### Checklist

I have:

### DIFF
--- a/pkg/resource/fake/mocks.go
+++ b/pkg/resource/fake/mocks.go
@@ -185,15 +185,6 @@ func (m *CompositeResourceReferencer) SetResourceReference(p *corev1.ObjectRefer
 // GetResourceReference gets the composite resource reference.
 func (m *CompositeResourceReferencer) GetResourceReference() *corev1.ObjectReference { return m.Ref }
 
-// RequirementReferencer is a mock that implements RequirementReferencer interface.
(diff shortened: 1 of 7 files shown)


negz:
I just noticed that the composite resource interface still expected composite resources to be reclaimable, which they are not. This would probably fit best in a separate PR but I'm going to try convince you to let it through as a separate commit in this one as that makes my life a little easier while I have my c/c branch pointed at this one. :)

PR #189, opened 2020-08-24
Add reference for ProviderConfig objects and deprecate Provider reference

muvaf:

We have decided to use `ProviderConfig` for the object that contains the information about how the provider should authenticate itself to its API.

Part of https://github.com/crossplane/crossplane/issues/1684

### Checklist

I have:


pkg/resource/resource.go
```diff
@@ -360,7 +360,10 @@ func GetExternalTags(mg Managed) map[string]string {
 	if mg.GetClassReference() != nil {
 		tags[ExternalResourceTagKeyClass] = mg.GetClassReference().Name
 	}
-	if mg.GetProviderReference().Name != "" {
+	switch {
+	case mg.GetProviderConfigReference() != nil && mg.GetProviderConfigReference().Name != "":
+		tags[ExternalResourceTagKeyProvider] = mg.GetProviderConfigReference().Name
+	case mg.GetProviderReference() != nil && mg.GetProviderReference().Name != "":
```
hasheddan, line 366: nitpick: would you mind adding a note or todo here to remove this branch once `Provider` type is removed?

apis/core/v1alpha1/resource.go
```diff
@@ -197,9 +197,14 @@ type ResourceSpec struct {
 	// +optional
 	ClassReference *corev1.ObjectReference `json:"classRef,omitempty"`
 
+	// ProviderConfigReference specifies the provider that will be used to create,
+	// observe, update, and delete this managed resource.
```
negz, line 201:
```suggestion
	// ProviderConfigReference specifies how the provider that will be used to create,
	// observe, update, and delete this managed resource should be configured.
```

apis/core/v1alpha1/resource.go
```diff
@@ -197,9 +197,14 @@ type ResourceSpec struct {
...
+	ProviderConfigReference *Reference `json:"providerConfigRef,omitempty"`
```
negz, line 202: Do you think we could get away with something shorter like `pcRef` here? `providerConfigRef` is probably better - just thinking out loud.

muvaf: `pcRef` sounds a bit cryptic to me. I'd go with `providerConfigRef`.

muvaf:
@hasheddan @negz all comments are addressed. Could you look at it again?

PR #192, opened 2020-09-04
Update with desired object in APIUpdatingApplicator

### LINKED ISSUE #191: APIUpdatingApplicator behavior is unexpected
### What happened?

`NewAPIUpdatingApplicator` moved from broken (see https://github.com/crossplane/crossplane-runtime/issues/165) to just having unexpected behavior in https://github.com/crossplane/crossplane-runtime/pull/176. The original proposal in that PR was to fix the problem by always forcing the resource version of the `desired` to match that of the `current` before calling `Update`. This would mean that whatever was in `desired` would overwrite what was `current`. There was rightful [concern](https://github.com/crossplane/crossplane-runtime/pull/176#discussion_r422405453) about doing this in the primary place this was being used (`KubernetesApplication`), but that case is not reflective of the typical desired use of `Update` (which is to overwrite what is existing).

In fact, one of the places where we are currently using the `APIUpdatingApplicator`, propagating the managed connection secret to the claim connection secret, is not functioning as expected. Because the `APIUpdatingApplicator` updates with `current` **those connection secrets actually never get updated**. This was [called out in the PR](https://github.com/crossplane/crossplane-runtime/pull/176#discussion_r422429911), but appears to have been lost among the discussion. 

You _can_ force the overwrite by just setting `current` to `desired` in an `UpdateFn`, but it requires you to know the underlying type of the object you are dealing with. More on this below.

### How can we reproduce it?

I was using the `APIUpdatingApplicator` to create and update objects installed by the new package manager. This is a case where we want an existing object to be completely overwritten by the new. This works fine as long as we know exactly what types we are dealing with. For example:
```go
if err := r.client.Applicator.Apply(ctx, res, resource.MustBeControllableBy(pr.GetUID()), resource.UpdateFn(func(current, desired runtime.Object) {
	c := current.(*apiextensionsv1alpha1.CompositeResourceDefinition)
	d := desired.(*apiextensionsv1alpha1.CompositeResourceDefinition)
	// Make desired owner references match current.
	d.SetOwnerReferences(c.GetOwnerReferences())

	// Applicator will have already checked that this resource is
	// controllable, so it is safe for us to apply a controller
	// reference.
	_ = meta.AddControllerReference(d, meta.AsController(meta.ReferenceTo(pr, pr.GetObjectKind().GroupVersionKind())))

	// Make desired resource version match current.
	d.SetResourceVersion(c.GetResourceVersion())

	// Set current to match desired.
	*c = *d
})); err != nil {
	log.Debug("cannot create XRD for package revision", "error", err)
	pr.SetConditions(runtimev1alpha1.Unavailable(), runtimev1alpha1.ReconcileSuccess())
	return reconcile.Result{RequeueAfter: shortWait}, errors.Wrap(r.client.Status().Update(ctx, pr), errUpdateStatus)
}
```
Will correctly update the object with the new schema because we are manipulating the underlying value of the `current` ptr. However, if we want to deal generally with any object type, for example:
```go
if err := r.client.Applicator.Apply(ctx, res, resource.MustBeControllableBy(pr.GetUID()), resource.UpdateFn(func(current, desired runtime.Object) {
	c := current.(resource.Object)
	d := desired.(resource.Object)

	// Make desired owner references match current.
	d.SetOwnerReferences(c.GetOwnerReferences())

	// Applicator will have already checked that this resource is
	// controllable, so it is safe for us to apply a controller
	// reference.
	_ = meta.AddControllerReference(d, meta.AsController(meta.ReferenceTo(pr, pr.GetObjectKind().GroupVersionKind())))

	// Make desired resource version match current.
	d.SetResourceVersion(c.GetResourceVersion())

	// Set current to match desired.
	c = d
})); err != nil {
	log.Debug("cannot create CRD for package revision", "error", err)
	pr.SetConditions(runtimev1alpha1.Unavailable(), runtimev1alpha1.ReconcileSuccess())
	return reconcile.Result{RequeueAfter: shortWait}, errors.Wrap(r.client.Status().Update(ctx, pr), errUpdateStatus)
}
```
It will not work because we are not changing the underlying value of `current`.

### Proposed Fix

I think it would be a better experience to revert back to updating with the passed in object, and setting the correct resource version automatically. This would mean that the default behavior would be to overwrite the existing with whatever the desired was, and users could manipulate that desired object in the `ApplyOptions`. This is what the updated function would look like:
```go
// not exist, or updated if it does.
func (a *APIUpdatingApplicator) Apply(ctx context.Context, o runtime.Object, ao ...ApplyOption) error {
	m, ok := o.(metav1.Object)
	if !ok {
	}

	if m.GetName() == "" && m.GetGenerateName() != "" {
	}

	current := o.DeepCopyObject()

	err := a.client.Get(ctx, types.NamespacedName{Name: m.GetName(), Namespace: m.GetNamespace()}, current)
	if kerrors.IsNotFound(err) {
	}
	if err != nil {
	}

	for _, fn := range ao {
		if err := fn(ctx, current, o); err != nil {
			return err
		}
	}

	// Always set the resource version of the desired to match existing.
	o.(metav1.Object).SetResourceVersion(current.(metav1.Object).GetResourceVersion())
	return errors.Wrap(a.client.Update(ctx, o), "cannot update object")
}
```

This would support a case like the following where the consumer desires to update an object, but persist the existing owner references:
```go
if err := r.client.Applicator.Apply(ctx, res, resource.MustBeControllableBy(pr.GetUID()), resource.UpdateFn(func(current, desired runtime.Object) {
	c := current.(resource.Object)
	d := desired.(resource.Object)

	// Make desired owner references match current.
	d.SetOwnerReferences(c.GetOwnerReferences())

	// Applicator will have already checked that this resource is
	// controllable, so it is safe for us to apply a controller
	// reference.
	_ = meta.AddControllerReference(d, meta.AsController(meta.ReferenceTo(pr, pr.GetObjectKind().GroupVersionKind())))
})); err != nil {
	log.Debug("cannot create object for package revision", "error", err)
	pr.SetConditions(runtimev1alpha1.Unavailable(), runtimev1alpha1.ReconcileSuccess())
	return reconcile.Result{RequeueAfter: shortWait}, errors.Wrap(r.client.Status().Update(ctx, pr), errUpdateStatus)
}
```

hasheddan:

This modifies the behavior of the APIUpdatingApplicator to update with
the desired object by setting its resource version to that of the
current. This ensures that a successful update will result in the object
being modified to match the desired by default. Consumers of this
Applicator no longer are required to pass in an UpdateFn to avoid no-op
updates.

Fixes #191 

### Checklist

I have:

### DIFF
--- a/pkg/resource/api.go
+++ b/pkg/resource/api.go
@@ -174,8 +174,7 @@ func NewAPIUpdatingApplicator(c client.Client) *APIUpdatingApplicator {
 }
 
 // Apply changes to the supplied object. The object will be created if it does
-// not exist, or updated if it does. If the object does exist and no
-// ApplyOptions are passed, the update will be a no-op.
+// not exist, or updated if it does.
 func (a *APIUpdatingApplicator) Apply(ctx context.Context, o runtime.Object, ao ...ApplyOption) error {
 	m, ok := o.(metav1.Object)
 	if !ok {
@@ -203,7 +202,10 @@ func (a *APIUpdatingApplicator) Apply(ctx context.Context, o runtime.Object, ao 
 		}
 	}
 
-	return errors.Wrap(a.client.Update(ctx, current), "cannot update object")
+	// NOTE(hasheddan): we must set the resource version of the desired object
+	// to that of the current or the update will always fail.
+	o.(metav1.Object).SetResourceVersion(current.(metav1.Object).GetResourceVersion())
+	return errors.Wrap(a.client.Update(ctx, o), "cannot update object")
 }
 
 // An APIFinalizer adds and removes finalizers to and from a resource.
--- a/pkg/resource/api.go
+++ b/pkg/resource/api.go
@@ -176,7 +176,7 @@ func NewAPIUpdatingApplicator(c client.Client) *APIUpdatingApplicator {
 // Apply changes to the supplied object. The object will be created if it does
 // not exist, or updated if it does.
 func (a *APIUpdatingApplicator) Apply(ctx context.Context, o runtime.Object, ao ...ApplyOption) error {
-	m, ok := o.(metav1.Object)
+	m, ok := o.(Object)
 	if !ok {
 		return errors.New("cannot access object metadata")
 	}
@@ -190,22 +190,22 @@ func (a *APIUpdatingApplicator) Apply(ctx context.Context, o runtime.Object, ao 
 	err := a.client.Get(ctx, types.NamespacedName{Name: m.GetName(), Namespace: m.GetNamespace()}, current)
 	if kerrors.IsNotFound(err) {
 		// TODO(negz): Apply ApplyOptions here too?
-		return errors.Wrap(a.client.Create(ctx, o), "cannot create object")
+		return errors.Wrap(a.client.Create(ctx, m), "cannot create object")
 	}
 	if err != nil {
 		return errors.Wrap(err, "cannot get object")
 	}
 
 	for _, fn := range ao {
-		if err := fn(ctx, current, o); err != nil {
+		if err := fn(ctx, current, m); err != nil {
 			return err
 		}
 	}
 
 	// NOTE(hasheddan): we must set the resource version of the desired object
 	// to that of the current or the update will always fail.
-	o.(metav1.Object).SetResourceVersion(current.(metav1.Object).GetResourceVersion())
-	return errors.Wrap(a.client.Update(ctx, o), "cannot update object")
+	m.SetResourceVersion(current.(metav1.Object).GetResourceVersion())
+	return errors.Wrap(a.client.Update(ctx, m), "cannot update object")
 }
 
 // An APIFinalizer adds and removes finalizers to and from a resource.
--- a/pkg/resource/api_test.go
+++ b/pkg/resource/api_test.go
@@ -18,7 +18,6 @@ package resource
 
 import (
 	"context"
-	"fmt"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
@@ -468,13 +467,10 @@ func TestAPIUpdatingApplicator(t *testing.T) {
 			c: &test.MockClient{
 				MockGet: test.NewMockGetFn(nil, func(o runtime.Object) error {
 					*o.(*object) = *current
-					fmt.Println(current.Name)
 					return nil
 				}),
 				MockUpdate: test.NewMockUpdateFn(nil, func(o runtime.Object) error {
-					fmt.Println(current.Name)
-					fmt.Println(o.(*object).Name)
-					if diff := cmp.Diff(*current, *o.(*object)); diff != "" {
+					if diff := cmp.Diff(*desired, *o.(*object)); diff != "" {
 						t.Errorf("r: -want, +got:\n%s", diff)
 					}
 					return nil


pkg/resource/api.go
```diff
@@ -203,7 +202,10 @@ func (a *APIUpdatingApplicator) Apply(ctx context.Context, o runtime.Object, ao
 		}
 	}
 
-	return errors.Wrap(a.client.Update(ctx, current), "cannot update object")
+	// NOTE(hasheddan): we must set the resource version of the desired object
+	// to that of the current or the update will always fail.
+	o.(metav1.Object).SetResourceVersion(current.(metav1.Object).GetResourceVersion())
```
hasheddan, line 207: In practice this type assertion should not be an issue. However, the `runtime.Object` interface is not a superset of `metav1.Object`, so maybe we should check this here?

pkg/resource/api.go
```diff
@@ -174,8 +174,7 @@ func NewAPIUpdatingApplicator(c client.Client) *APIUpdatingApplicator {
 }
 
 // Apply changes to the supplied object. The object will be created if it does
-// not exist, or updated if it does. If the object does exist and no
-// ApplyOptions are passed, the update will be a no-op.
+// not exist, or updated if it does.
 func (a *APIUpdatingApplicator) Apply(ctx context.Context, o runtime.Object, ao ...ApplyOption) error {
 	m, ok := o.(metav1.Object)
```
muvaf, line 179:
```suggestion
	m, ok := o.(Object)
```

`resource.Object` captures both `metav1.Object` and `runtime.Object` and you can use `m` instead of `o` in the rest of the function to not have to do the cast later when you do `SetResourceVersion`

pkg/resource/api.go
muvaf: I think that's fine. I don't really expect a case where an object is `runtime.Object` but not `metav1.Object`.

muvaf (approved):
This still feels weird to me as I can see that if user doesn't take `current` as base in an `ApplyOption` to update their own `desired` template, it will almost certainly fail because of the changes. But not every object is exposed to many changes besides a single controller. So, this could turn out to be fine for some cases.

pkg/resource/api.go
hasheddan: True. I was trying to keep it matching more closely to the patching applicator, but I really don't think that is too much of a concern at this point 👍 Will update

hasheddan: Actually since we are only setting resource version we can use `m` as `metav1.Object`

hasheddan: @muvaf pushed an update, could you indicate your approval still applies?

muvaf: > Actually since we are only setting resource version we can use m as metav1.Object

The underlying object of both `o` and `m` is same but I think it'd be more straight-forward if we used `m` instead of `o` right after the type cast. Setting the resource version on `m` but actually updating `o` could be confusing for example.

hasheddan: Okay, I updated to match.

muvaf (approved):
LGTM twice! :)

hasheddan:
@negz you indicated on #191 that you were fine with this approach, but could you give a drive by approval just to confirm you are aware of this API change?

negz (approved):

PR #194, opened 2020-09-05
Add package parser

hasheddan:

Adds a package parser (and some parser backends) that reads from a YAML stream and serializes objects according the the passed `meta` and `object` schemes.

See initial review at https://github.com/crossplane/crossplane/pull/1701

### Checklist

I have:

### DIFF
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -1,0 +1,259 @@
+/*
+Copyright 2020 The Crossplane Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


pkg/parser/parser.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+// DefaultParser is the default Parser implementation.
+type DefaultParser struct {
+	metaScheme *runtime.Scheme
+	objScheme  *runtime.Scheme
+}
+
+// New returns a new DefaultParser.
+func New(meta, obj *runtime.Scheme) *DefaultParser {
```
negz, line 67: What do you think about taking an interface rather than the concrete *Scheme type? Would it help with testing?

negz (approved):

pkg/parser/parser.go
hasheddan: Sounds good :) I meant to implement that but got sidetracked on how I wanted to do it. I think I'll add an `ObjectCreaterTyper` interface that wraps `runtime.ObjectCreater` and `runtime.ObjectTyper` from `runtime` as you suggested in Slack

PR #195, opened 2020-09-07
Add TypedReferenceTo meta utility

hasheddan:

Adds a TypedReferenceTo utility function that matches the ReferenceTo
implementation for ObjectReference, but returns a TypedReference
instead.

### Checklist

I have:

### DIFF
--- a/pkg/meta/meta.go
+++ b/pkg/meta/meta.go
@@ -27,6 +27,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
+
+	"github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
 )
 
 /*
@@ -58,6 +60,18 @@ func ReferenceTo(o metav1.Object, of schema.GroupVersionKind) *corev1.ObjectRefe
 		APIVersion: v,
 		Kind:       k,
 		Namespace:  o.GetNamespace(),
+		Name:       o.GetName(),
+		UID:        o.GetUID(),
+	}
+}
+
+// TypedReferenceTo returns a typed object reference to the supplied object,
+// presumed to be of the supplied group, version, and kind.
+func TypedReferenceTo(o metav1.Object, of schema.GroupVersionKind) *v1alpha1.TypedReference {
+	v, k := of.ToAPIVersionAndKind()
+	return &v1alpha1.TypedReference{
+		APIVersion: v,
+		Kind:       k,
 		Name:       o.GetName(),
 		UID:        o.GetUID(),
 	}
--- a/pkg/meta/meta.go
+++ b/pkg/meta/meta.go
@@ -81,7 +81,7 @@ func TypedReferenceTo(o metav1.Object, of schema.GroupVersionKind) *v1alpha1.Typ
 }
 
 // AsOwner converts the supplied object reference to an owner reference.
-func AsOwner(r *corev1.ObjectReference) metav1.OwnerReference {
+func AsOwner(r *v1alpha1.TypedReference) metav1.OwnerReference {
 	return metav1.OwnerReference{
 		APIVersion: r.APIVersion,
 		Kind:       r.Kind,
@@ -92,7 +92,7 @@ func AsOwner(r *corev1.ObjectReference) metav1.OwnerReference {
 
 // AsController converts the supplied object reference to a controller
 // reference. You may also consider using metav1.NewControllerRef.
-func AsController(r *corev1.ObjectReference) metav1.OwnerReference {
+func AsController(r *v1alpha1.TypedReference) metav1.OwnerReference {
 	c := true
 	ref := AsOwner(r)
 	ref.Controller = &c
--- a/pkg/meta/meta.go
+++ b/pkg/meta/meta.go
@@ -31,13 +31,6 @@ import (
 	"github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
 )
 
-/*
-	Prefer taking *corev1.ObjectReference as an argument where possible when
-	adding new functions to this package. It's easier to convert an object to
-	an ObjectReference using ReferenceTo() than it is to make an ObjectReference
-	satisfy metav1.Object.
-*/
-
 // AnnotationKeyExternalName is the key in the annotations map of a resource for
 // the name of the resource as it appears on provider's systems.
 const AnnotationKeyExternalName = "crossplane.io/external-name"
--- a/pkg/meta/meta.go
+++ b/pkg/meta/meta.go
@@ -54,6 +54,9 @@ const (
 
 // ReferenceTo returns an object reference to the supplied object, presumed to
 // be of the supplied group, version, and kind.
+// Deprecated: use a more specific reference type, such as TypedReference or
+// Reference instead of the overly verbose ObjectReference.
+// See https://github.com/crossplane/crossplane-runtime/issues/49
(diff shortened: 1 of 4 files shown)


pkg/meta/meta.go
```diff
@@ -63,6 +65,18 @@ func ReferenceTo(o metav1.Object, of schema.GroupVersionKind) *corev1.ObjectRefe
...
 
+// TypedReferenceTo returns a typed object reference to the supplied object,
+// presumed to be of the supplied group, version, and kind.
+func TypedReferenceTo(o metav1.Object, of schema.GroupVersionKind) *v1alpha1.TypedReference {
+	v, k := of.ToAPIVersionAndKind()
+	return &v1alpha1.TypedReference{
+		APIVersion: v,
+		Kind:       k,
+		Name:       o.GetName(),
+		UID:        o.GetUID(),
```
negz, line 76: What do you think about omitting the UID when this helper function is used? Including it would better match the existing `ReferenceTo` implementation, but I think in most cases we don't want or need to include the UID in references. I believe they're typically purely informational (nothing actually verifies that the referenced object has a matching UID), and I suspect we don't have logic that updates the reference UID when an object is recreated (e.g. during a backup and restore).

pkg/meta/meta.go
```diff
@@ -63,6 +65,18 @@ func ReferenceTo(o metav1.Object, of schema.GroupVersionKind) *corev1.ObjectRefe
 	}
 }
 
+// TypedReferenceTo returns a typed object reference to the supplied object,
+// presumed to be of the supplied group, version, and kind.
+func TypedReferenceTo(o metav1.Object, of schema.GroupVersionKind) *v1alpha1.TypedReference {
```
negz, line 70: Should we mark the existing `ReferenceTo` implementation deprecated, in service of https://github.com/crossplane/crossplane-runtime/issues/49?

negz (approved):
Two comments, but I'm happy to defer them both to some future point if they're cumbersome to do at the moment.

pkg/meta/meta.go
hasheddan: Good idea! Will update :+1:

pkg/meta/meta.go
hasheddan: So in the case I am using this, we are actually updating the status with a reference to the most recently gotten object, so while the UID could be stale for a while, it will eventually get updated. I would prefer to keep it if all the same to you.

negz: I'm not super convinced there's value in including the UID, but given we more often do so today I'm happy to leave that for a future change.

pkg/meta/meta.go
```diff
@@ -63,6 +68,18 @@ func ReferenceTo(o metav1.Object, of schema.GroupVersionKind) *corev1.ObjectRefe
...
+	v, k := of.ToAPIVersionAndKind()
+	return &v1alpha1.TypedReference{
+		APIVersion: v,
+		Kind:       k,
+		Name:       o.GetName(),
```
muvaf, line 78:
```suggestion
		Name:       o.GetName(),
		Namespace: o.GetNamespace(),
```

My immediate expectation from a `TypedReferenceTo` function is for it to have a full reference but it doesn't include namespace. So, I think namespace should be included unless there is an explicit sign(in its name or how argument list is constructed) that says this function is specifically for cluster-scoped resources. Also, it cannot be a full replacement of `ReferenceTo` if it doesn't include namespace.

hasheddan: It sounds like you have issue with the `TypedReference` type rather than this function. `TypedReference` does not have a namespace so one cannot be returned here. In regards to replacing `ObjectReference`, I don't think we are saying `TypedReference` should be used in all cases instead of `ObjectReference`. We are saying that a more scoped reference should be used instead on `ObjectReference`, so if one does not exist that fits your use case then you should add it.

pkg/meta/meta.go
hasheddan: I think that's fair, but I think it should be removed from `TypedReference` as opposed to just not returning it here.

hasheddan: Would you like for me to go ahead and remove it from both?

pkg/meta/meta.go
muvaf: I'm actually fine with `TypedReference` being cluster-scoped but I don't think it replaces `ObjectReference` in this specific context. For example, if I used `ReferenceTo` for connection secrets then this deprecation means that I have to implement my own function now instead of using the suggested `TypedReferenceTo` function. So, I'd be fine with this function if it didn't aim to deprecate & replace `ReferenceTo`.

hasheddan: @muvaf the deprecation notice does not indicate that you should always use `TypedReference` or `TypedReferenceTo`. It gives it as an example, but says to just use a generally more scoped reference type. We _are_ saying that if you want a different reference type then you should add it and a helper function for it. We are trying to moved away from using `ObjectReference` per the issue linked by @negz

muvaf: Alright, I think we need the namespaced alternative but that doesn't seem to be in the scope of this PR. I guess we can open another one that adds the necessary type and helper function when needed by core Crossplane.

hasheddan: I am happy to add it in this PR if we have a use case for it.

muvaf: I think all logics related to connection secret creation rely on that function for getting an owner ref of claim, like [here](https://github.com/crossplane/crossplane-runtime/blob/master/pkg/resource/resource.go#L119). Also, I use it in crossplane agent, too, for setting owner ref of the propagated secret to the claim.

hasheddan: Hmm for the sake of ownerReferences, it may make sense to no deprecate this, but rather just say "please use a more scoped reference if possible. What do you think? /cc @negz

muvaf: Well, I just realized that `OwnerReference` struct doesn't have `Namespace` field anyway since it's either namespace<>namespace or cluster-scoped<>cluster-scoped. So, replacing `AsOwner` and `AsController` functions with the ones that accept `TypedReference` could be sufficient to cover that use case. As you pointed out, if another thing comes up that requires namespace, we can consider an addition.

hasheddan: @muvaf nice find! This goes a long way towards addressing https://github.com/crossplane/crossplane-runtime/issues/49 as well. Update coming!

muvaf (approved):
LGTM! I believe `AsOwner` and `AsController` changes are breaking, I'm leaving it to you whether you want alternative impl. and deprecation of the current ones.

hasheddan:
@muvaf thanks! Though we would typically want to deprecate and then remove in a few releases, the blast radius is pretty small here, and functionally the uses will remain the same after changing the function signature. I just updated `crossplane/crossplane` and it was very straightforward. I'll make sure to include breaking change in release notes 👍

pkg/meta/meta.go
```diff
@@ -27,6 +27,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
+
+	"github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
 )
 
 /*
```
negz, line 34: We should either update this comment or remove it.

pkg/meta/meta.go
negz: As you mentioned out of band, this function is now being used for owner references, which require UIDs. I'm happy to leave the UIDs in. I suspect doing so will continue a pattern of including UIDs in some contexts where they're not particularly useful, but that's not a regression compared to the existing `ReferenceTo`.

PR #199, opened 2020-09-10
fieldpath: add GetInteger to retrieve int64 number

muvaf:

In practice, it's `int64` most of the time as I have discovered during https://github.com/crossplane/crossplane/pull/1713 . Details are left as comment in the code.

### Checklist

I have:

### DIFF
--- a/pkg/fieldpath/paved.go
+++ b/pkg/fieldpath/paved.go
@@ -17,10 +17,9 @@ limitations under the License.
 package fieldpath
 
 import (
-	"encoding/json"
-
 	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/json"
 )
 
 type errNotFound struct {
--- a/pkg/fieldpath/paved.go
+++ b/pkg/fieldpath/paved.go
@@ -220,7 +220,15 @@ func (p *Paved) GetBool(path string) (bool, error) {
 	return b, nil
 }
 
+// NOTE(muvaf): If there is no CRD, unstructured.Unstructured reads numbers as
+// float64. However, in practice, use of float64 is discouraged and when you fetch
+// an instance of a CRD whose number fields are int64, you'll get int64. So,
+// it's not really possible to test this without an api-server but that's the
+// actual behavior.
+
 // GetNumber value of the supplied field path.
+// Deprecated: Use of float64 is discouraged. Please use GetInteger.
+// See https://github.com/kubernetes/community/blob/c9ae475/contributors/devel/sig-architecture/api-conventions.md#primitive-types
 func (p *Paved) GetNumber(path string) (float64, error) {
 	v, err := p.GetValue(path)
 	if err != nil {
@@ -230,6 +238,20 @@ func (p *Paved) GetNumber(path string) (float64, error) {
 	f, ok := v.(float64)
 	if !ok {
 		return 0, errors.Errorf("%s: not a (float64) number", path)
+	}
+	return f, nil
+}
+
+// GetInteger value of the supplied field path.
+func (p *Paved) GetInteger(path string) (int64, error) {
+	v, err := p.GetValue(path)
+	if err != nil {
+		return 0, err
+	}
+
+	f, ok := v.(int64)
+	if !ok {
+		return 0, errors.Errorf("%s: not a (int64) number", path)
 	}
 	return f, nil
 }
--- a/pkg/fieldpath/paved_test.go
+++ b/pkg/fieldpath/paved_test.go
@@ -17,13 +17,13 @@ limitations under the License.
 package fieldpath
 
 import (
-	"encoding/json"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/pkg/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/json"
 
 	"github.com/crossplane/crossplane-runtime/pkg/test"
 )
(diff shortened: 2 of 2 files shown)


pkg/fieldpath/paved.go
```diff
@@ -220,7 +220,15 @@ func (p *Paved) GetBool(path string) (bool, error) {
 	return b, nil
 }
 
+// NOTE(muvaf): If there is no CRD, unstructured.Unstructured reads numbers as
+// float64. However, in practice, use of float64 is discouraged and when you fetch
```
hasheddan, line 224: I had a little trouble parsing this. Are you saying that if you paved a `struct` with `int64` fields then unstructured is going to treat them as `float64`, but if you get an object from the api-server into `unstructured.Unstructured` then it is going to treat them as `int64`? I am guessing this is due to an issue with `runtime.DefaultUnstructuredConverter` (which we use above https://github.com/crossplane/crossplane-runtime/pull/199/files#diff-1c5d867f3ffb167cee0af55ba02e05e4R53) I worked on with some of the Velero folks, which has now been fixed upstream, but will not be included until `1.20` release.

Issue: https://github.com/kubernetes/kubernetes/issues/87675
PR: https://github.com/kubernetes/kubernetes/pull/93250

Anyway, not sure this directly impacts your change here, but might be worth leaving a note. We could avoid this by doing something like https://github.com/crossplane/crossplane/blob/b6c6ae4c6b64d29866e70d8e20f5dee3e07f97e2/pkg/parser/conversion.go#L30

hasheddan (commented):
I have not used the `fieldpath` library as extensively as you and @negz, so I will leave approval to y'all 👍 I did add a comment as to why you may be seeing the behavior that you mentioned.

pkg/fieldpath/paved.go
muvaf: For example, if we supply `{"metadata":{"version":2}}`, it [will show that](https://github.com/crossplane/crossplane-runtime/blob/bd09d0c/pkg/fieldpath/paved_test.go#L491) the Go type of `metadata.version` is `float64`. However, if you run this code against an api-server, it will treat a numerical field as `int64`. I made a connection there about CRD in the api-server, i.e. api-server might be sending additional type information since it knows but it could be due to something else, too. In the end, I think it's probably not possible to retrieve `float64` from api-server when using `Unstructured`. @hasheddan is this what's fixed there?

negz:
Exposing only `GetNumber` here was actually an intentional decision - I should have left a comment explaining this.

When the plain old Go JSON decoder decodes a JSON number into an an `interface{}{}` it has no type data to determine whether the number should be an int or float, so it always decodes to `float64`. My thinking was it was safest to return the data as an 'Number' that the caller could then convert to an int64 if so desired. This way there was no implication that we were validating the number was actually supposed to be an integer.

It sounds like in practice we're not seeing the behaviour of the Go JSON unmarshaller, and in fact we're getting a `map[string]interface{}` that contains int64 numbers?

pkg/fieldpath/paved.go
```diff
@@ -220,7 +220,15 @@ func (p *Paved) GetBool(path string) (bool, error) {
...
+// an instance of a CRD whose number fields are int64, you'll get int64. So,
+// it's not really possible to test this without an api-server but that's the
+// actual behavior.
+
 // GetNumber value of the supplied field path.
+// Deprecated: Use of float64 is discouraged. Please use GetInteger.
```
negz, line 230: While Kubernetes API conventions discourage float64, I'm not sure we can get away with deprecating this because I think the value will actually contain float64 sometimes. When we call `SetValue` [we round trip the value through JSON encoding](https://github.com/crossplane/crossplane-runtime/blob/9b321c/pkg/fieldpath/paved.go#L238) which will make it a `float64`. The same is true when the DefaultUnstructuredConvert is running with mismatchDetection enabled.

muvaf:
> As far as I can tell the only way we'd end up with a field value or array index containing an int64 would be if we were Paving a runtime.Object that was actually a pointer to a struct, in which case the DefaultUnstructuredConverter would use reflection rather than JSON to build the map[string]interface{}.

@negz It's definitely `unstructured.Unstructured`, I tried using it [here](https://github.com/crossplane/crossplane/pull/1713/files#diff-f1e1f1a3bf693ac49ddf11f905d9ad1bR209) for composite resources where it checks an equality.

> It sounds like in practice we're not seeing the behaviour of the Go JSON unmarshaller, and in fact we're getting a map[string]interface{} that contains int64 numbers? Do you know how this happens?

~~I don't really know for sure.~~ The unstructured client uses a custom json marshaller and that [tries to cast any number into `int64`](https://github.com/kubernetes/apimachinery/blob/master/pkg/util/json/json.go#L49) and if that fails, only then tries to cast [it into `float64`](https://github.com/kubernetes/apimachinery/blob/master/pkg/util/json/json.go#L150). All this happens during the `Get` call. That's why our unit tests don't really see that behavior because we use Go json unmarshaller.

I believe we can keep a function that returns `float64` but maybe not name it as `GetNumber`. Something very specific like `GetFloat` or even delete and let people use `GetValue` since it seems to be kind of an edge case.

pkg/fieldpath/paved.go
muvaf: It seems like the number stuff is the only reason apimachinery uses a custom json unmarshaller. Maybe we should switch to that? Also, do you think that round trip is really needed? I couldn't immediately see a case where the given map is not a valid one and user doesn't get an error if we don't do the round-trip.

muvaf:
I have pushed a new commit that makes unit tests use the json library of apimachinery and it validates the behavior I described above. Now both `GetNumber` and `GetInteger` have the necessary tests.

pkg/fieldpath/paved.go
negz: Yeah, let's definitely switch this to the custom JSON unmarshaller that apimachinery is using.

We need this round trip for at least the following situation:

```go
type Cool struct {
  Coolness int
}

// We need &Cool{} to be converted to a map[string]interface{}
// or else GetValue won't be able to return its fields.
p.SetValue("very.cool.field", &Cool{})
```

negz: Given that `float64` is discouraged and the JSON marshaller tries hard to use `int64` instead I feel okay deprecating this function. It seems like it's still _possible_ to use a `float64` (against conventions), but we could use `GetValue` if we _really_ wanted a `float64`.

negz: Heads up, I'm happy to approve and merge this once the JSON unmarshaller we're using in `setValue` is changed to use the Kubernetes one.

negz (approved):

PR #200, opened 2020-09-18
Indicate whether Apply errors are due to a resource being uncontrollable

negz:
This allows users of the MustBeControllableBy ApplyOptions to determine how to handle the case in which a resource cannot be applied because it is not controllable - e.g. to abort the apply operation without returning an error.

### Checklist

I have:


hasheddan (approved):
LGTM

PR #201, opened 2020-09-21
Remove support for resource claims and classes

negz:

Fixes https://github.com/crossplane/crossplane-runtime/issues/179
Updates https://github.com/crossplane/crossplane/issues/1670

See https://github.com/crossplane/crossplane/issues/1670 for context.

### Checklist

I have:


hasheddan (approved):
This LGTM! I wonder if we should also consider removing target-related things here and KA in crossplane while we are at it. Also just a note for anyone else following along that this will require large updates in each provider for them to be compatible with Crossplane v0.13

pkg/resource/fake/mocks.go
```diff
@@ -225,67 +189,14 @@ func (o *Object) DeepCopyObject() runtime.Object {
...
 // Managed is a mock that implements Managed interface.
 type Managed struct {
 	metav1.ObjectMeta
-	ClassReferencer
-	ClaimReferencer
 	ProviderReferencer
 	ProviderConfigReferencer
 	ConnectionSecretWriterTo
 	Orphanable
-	Reclaimer
```
muvaf, line 286: Others are meaningful only if the managed resource was used with a claim, which is being removed. But `Reclaimer` actually corresponds to a behavior in managed resource level. So, that would be a breaking change on `beta` level. In the release notes, we should make sure that this specific change is emphasized.

muvaf (approved):
LGTM! Let's make sure we document this change well during the release.

pkg/resource/fake/mocks.go
negz: Yeah, I added the `very-breaking-change` to help remind us to pay close attention to this PR when writing release notes.

negz:
I'm feeling pretty comfortable with this change having tested it with a branch of provider-gcp.

PR #202, opened 2020-09-21
Default ProviderConfig is automatically selected

muvaf:

If `spec.providerConfigRef` is empty, it will be filled with `default`.

Note that this PR adds that new initializer to the default initializer list but there are managed resources who supply their own initializers. So, we'll need to update them manually.

Fixes https://github.com/crossplane/crossplane/issues/1686

### Checklist

I have:

### DIFF
--- a/pkg/reconciler/managed/api.go
+++ b/pkg/reconciler/managed/api.go
@@ -24,6 +24,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
 	"github.com/crossplane/crossplane-runtime/pkg/meta"
 	"github.com/crossplane/crossplane-runtime/pkg/resource"
 )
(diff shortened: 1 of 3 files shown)


hasheddan (commented):
> Note that this PR adds that new initializer to the default initializer list but there are managed resources who supply their own initializers. So, we'll need to update them manually.

Would it make since for this just to be in the managed reconciler body? Do you think there will be cases where we don't want this behavior?

Alternatively, we could have some sort of defaulter interface to separate this from `NewNameAsExternalName()` which feels like something which may be overridden more frequently. Not strictly opposed to what you have here, but just wanted to float a few more ideas :)

muvaf:
> Alternatively, we could have some sort of defaulter interface to separate this from NewNameAsExternalName() which feels like something which may be overridden more frequently.

Actually, the reason we had introduced the `Initializer` interface was defaulting. I agree that most probably, it will not be overridden in conventional infra providers. However, it's still a defaulting mechanism so it fits into why we have `Initializer` in the first place. So, the ability to override is kind of a positive side-effect in this case for users who develop unconventional or internal providers (maybe wish to use a mechanism to create a `ProviderConfig` that uses ec2 metadata by default rather than assuming one exists) but not my main goal here.

> Not strictly opposed to what you have here, but just wanted to float a few more ideas :)

Keep them coming, it's always helpful for me to reconsider my thought process :)

negz:
> Would it make since for this just to be in the managed reconciler body?

I'd like us to have a pretty high bar for adding more logic to the managed reconciler body given how complex it is already.

negz (approved):
This approach didn't occur to me, but it seems like a good one. I presume we'll also need to update any controllers we still have lying around that don't use the managed resource reconciler?

muvaf:
> I presume we'll also need to update any controllers we still have lying around that don't use the managed resource reconciler?

Yes, that looks like the case :/

PR #203, opened 2020-09-23
managed reconciler: external.Connect should be called after initializers and ref resolvers

muvaf:

Calling `external.Connect` before initializers and reference resolvers are unnecessary since they can return error and we'd not be able to use the returned client anyway. Particularly, it needs to be called after the initializers for https://github.com/crossplane/crossplane-runtime/pull/202 to work and it needs to be called after ref resolvers for https://github.com/crossplane/crossplane/issues/1699 to be implementable.

Tested manually.

### Checklist

I have:


hasheddan (approved):
Seems reasonable to me, thanks @muvaf!

PR #204, opened 2020-09-24
Block deletion when a resource's secret is in use

negz:

Managed resources write connection secrets. These secrets may be consumed by other resource - for example a Kubernetes cluster's connection secret may be used as a Helm ProviderConfig's credentials secret. This commit blocks a managed resource from being deleted if it has written a connection secret and that connection secret is owned by (i.e. has an owner reference to) another resource.

See https://github.com/crossplane/crossplane/issues/1737#issuecomment-698147088 for context. I'm opening this as a draft for the moment because I haven't had a chance to test it.

### Checklist

I have:

### DIFF
--- a/pkg/reconciler/managed/api.go
+++ b/pkg/reconciler/managed/api.go
@@ -22,6 +22,7 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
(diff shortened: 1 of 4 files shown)


hasheddan (commented):
@negz this looks pretty good to me! One thing I did consider while thinking about this scenario is whether a finalizer would be more appropriate. I don't think it necessarily has any advantages from a technical standpoint, and similar to the usage of owner references here we would still be implementing a layer on top (i.e. for owner refs we are explicitly checking vs just removing managed resource owner ref and letting the provider config owner ref block k8s GC, and for finalizer we would be explicitly checking vs deleting and having the existing finalizer block deletion). However, from a UX perspective it may be less surprising to folks if they see that the managed resource fails to be deleted and then see that a finalizer is present on its secret (even though the finalizer typically would not prevent a common parent from being deleted).

Anyway, not necessarily a superior route from a technical perspective, but just wanted to float it by you in case you had not considered it already 👍

negz:
Interestingly owner references were what I found most intuitive here, but I can see how finalizers would make more sense. I think in general this idea needs a little more work - I'm going to close this for now as we may take a different approach per https://github.com/crossplane/crossplane/issues/1782.

PR #206, opened 2020-09-26
Add machinery for tracking ProviderConfig usage

negz:

This commit adds a reconciler and some utiltiies that allow Crossplane providers to track the usage of their ProviderConfig resources. To do so, a provider must:

* Use the tracker to Apply a `ProviderConfigUsage` each time a managed resource is about to try to use a `ProviderConfig`.
* Start a controller that watches for `ProviderConfig` and `ProviderConfigUsage` using the supplied reconciler.

A `ProviderConfigUsage` can be created or updated using a `*resource.ProviderConfigUsageTracker`:

```go
u := resource.NewProviderConfigUsageTracker(kube, &v1alpha1.ProviderConfigUsage{})
err := u.Track(ctx, mg)
if resource.IsMissingReference(err) {
	// The supplied managed resource doesn't reference a ProviderConfig. Try something else.
	return ConnectUsingDeprecatedProvider(ctx, mg)
}
```

The controller is added as follows:

```go
// Setup adds a controller that reconciles ProviderConfigs by accounting for
// their current usage.
func Setup(mgr ctrl.Manager, l logging.Logger) error {
	name := providerconfig.ControllerName(v1alpha1.ProviderConfigKind)

	of := resource.ProviderConfigKinds{
		Config:    v1alpha1.ProviderConfigGroupVersionKind,
		UsageList: v1alpha1.ProviderConfigUsageListGroupVersionKind,
	}

	return ctrl.NewControllerManagedBy(mgr).
		Named(name).
		For(&v1alpha1.ProviderConfig{}).
		Watches(&source.Kind{Type: &v1alpha1.ProviderConfigUsage{}}, &resource.EnqueueRequestForProviderConfig{}).
		Complete(providerconfig.NewReconciler(mgr, of,
			providerconfig.WithLogger(l.WithValues("controller", name)),
			providerconfig.WithRecorder(event.NewAPIRecorder(mgr.GetEventRecorderFor(name)))))
}
```

See https://github.com/crossplane-contrib/provider-helm/pull/31 for details on how I've tested this.

### Checklist

I have:


negz:
I believe this is now ready for review. See https://github.com/crossplane-contrib/provider-helm/pull/31 for details on how I've tested it.

pkg/reconciler/providerconfig/reconciler_test.go
```diff
@@ -0,0 +1,331 @@
...
+					Config:    fake.GVK(&fake.ProviderConfig{}),
+					UsageList: fake.GVK(&ProviderConfigUsageList{}),
+				},
+			},
+			want: want{
+				result: reconcile.Result{RequeueAfter: shortWait},
+			},
+		},
+		"BlockDeleteWhileInUse": {
+			reason: "We should return without requeueing if the provider config is still in used",
```
hasheddan, line 167: nitpick:
```suggestion
			reason: "We should return without requeueing if the provider config is still in use",
```

apis/core/v1alpha1/resource.go
```diff
@@ -167,15 +171,31 @@ type ProviderSpec struct {
...
 	// CredentialsSecretRef references a specific secret's key that contains
 	// the credentials that are used to connect to the provider.
 	// +optional
 	CredentialsSecretRef *SecretKeySelector `json:"credentialsSecretRef,omitempty"`
 }
 
+// A ProviderConfigStatus defines the observed status of a ProviderConfig.
+type ProviderConfigStatus struct {
+	// Users of this provider configuration.
+	Users *int64 `json:"users,omitempty"`
```
hasheddan, line 186: Does this need to be a pointer type? I think we have typically not used pointers in status. It seems like having the zero-value of `int64` (0) when not set would be fine here

pkg/resource/enqueue_handlers.go
```diff
@@ -71,3 +72,41 @@ func addPropagated(obj runtime.Object, queue adder) {
...
+}
+
+// Generic adds a NamespacedName for the supplied GenericEvent if its Object is
+// a ProviderConfigReferencer.
+func (e *EnqueueRequestForProviderConfig) Generic(evt event.GenericEvent, q workqueue.RateLimitingInterface) {
+	addProviderConfig(evt.Object, q)
+}
+
+func addProviderConfig(obj runtime.Object, queue adder) {
+	pcr, ok := obj.(RequiredProviderConfigReferencer)
```
muvaf, line 106: I guess you're trying to make sure a reconcile request is accepted for `ProviderConfig` via events on `ProviderConfigUsage` that has reference to that `ProviderConfig`. However, there isn't anything that filters out other `ProviderConfigUsage`s which reference a different `ProviderConfig`. It seems this only does a check for an interface.

hasheddan (commented):
This looks pretty good to me, although I am a little hesitant about the idea of requiring provider authors to include logic about creating a `ProviderConfigUsage`. I can't think of a similar scenario in which we produce side-effect objects in the cluster as a provider. However, maybe we take the approach that providers are only written by "advanced" users and they should know how to perform these types of operations. Would it be possible to plumb this into the managed reconciler? All managed resources implement `ProviderConfigReferencer`, so it seems like passing in a function that produces their flavor of `ProviderConfigUsage` would make moving the machinery to create one for every managed resource that uses a `ProviderConfig` to the managed reconcile loop possible.

pkg/resource/interfaces.go
```diff
@@ -114,6 +121,12 @@ type CompositeResourceReferencer interface {
 	GetResourceReference() *corev1.ObjectReference
 }
 
+// A UserCounter can count how many users it has.
+type UserCounter interface {
+	SetUsers(i *int64)
+	GetUsers() *int64
```
muvaf, line 127:
```suggestion
	SetUsers(i int64)
	GetUsers() int64
```

I think `0` as default value makes sense for this information. How does the following sound for users field:
```go
// A ProviderConfigStatus defines the observed status of a ProviderConfig.
type ProviderConfigStatus struct {
	// Users of this provider configuration.
	Users int64 `json:"users,omitempty"`
}
```

pkg/reconciler/providerconfig/reconciler.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	users := int64(len(l.GetItems()))
+	for _, pcu := range l.GetItems() {
+		if metav1.GetControllerOf(pcu) == nil {
+			// Usages should always have a controller reference. If this one has
+			// none it's probably been stripped off (e.g. by a Velero restore).
+			// We can safely delete it - it's either stale, or will be recreated
+			// next time the relevant managed resource connects.
+			if err := r.client.Delete(ctx, pcu); resource.IgnoreNotFound(err) != nil {
```
muvaf, line 159: What happens if `spec.providerConfigRef` is changed to another value? Checking with https://github.com/crossplane/provider-gcp/pull/262 , it seems this would block deletion of the old `ProviderConfig` without any reason. Should we make sure the object pointed with resource reference in PCU is still using that PC? If not, delete the PCU?

apis/core/v1alpha1/resource.go
```diff
@@ -44,6 +44,10 @@ const (
 	ResourceCredentialsSecretKubeconfigKey = "kubeconfig"
 )
 
+// LabelKeyProviderName is added to ProviderConfigUsages to relate them to their
+// ProviderConfig.
+const LabelKeyProviderName = "crossplane.io/provider-config"
```
muvaf, line 49: It feels weird to label the resource even if that information exists under spec but I understand that we do this to delegate the filterin to api-server instead of doing it via a predicate which makes the client do the filtering.

pkg/reconciler/providerconfig/reconciler.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+	log = log.WithValues("usages", users)
+
+	if meta.WasDeleted(pc) {
+		if users > 0 {
+			log.Debug("Blocking deletion while usages still exist")
+			r.record.Event(pc, event.Warning(reasonAccount, errors.New("Blocking deletion while usages still exist")))
+
+			// We're watching our usages, so we'll be requeued when they go.
+			pc.SetUsers(&users)
```
muvaf, line 175: Should we set a condition, too? Something like `Terminating`.

muvaf:
> Would it be possible to plumb this into the managed reconciler? All managed resources implement ProviderConfigReferencer, so it seems like passing in a function that produces their flavor of ProviderConfigUsage would make moving the machinery to create one for every managed resource that uses a ProviderConfig to the managed reconcile loop possible.

I agree with @hasheddan here. It's a machinery with known bounds by us, we only need the actual structs to be given. In fact, I looked for such a logic in the managed reconciler first before realizing it's the provider that does this operation.

negz:
> I am a little hesitant about the idea of requiring provider authors to include logic about creating a ProviderConfigUsage. I can't think of a similar scenario in which we produce side-effect objects in the cluster as a provider. However, maybe we take the approach that providers are only written by "advanced" users and they should know how to perform these types of operations.

This hesitance is a little surprising to me given that we already expect provider authors to implement their `ProviderConfig` type and logic to acquire it and credentials from it. Establishing a convention that they should also apply a `ProviderConfigUsage` doesn't seem too different to me.

> Would it be possible to plumb this into the managed reconciler?

Would you also expect the managed reconciler to get the relevant `ProviderConfig`?

pkg/resource/enqueue_handlers.go
negz: I don't follow. We watch a particular GVK of `ProviderConfigUsage` - e.g. GCP PCUs - and enqueue a request for the `ProviderConfig` they reference below. 

https://github.com/crossplane/provider-gcp/pull/262/files#diff-37d5b8a93884a62afcabf3fc6dceeea0R44

pkg/resource/interfaces.go
negz: Makes sense. I forget why I switched it to a pointer.

pkg/reconciler/providerconfig/reconciler.go
negz: The PCU is updated on every managed resource reconcile loop if it has changed from what is in cache. So if the managed resource's `spec.providerConfigRef` changes:

* The PCU's `providerConfigRef` also changes to reflect the new name.
* The PCU's `crossplane.io/provider-config` label changes to reflect the new name.
* The PCU would no longer be included in the `List` call above for the old PC.

I think the only risk of a PCU getting stuck pointing to an old PC is if:

* The managed resource's `spec.providerConfigRef` is updated.
* The managed resource fails to get the new PC, or its secret, and thus never writes its new PCU.

I think a reasonable way to avoid this that would not involve having to get and validate the managed resource would be to simply write the PCU _before_ attempting to get and use the PC. i.e. The PCU declares the intent to use a PC, rather than a successful use of the PC.

That all said, I'm skeptical that changing the `spec.providerConfigRef` of a managed resource is something we should allow. It seems like it could easily result in a managed resource becoming orphaned or broken. I'm not immediately sure what reason someone would have to change what PC a managed resource was using.

apis/core/v1alpha1/resource.go
negz: What makes it weird?

This is done mostly for UX rather than anything else (though it does help with listing). i.e. `kubectl get providerconfigusage -l crossplane.io/provider-config=foo`

pkg/reconciler/providerconfig/reconciler.go
negz: We could, though we'd presumably need some condition that implied something like "not terminating, but not really confirmed to actually be ready/working/etc". Maybe `Active`?

hasheddan:
> This hesitance is a little surprising to me given that we already expect provider authors to implement their ProviderConfig type and logic to acquire it and credentials from it. Establishing a convention that they should also apply a ProviderConfigUsage doesn't seem too different to me.

@negz this is definitely a fair outlook, but I would argue that needing to get the `ProviderConfig` as part of the `Connect()` feels a little more in line with steps to talk to the cloud provider API (i.e. informing how to connect), while the PCU step feels like Crossplane machinery that a provider author shouldn't necessarily have to be knowledgeable of. Where possible, it seems like minimizing the implementation onus on provider authors is preferable. That being said, implementing it in each provider is not wrong, was there a specific reason why you wanted to go that route?

> Would you also expect the managed reconciler to get the relevant ProviderConfig?

We could get it and just provide it to the `Connect()` method, that actually sounds pretty good to me.

negz:
> We could get it and just provide it to the Connect() method, that actually sounds pretty good to me.

Yeah, I'm experimenting with something like this. I think it should be either/or; it would be strange to me if we automatically created the PCU but expected the implementor to deal with the PC. I think we should either handle both, or neither.

negz:
> I'm experimenting with something like this.

This idea is interesting in theory but I'm not having much luck implementing it in a fashion that accounts for all the variables involved in connecting to a provider (`ProviderConfig` vs the old `Provider`, secrets vs IAM, different clients needed by different resources, needing type specific managed resource fields such as a region). It could be that a cleaner way to handle this will arise in future, but my initial suspicion is that it would make building a provider more obtuse than the current approach, in which provider authors must (or may) create a PCU manually. I don't feel that this dramatically expands on the contract they already need to fulfil - i.e. get a ProviderConfig, fall back to a Provider, potentially handle IAM auth, get credentials from a secret, etc.

negz:
I've added a WIP commit that shows how we could move some of the logic of fetching a `ProviderConfig` and creating a `ProviderConfigUsage` into the `managed.Reconciler` without breaking the existing `ExternalConnector` API. I'm on the fence as to whether this approach is an improvement compared to provider authors handling it themselves. That said, a lot of the complexity comes from the fact that we need to handle falling back to a `Provider` at the moment while that type is deprecated but not removed.

I also considered using an initialiser here, but I believe fetching and registering the use of a provider fits better in the connection logic.

pkg/resource/enqueue_handlers.go
muvaf: Hmm, I think you can achieve that by using `Owns(&v1beta1.ProviderConfigUsage{})`.

apis/core/v1alpha1/resource.go
muvaf: It's a duplication of the information that already exists in the same object.

pkg/reconciler/providerconfig/reconciler.go
muvaf: > I'm skeptical that changing the spec.providerConfigRef of a managed resource is something we should allow. It seems like it could easily result in a managed resource becoming orphaned or broken. I'm not immediately sure what reason someone would have to change what PC a managed resource was using.

I can't see big use cases, too. But I wouldn't be able explain why a user wouldn't be able to change if they wanted to, either. Normally, the resource is there and cloud providers allow you to manipulate it with any creds you want as long as they got the necessary permissions. Diverging from that doesn't really feel right.

> to simply write the PCU before attempting to get and use the PC. i.e. The PCU declares the intent to use a PC, rather than a successful use of the PC.

This sounds good. Though I wouldn't frame PCU as the intent to use. In the end, we don't reject the change on `spec.providerConfigRef` and that's the ultimate source of truth, i.e. we don't use PCU to decide what `ProviderConfig` to use during `Connect`. It's mostly machinery and not being able to reconcile the managed resource if PCU couldn't be updated looks like a negative side effect of that machinery, though I don't expect that to happen frequently anyway.

pkg/reconciler/providerconfig/reconciler.go
muvaf: `Active`, `Used` or `Connected` all sound good to me.

muvaf (commented):
> That said, a lot of the complexity comes from the fact that we need to handle falling back to a Provider at the moment while that type is deprecated but not removed.

@negz do we really need to care about `Provider` type? It's still used but deprecated so I wouldn't expect users to ask for PCU feature for `Provider` type. In the adapter, we'd return early if `spec.providerConfigRef` is empty.

I guess it'd look like the following, which is pretty close to what we have today:
```go
// SetupCertificate adds a controller that reconciles Certificates.
func SetupCertificate(mgr ctrl.Manager, l logging.Logger) error {
	name := managed.ControllerName(v1alpha1.CertificateGroupKind)

	pca := managed.NewProviderConfigAdapter(mgr,
		resource.ProviderConfigKinds{
			Config: v1beta1.ProviderConfigGroupVersionKind,
			Usage:  v1beta1.ProviderConfigUsageGroupVersionKind,
		},
		&connecter{client: mgr.GetClient(), newClientFn: acm.NewClient},
	)

	return ctrl.NewControllerManagedBy(mgr).
		Named(name).
		For(&v1alpha1.Certificate{}).
		Complete(managed.NewReconciler(mgr,
			resource.ManagedKind(v1alpha1.CertificateGroupVersionKind),
			managed.WithExternalConnecter(pca),
			managed.WithConnectionPublishers(),
			managed.WithReferenceResolver(managed.NewAPISimpleReferenceResolver(mgr.GetClient())),
			managed.WithInitializers(managed.NewDefaultProviderConfig(mgr.GetClient())),
			managed.WithLogger(l.WithValues("controller", name)),
			managed.WithRecorder(event.NewAPIRecorder(mgr.GetEventRecorderFor(name)))))
}

// The rest stays same as today.

type connector struct {
	client      client.Client
	newClientFn func(aws.Config) acm.Client
}

func (c *connector) Connect(ctx context.Context, mg resource.Managed) (managed.ExternalClient, error) {
	cr, ok := mg.(*v1alpha1.Certificate)
	if !ok {
		return nil, errors.New(errUnexpectedObject)
	}
	cfg, err := awscommon.GetConfig(ctx, c.client, mg, cr.Spec.ForProvider.Region)
	if err != nil {
		return nil, err
	}
	return &external{c.newClientFn(*cfg), c.client}, nil
}
```

> I'm on the fence as to whether this approach is an improvement compared to provider authors handling it themselves.

Yeah, I can see that. Though if we don't have to use PCU for `Provider`, I think the only thing they will need to handle is the adapter initializing shown in your example, i.e. specific authentication functions wouldn't need to change as opposed to former approach, which is an improvement IMO.

pkg/resource/enqueue_handlers.go
muvaf: >// Owns defines types of Objects being *generated* by the ControllerManagedBy, and configures the ControllerManagedBy to respond to
// create / delete / update events by *reconciling the owner object*.  This is the equivalent of calling
// Watches(&source.Kind{Type: <ForType-forInput>}, &handler.EnqueueRequestForOwner{OwnerType: apiType, IsController: true})

Oh, nevermind. PCs are not owner of PCUs.

pkg/reconciler/providerconfig/reconciler.go
negz: SGTM - I'll probably go with `Active`. I don't like `Connected` in particular, because it could be interpreted as resources have successfully "connected" to the cloud using this `ProviderConfig`, which we don't guarantee.

pkg/reconciler/providerconfig/reconciler.go
negz: > Though I wouldn't frame PCU as the intent to use.

I didn't mean 'declares' in the sense that it's the declarative source of truth. 'Indicates' might have been clearer - i.e. a PCU indicates that a resource is trying or intending to use a PC, not that it has successfully done so.

pkg/reconciler/managed/connector_adapter.go
```diff
@@ -0,0 +1,222 @@
...
+func IsMissingReference(err error) bool {
+	_, ok := err.(interface {
+		MissingReference() bool
+	})
+	return ok
+}
+
+// A ProviderConfigConnector connects to a provider using information extracted
+// from the supplied provider configuration and managed resource.
+type ProviderConfigConnector interface {
```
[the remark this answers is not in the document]
negz: I think that assumes that the `ProviderConfigAdapter` or whatever middleware we use would be responsible for creating the PCU, but not for getting and passing the PC. That feels weird to me - I'd expect us to either be 'full service' and handle both, or handle neither. I suspect the idea that Crossplane should automate the management of PCUs but not automate the fetching of PCs might be would be unusual to folks who aren't as used to the way it works today as we are.

apis/core/v1alpha1/resource.go
negz: I feel like that's not unidiomatic for Kubernetes - e.g node are automatically labelled by the Kubelet with labels that contain the same information as the node object to make it easier to select and match them.

pkg/reconciler/managed/connector_adapter.go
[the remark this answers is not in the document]
muvaf: > I suspect the idea that Crossplane should automate the management of PCUs but not automate the fetching of PCs might be would be unusual to folks who aren't as used to the way it works today as we are.

Conceptually, I see two operations (1) PCU machinery (2) connecting to cloud provider. Fetching the `PC` happens to be a common operation for them but I can't really see it as something weird for both of them doing it separately if that makes the code simpler by not introducing a new interface that is very similar to an existing one. If you feel strongly that `PC` fetching will always happen all the time when managed reconciler is used, we can discuss changing `ExternalConnector` interface to how `ProviderConfigAdapter` looks, i.e. `Connect(ctx context.Context, pc resource.Provider, mg resource.Managed) (ExternalClient, error)`. Though anecdotally, that would've made it more painful to do `Provider` -> `ProviderConfig` if we had designed it that way in the beginning.

In fact, I could even be convinced that managed reconciler could have a separate call than `Connect` to do the PCU operation if we'd like to have it abstracted with its own interface. One is completely Crossplane specific and the other one is completely cloud-related.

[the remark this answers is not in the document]
negz: I honestly feel there's a bias here that handling a PCU is "machinery" that should be done for the provider author while the act of finding and fetching a PC is not. I think the only difference here is that PCU is new and controversial, while finding and fetching a PC is an established pattern that is second nature to us.

That said, I've abandoned the approach of building an adaptor that automates PCU tracking and PC fetching. It could be worth pursuing that approach in future, but it felt like too much abstraction while we still need to deal with the deprecated Provider approach.

pkg/reconciler/providerconfig/reconciler.go
negz: In retrospect, we don't need any inverse condition. It's fine for the `Terminating` condition to appear only when the PC is being deleted - it will never transition out of that condition.

negz:
I abandoned the approach of trying to build connection adaptors that would take care of the PCU creation - it felt too abstracted and messy. It could be worth another try once we don't need to be concerned with the deprecated Provider type.

Instead I've hopefully addressed some of the concern about requiring provider authors to explicitly track PCUs by providing a simple utility that handles doing so for them. This reduces the burden on providers setting up the `ProviderConfig` controller, and calling `u.Track(ctx, mg)` any time they're about to try to use `mg`'s referenced `ProviderConfig`. I like that this makes tracking explicit, like fetching and using a `ProviderConfig` is explicit, but makes tracking hard to do wrong.

muvaf (approved):
LGTM! I love the simplification. So, users will basically add the following in the beginning of their `Connect` command, right?
```
if err := c.Track(ctx, mg); err != nil {
  return err
}
```

They can opt-out if they want their controller to authenticate in a different way.

negz:
> So, users will basically add the following in the beginning of their Connect command, right?

@muvaf Right - see https://github.com/crossplane/provider-aws/pull/353/files#diff-8d5fdb22000146b2ac94a304b53ed4c2R73

I've tested this using the above provider-aws PR and it appears to work as expected. I'm going to merge this and update the rest of the PRs that depend on it.

PR #210, opened 2020-10-02
Update controller runtime to v0.8.0

saschagrunert:

This updates the controller-runtime and all of its dependencies to the
latest release.

This includes a bunch of changes which break backwards compatibility
introduced by controller-runtime 0.7.0 and 0.8.0, for example:

> A number of methods that previously took runtime.Object & internally
> type-asserted them to metav1.Object now take client.Object (for non-list
> objects) or client.ObjectList (for lists). The practical upshot of this
> is more type-safety and clarity around what's required for particular
> methods.
> All concrete API types (anything that implements runtime.Object & has
> a metadata field) already implement client.Object or client.ObjectList,
> so practical impact should be limited to folks who pass around
> runtime.Object values instead of concrete types.

https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.8.0
https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.7.0

I have:

Refers to https://github.com/kubernetes-sigs/seccomp-operator/pull/137


pkg/resource/api_test.go
```diff
@@ -250,17 +250,20 @@ func TestAPIPatchingApplicator(t *testing.T) {
...
-		"NotAMetadataObject": {
-			reason: "An error should be returned if we can't access the object's metadata",
-			c:      &test.MockClient{MockGet: test.NewMockGetFn(errBoom)},
-			args: args{
-				o: &nopeject{},
-			},
-			want: want{
-				o:   &nopeject{},
-				err: errors.New("cannot access object metadata"),
+		// TODO: fix this crashing test
```
saschagrunert, line 253: I have to find out why this test segfaults. :thinking:

hasheddan (changes requested):
@saschagrunert thanks so much for looking into this! For full context, we are currently tagging releases on `master` for this library so we likely will not merge this until cutting a new release next week for us to pin `crossplane` to before its release. After that, we can likely move forward, but also want to see what `controller-runtime` release schedule is so that we don't find ourselves needing new functionality here and depending on an `alpha` release of `controller-runtime` 👍

pkg/resource/api_test.go
saschagrunert: Yeah that's true, the question is if we still need those checks or if we can remove the test case…

pkg/controller/engine.go
```diff
@@ -196,11 +197,11 @@ func (e *Engine) Start(name string, o controller.Options, w ...Watch) error {
 
 	go func() {
 		<-e.mgr.Elected()
-		e.done(name, errors.Wrap(ca.Start(stop), errCrashCache))
+		e.done(name, errors.Wrap(ca.Start(context.Background()), errCrashCache))
```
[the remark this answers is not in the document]
saschagrunert: Yes, I added a new struct `stop` which holds the context as well as the cancel function. We now cancel the context on engine stop.

saschagrunert:
@hasheddan I updated the PR to v0.8.0. Do you think we can give this a fresh look?

pkg/resource/unstructured/client.go
```diff
@@ -53,7 +52,7 @@ type WrapperClient struct {
 // Get retrieves an obj for the given object key from the Kubernetes Cluster.
```
khos2ow, line 52: Newly added `Scheme()` and `RESTMapper()` are missing in `WrapperClient`. I believe adding something like this would work:

```go
// Scheme returns the scheme this client is using.
func (c *WrapperClient) Scheme() *runtime.Scheme {
	return c.kube.Scheme()
}

// RESTMapper returns the rest this client is using.
func (c *WrapperClient) RESTMapper() meta.RESTMapper {
	return c.kube.RESTMapper()
}
```

saschagrunert: Alright, added the two methods as suggested. :+1:

khos2ow (approved):
lgtm (non-binding)

pkg/controller/engine.go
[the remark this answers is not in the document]
hasheddan: @saschagrunert I am missing here where we are cancelling context on engine `Stop`, could you point it out to me? It looks like we still close the stop channel, which is an empty struct.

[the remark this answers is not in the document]
saschagrunert: Ah this has been reverted after my rebase on top of 0.8.0. We have to provide a context, which will be then passed around. The runtime docs says that it will block until the context gets cancelled, so I'll now use the cancelfunc of the context instead of a close channel. WDYT?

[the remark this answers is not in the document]
hasheddan: Looks pretty nice to me :) Let me do a little testing just to validate my assumptions here :+1:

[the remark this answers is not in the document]
saschagrunert: Awesome, thank you! :pray:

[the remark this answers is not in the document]
khos2ow: @hasheddan I've also opened https://github.com/crossplane/crossplane/pull/2082 which might facilitate helping.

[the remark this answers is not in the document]
hasheddan: Awesome, thanks so much @khos2ow!

pkg/controller/engine_test.go
```diff
@@ -123,12 +124,12 @@ func TestEngine(t *testing.T) {
...
 				WithNewCacheFn(func(*rest.Config, cache.Options) (cache.Cache, error) {
-					c := &MockCache{MockStart: func(stop <-chan struct{}) error { return errBoom }}
+					c := &MockCache{MockStart: func(stop context.Context) error { return errBoom }}
 					return c, nil
 				}),
 				WithNewControllerFn(func(string, manager.Manager, controller.Options) (controller.Controller, error) {
-					c := &MockController{MockStart: func(stop <-chan struct{}) error {
-						<-stop
+					c := &MockController{MockStart: func(stop context.Context) error {
+						//TODO:stop
```
hasheddan, line 132: @saschagrunert are these `TODO` comments needed here?

hasheddan (approved):
One nitpick below but otherwise this LGTM, thanks so much @saschagrunert! @negz @muvaf could I get y'all to give this a quick once over since it is a fairly large breaking change?

pkg/controller/engine.go
```diff
@@ -147,14 +148,14 @@ func (e *Engine) done(name string, err error) {
 
 // Watch an object.
 type Watch struct {
-	kind       runtime.Object
+	kind       client.Object
```
negz, line 151: Is this change required due to an upstream change, or would `runtime.Object` still work here? I'd prefer to stick with `runtime.Object` if it's easy to do so, since we use it throughout our codebase.

negz (commented):
This seems good to me. @saschagrunert could you please update the PR description to summarise any headline/relevant changes that have happened since v0.6.2? It would be good to have a little context for the code changes in this PR.

Has anyone looked into what kind of impact updating to controller-runtime 0.8 will have on our providers?

pkg/controller/engine.go
khos2ow: In regard to providers, what I've been briefly tried out, most of them are fine except for some tests which are defining `MockGet`. The signature of `MockGet` (and rest of `MockXXX` for that matter) has changed a providers need a bit of adjustment for this to compile properly on them.

khos2ow (changes requested):
Testing a provider as an example I noticed `Reconcile()` in `pkg/reconciler/providerconfig/reconciler.go` is missing the `context.Context` as first argument.

pkg/controller/engine_test.go
saschagrunert: No, I'll remove them :+1:

saschagrunert:
> Testing a provider as an example I noticed `Reconcile()` in `pkg/reconciler/providerconfig/reconciler.go` is missing the `context.Context` as first argument.

Fixed :+1:

saschagrunert:
> This seems good to me. @saschagrunert could you please update the PR description to summarise any headline/relevant changes that have happened since v0.6.2? It would be good to have a little context for the code changes in this PR.

Yes, I updated the commit message/PR to reflect some of the major breaking changes.

pkg/reconciler/providerconfig/reconciler_test.go
```diff
@@ -27,6 +27,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
hasheddan, line 30: @saschagrunert looks like we need to pass in a context at the bottom of the test block here in our call to `Reconcile()`

saschagrunert: Yes missed that, should be fixed now.

hasheddan (approved):
Awesome, thanks @saschagrunert :)

PR #211, opened 2020-10-03
Move common package linter functionality to crossplane-runtime

hasheddan:

Linter functionality has been implemented in `crossplane/crossplane`, but the core linter loop and helper functions can be generalized for consumption by any `Linter` implementer. Passed `___LinterFn` that depend on `crossplane/crossplane` types will be kept in `c/c`, but will likely be moved in https://github.com/crossplane/crossplane/pull/1795 to the new `xpkg` pkg, which is contains functionality for interacting with Crossplane packages.

This PR also renames `DefaultParser` to `PackageParser`, which is technically a breaking API change, but we have not released since this was added and it won't break downstream consumers who only invoke `parser.New()` then deal with the `Parser` interface methods.

I have:

`make test` and the same functionality already exists in `c/c`

### DIFF
--- a/pkg/parser/linter.go
+++ b/pkg/parser/linter.go
@@ -1,0 +1,109 @@
+/*
+Copyright 2020 The Crossplane Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


pkg/parser/linter.go
```diff
@@ -0,0 +1,111 @@
...
+	return func(o runtime.Object) error {
+		if a == nil || b == nil {
+			return errors.New(errNilLinterFn)
+		}
+		aErr := a(o)
+		bErr := b(o)
+		if aErr == nil || bErr == nil {
+			return nil
+		}
+		return fmt.Errorf(errOrFmt, aErr, bErr)
```
negz, line 109:
```suggestion
		return errors.Errorf(errOrFmt, aErr, bErr)
```

negz (approved):
One nit, but otherwise LGTM

PR #212, opened 2020-10-06
Introduce a distinct ProviderConfig credentials field

negz:

This PR introduces a new `spec.credentials` field that distinguishes credential configuration from other (currently hypothetical) aspects of provider configuration. It also introduces a `spec.credentials.source` field that can be used to specify which credentials source a provider intends to use. Furthermore, it removes the getters and setters for provider config credentials secrets. In practice we don't use these getters and setters anywhere where we aren't operating on a concrete type of provider configuration; they're not actually useful to the `resource.ProviderConfig` interface.

Currently the vast majority of providers support only one credentials source - a secret - but we imagine that others may be supported in future. We recently switched the `CredentialsSecretRef` for most providers from optional, to required, and back to optional. I was about to switch the field back to required in response to https://github.com/crossplane/provider-gcp/pull/262#discussion_r498757852, when I noticed that API conventions [advised against doing so](https://github.com/kubernetes/community/blob/f496e21e418175f9f40a0f11cd42f63bb85c0583/contributors/devel/sig-architecture/api-conventions.md#unions
):

> Sometimes, when a new type is created, the api designer may anticipate that a union will be needed in the future, even if only one field is allowed initially. In this case, be sure to make the field Optional In the validation, you may still return an error if the sole field is unset.

This approach is a compromise; it leaves each possible secret source optional, but requires the expected source to be specified. Providers may choose to avoid embedding this type if they have unique configuration needs, but it should cover most cases, e.g.:

```yaml
apiVersion: gcp.crossplane.io/v1beta1
kind: ProviderConfig
metadata:
  name: default
spec:
  # Fetch credentials from a secret.
  credentials:
    source: Secret
    secretRef:
      namespace: default
      name: gcp
      key: credentials.json
```

```yaml
apiVersion: aws.crossplane.io/v1beta1
kind: ProviderConfig
metadata:
  name: default
spec:
  # Fetch credentials natively (e.g. via IRSA for AWS)
  credentials:
    source: InjectedIdentity
```

I have:

I haven't tested this yet; I'd like to make sure the change is uncontroversial before I do.


apis/core/v1alpha1/resource.go
```diff
@@ -171,13 +171,43 @@ type ProviderSpec struct {
...
+const (
+	// CredentialsSourceNone indicates that a provider does not require
+	// credentials.
+	CredentialsSourceNone CredentialsSource = "None"
+
+	// CredentialsSourceSecret indicates that a provider should acquire
+	// credentials from a secret.
+	CredentialsSourceSecret CredentialsSource = "Secret"
+
+	// CredentialsSourceNative indicates that a provider should acquire
```
muvaf, line 196: As listed here, there are many different auth methods and they are all very specific to each provider. Some could even require more details we don't anticipate right now, like a role ID or different types of _Native_ authentication. I know they can always make additions to this in their provider but by giving a source type, this struct is authoritative over other fields and trying to cover them all from a central place doesn't really feel future-proof to me.

In fact, I think it's a losing game to try to standardize on a `ProviderConfig` format. I would even push for having only some interfaces here in crossplane-runtime (to cover `ProviderConfigUsage` machinery) and leave the whole implementation to providers. Since it's only one kind per provider, maintenance effort would be negligible compared to all other inline structs we have, which usually have N implementations per provider.

negz: > I think it's a losing game to try to standardize on a ProviderConfig format.

It's not a zero sum game. :) I'm not suggesting that all providers _must_ embed this struct; just offering it if they would like to. In my opinion there's value in doing is. It provides:

* A consistent default for providers with simple credential need - e.g. just secrets.
* A consistent way to add a very common field to all providers, if we ever want to do so.
* A reference example for suggested names and docstrings for providers that want to do something different.

I think the fact that this type would be suitable for every extant Crossplane provider speaks to the fact that it's likely to be useful in a lot of cases. If and when providers appear that have different use cases, they could DIY but be guided by this struct; for example perhaps a provider would use the "standard" `source: Native" but implement its own `spec.credentials.native` options.

negz: FWIW, I think we should proceed with adding this struct, but if we didn't I would still advocate for the the API changes it contains to be manually made to every provider. I suspect doing so would result in avoidable, unnessaccary divergence and inconsistency in ProviderConfig APIs over time.

apis/core/v1alpha1/resource.go
```diff
@@ -171,13 +171,43 @@ type ProviderSpec struct {
...
 type ProviderConfigSpec struct {
-	// CredentialsSecretRef references a specific secret's key that contains
-	// the credentials that are used to connect to the provider.
+	// Credentials required to authenticate to this provider.
+	Credentials ProviderCredentials `json:"credentials"`
+}
+
+// A CredentialsSource is a source from which provider credentials may be
+// acquired.
+type CredentialsSource string
```
hasheddan, line 185: I like this pattern! One thing that came to mind as a question is I think that we have discussed before some sort of credentials chain (i.e. if first option doesn't work, use the second, etc.). Do you think that would likely be implemented more at the managed resource level as opposed to the `ProviderConfig` level (i.e. managed resources have a chain of provider config refs rather than `ProviderConfig` having a chain of credentials sources)?

apis/core/v1alpha1/resource.go
muvaf: > I think the fact that this type would be suitable for every extant Crossplane provider speaks to the fact that it's likely to be useful in a lot of cases.

It's only the secret that's common though. For example, if I implement this functionality in provider-aws, I would not use `Native` as source name. I'd go for `IRSA` or `EC2Metadata`, or `WorkloadIdentity` for GCP. `Native` gives clue for all of them but doesn't really cover any of them.

> I suspect doing so would result in avoidable, unnessaccary divergence and inconsistency in ProviderConfig APIs over time.

It's one `kind` in a provider and a set of authentication methods which are already divergent due to nature of cloud authentication. To have a pattern consistency, we can refer to Kubernetes API Patterns and `ProviderConfig` implementation of AWS that has this source pattern with its own source types so that other providers can implement theirs similarly and that'd be pretty much it. It isn't like managed resource patterns that we do everything we can to standardize because of all the common machinery and number of managed resources in each provider. So, I disagree that the divergence is unnecessary.

apis/core/v1alpha1/resource.go
negz: Hrm, that's a good question. I feel that an array here could be a good UX for that, but it would increase the blast radius of this change (i.e. the complexity of all connect methods would be increased):

```yaml
apiVersion: aws.crossplane.io/v1beta1
kind: ProviderConfig
metadata:
  name: default
spec:
  # Fetch credentials natively (i.e. via IRSA)
  credentials:
  - source: Native
  # Fall back to fetching credentials from a secret.
  - source: Secret
     secretRef:
       namespace: default
       name: gcp
       key: credentials.json
```

negz: Ah - there is a way to implement this in future without a breaking API change. @kasey suggested a distinct `Chained` source. I imagine this could take an array of other sources, e.g.:

```yaml
apiVersion: example.org/v1alpha1
kind: ProviderConfig
metadata:
  name: example
spec:
  credentials:
    source: Chained
    chain:
    - source: Native
    - source: ...
```

hasheddan (approved):
This seems like a good standardization while reducing lock-in and not restricting further API flexibility in the future :+1:

negz:
I've tested this with provider-aws; PRs coming soon.

PR #214, opened 2020-10-09
add more test about periods

wonderflow:
Signed-off-by: 天元 <jianbo.sjb@alibaba-inc.com>

Fixes #

I have:

### DIFF
--- a/pkg/fieldpath/fieldpath_test.go
+++ b/pkg/fieldpath/fieldpath_test.go
@@ -84,6 +84,11 @@ func TestFieldOrIndex(t *testing.T) {
 			s:      "'coolField'",
 			want:   Segment{Type: SegmentField, Field: "coolField"},
 		},
+		"QuotedFieldWithPeriods": {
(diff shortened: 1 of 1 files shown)


pkg/fieldpath/fieldpath_test.go
```diff
@@ -212,6 +217,17 @@ func TestParse(t *testing.T) {
 				},
 			},
 		},
+		"QuotedFieldWithPeriod": {
+			reason: "A field name specified using quote can include a period",
+			path:   "metadata.labels[app.hash]",
```
hasheddan, line 222: Did you mean for this test case to have `' '` around some fields? Otherwise, being quoted in this context just means the `path` variable is a string I believe, so not sure if it is testing the use case you want.

wonderflow: Sorry, I have mistakes in writing case. I found this fieldPath package written by Nic is really good to use! I write this case to help users to know more use cases its support.

hasheddan (approved):

PR #216, opened 2020-10-13
Have the managed reconciler update late initialized resources

negz:

A design goal of the managed.Reconciler implementation was to insulate folks who are authoring an ExternalClient implementation from needing to interact with the Kubernetes API directly. The separation of concerns was intended to be:

* ExternalClient accepts a managed resource struct and reconciles it with the provider. It may update the struct it is passed, but otherwise does not interact with the Kubernetes API.
* ExternalConnector interacts with the Kubernetes API in order to produce an ExternalClient.
* Reconciler orchestrates the ExternalConnector and ExternalClient, propagating updates from the ExternalClient back to the Kubernetes API as necessary.

The desire to late initialize managed resources sullied this separation of concern (or single responsibility) by plumbing a Kubernetes client down into the ExternalClient implementations. I believe this was done because:

* We must update the resource spec before we update its status, or changes made to the status will be overwritten.
* We don't want to update the resource spec on every reconcile if it's possible to avoid doing so.

This small change to the managed resource reconciler allows an ExternalClient to signal that it has late initialized a managed resource, and thus defer updating that managed resource to the Reconciler. I believe this change is backward compatible (ExternalClient implementations could continue to 'manually' late initialize the resource and not return ResourceLateInitialized).

I have:

I updated the GCP CloudSQLInstance managed resource in https://github.com/crossplane/provider-gcp/pull/269 and confirmed it works as expected.

### DIFF
--- a/pkg/reconciler/managed/reconciler.go
+++ b/pkg/reconciler/managed/reconciler.go
@@ -282,6 +282,14 @@ type ExternalObservation struct {
 	// the external resource.
 	ResourceUpToDate bool
 
+	// ResourceLateInitialized should be true if the managed resource's spec was
+	// updated during its observation. A Crossplane provider may update a
+	// managed resource's spec fields after it is created or updated, as long as
+	// the updates are limited to setting previously unset fields, and adding
+	// keys to maps. Crossplane uses this information to determine whether there
+	// changes to the spec were made during observation that must be persisted.
+	ResourceLateInitialized bool
+
 	// ConnectionDetails required to connect to this resource. These details
 	// are a set that is collated throughout the managed resource's lifecycle -
 	// i.e. returning new connection details will have no affect on old details
@@ -680,6 +688,15 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error) 
 		record.Event(managed, event.Normal(reasonCreated, "Successfully requested creation of external resource"))
 		managed.SetConditions(v1alpha1.ReconcileSuccess())
 		return reconcile.Result{RequeueAfter: r.shortWait}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
+	}
+
+	if observation.ResourceLateInitialized {
+		if err := r.client.Update(ctx, managed); err != nil {
+			log.Debug(errUpdateManaged, "error", err, "requeue-after", time.Now().Add(r.shortWait))
+			record.Event(managed, event.Warning(reasonCannotUpdate, err))
+			managed.SetConditions(v1alpha1.ReconcileError(errors.Wrap(err, errUpdateManaged)))
+			return reconcile.Result{RequeueAfter: r.shortWait}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
+		}
 	}
 
 	if observation.ResourceUpToDate {
--- a/pkg/reconciler/managed/reconciler.go
+++ b/pkg/reconciler/managed/reconciler.go
@@ -286,8 +286,14 @@ type ExternalObservation struct {
 	// updated during its observation. A Crossplane provider may update a
 	// managed resource's spec fields after it is created or updated, as long as
 	// the updates are limited to setting previously unset fields, and adding
-	// keys to maps. Crossplane uses this information to determine whether there
+	// keys to maps. Crossplane uses this information to determine whether
 	// changes to the spec were made during observation that must be persisted.
+	// Note that changes to the spec will be persisted before changes to the
+	// status, and that pending changes to the status may be lost when the spec
+	// is persisted. Status changes will be persisted by the first subsequent
+	// observation that _does not_ late initialize the managed resource, so it
+	// is important that Observe implementations do not late initialize the
+	// resource every time they are called.
 	ResourceLateInitialized bool
 
 	// ConnectionDetails required to connect to this resource. These details
@@ -691,6 +697,13 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error) 
 	}
 
 	if observation.ResourceLateInitialized {
+		// Note that this update may reset any pending updates to the status of
+		// the managed resource from when it was observed above. This is because
+		// the API server replies to the update with its unchanged view of the
+		// resource's status, which is subsequently deserialized into managed.
+		// This is usually tolerable because the update will implicitly requeue
+		// an immediate reconcile which should re-observe the external resource
+		// and persist its status.
 		if err := r.client.Update(ctx, managed); err != nil {
 			log.Debug(errUpdateManaged, "error", err, "requeue-after", time.Now().Add(r.shortWait))
 			record.Event(managed, event.Warning(reasonCannotUpdate, err))
--- a/pkg/reconciler/managed/reconciler.go
+++ b/pkg/reconciler/managed/reconciler.go
@@ -267,18 +267,46 @@ func (c *NopClient) Delete(ctx context.Context, mg resource.Managed) error { ret
 // An ExternalObservation is the result of an observation of an external
 // resource.
 type ExternalObservation struct {
-	ResourceExists    bool
-	ResourceUpToDate  bool
+	// ResourceExists must be true if a corresponding external resource exists
+	// for the managed resource. Typically this is proven by the presence of an
+	// external resource of the expected kind whose unique identifier matches
+	// the managed resource's external name. Crossplane uses this information to
+	// determine whether it needs to create or delete the external resource.
+	ResourceExists bool
+
+	// ResourceUpToDate should be true if the corresponding external resource
+	// appears to be up-to-date - i.e. updating the external resource to match
+	// the desired state of the managed resource would be a no-op. Keep in mind
+	// that often only a subset of external resource fields can be updated.
+	// Crossplane uses this information to determine whether it needs to update
+	// the external resource.
+	ResourceUpToDate bool
+
+	// ConnectionDetails required to connect to this resource. These details
+	// are a set that is collated throughout the managed resource's lifecycle -
+	// i.e. returning new connection details will have no affect on old details
+	// unless an existing key is overwritten. Crossplane may publish these
+	// credentials to a store (e.g. a Secret).
 	ConnectionDetails ConnectionDetails
 }
 
 // An ExternalCreation is the result of the creation of an external resource.
 type ExternalCreation struct {
+	// ConnectionDetails required to connect to this resource. These details
+	// are a set that is collated throughout the managed resource's lifecycle -
+	// i.e. returning new connection details will have no affect on old details
+	// unless an existing key is overwritten. Crossplane may publish these
+	// credentials to a store (e.g. a Secret).
 	ConnectionDetails ConnectionDetails
 }
 
 // An ExternalUpdate is the result of an update to an external resource.
 type ExternalUpdate struct {
+	// ConnectionDetails required to connect to this resource. These details
+	// are a set that is collated throughout the managed resource's lifecycle -
+	// i.e. returning new connection details will have no affect on old details
+	// unless an existing key is overwritten. Crossplane may publish these
+	// credentials to a store (e.g. a Secret).
 	ConnectionDetails ConnectionDetails
 }
 
--- a/pkg/reconciler/managed/reconciler_test.go
+++ b/pkg/reconciler/managed/reconciler_test.go
@@ -577,6 +577,43 @@ func TestReconciler(t *testing.T) {
 			},
 			want: want{result: reconcile.Result{RequeueAfter: defaultManagedShortWait}},
 		},
+		"LateInitializeUpdateError": {
+			reason: "Errors updating a managed resource to persist late initialized fields should trigger a requeue after a short wait.",
+			args: args{
+				m: &fake.Manager{
+					Client: &test.MockClient{
+						MockGet:    test.NewMockGetFn(nil),
+						MockUpdate: test.NewMockUpdateFn(errBoom),
+						MockStatusUpdate: test.MockStatusUpdateFn(func(_ context.Context, obj runtime.Object, _ ...client.UpdateOption) error {
+							want := &fake.Managed{}
+							want.SetConditions(v1alpha1.ReconcileError(errors.Wrap(errBoom, errUpdateManaged)))
+							if diff := cmp.Diff(want, obj, test.EquateConditions()); diff != "" {
+								reason := "Errors updating a managed resource should be reported as a conditioned status."
+								t.Errorf("\nReason: %s\n-want, +got:\n%s", reason, diff)
+							}
+							return nil
+						}),
+					},
+					Scheme: fake.SchemeWith(&fake.Managed{}),
+				},
+				mg: resource.ManagedKind(fake.GVK(&fake.Managed{})),
+				o: []ReconcilerOption{
+					WithInitializers(),
+					WithReferenceResolver(ReferenceResolverFn(func(_ context.Context, _ resource.Managed) error { return nil })),
+					WithExternalConnecter(ExternalConnectorFn(func(_ context.Context, mg resource.Managed) (ExternalClient, error) {
+						c := &ExternalClientFns{
+							ObserveFn: func(_ context.Context, _ resource.Managed) (ExternalObservation, error) {
+								return ExternalObservation{ResourceExists: true, ResourceUpToDate: true, ResourceLateInitialized: true}, nil
+							},
+						}
+						return c, nil
+					})),
+					WithConnectionPublishers(),
+					WithFinalizer(resource.FinalizerFns{AddFinalizerFn: func(_ context.Context, _ resource.Object) error { return nil }}),
+				},
+			},
+			want: want{result: reconcile.Result{RequeueAfter: defaultManagedShortWait}},
+		},
 		"ExternalResourceUpToDate": {
 			reason: "When the external resource exists and is up to date a requeue should be triggered after a long wait.",
 			args: args{


pkg/reconciler/managed/reconciler.go
```diff
@@ -267,18 +267,54 @@ func (c *NopClient) Delete(ctx context.Context, mg resource.Managed) error { ret
...
+	// that often only a subset of external resource fields can be updated.
+	// Crossplane uses this information to determine whether it needs to update
+	// the external resource.
+	ResourceUpToDate bool
+
+	// ResourceLateInitialized should be true if the managed resource's spec was
+	// updated during its observation. A Crossplane provider may update a
+	// managed resource's spec fields after it is created or updated, as long as
+	// the updates are limited to setting previously unset fields, and adding
+	// keys to maps. Crossplane uses this information to determine whether there
```
hasheddan, line 289: nitpick:
```suggestion
	// keys to maps. Crossplane uses this information to determine whether
```

hasheddan (commented):
I really like this and seems like a low risk change given that current functionality can still be supported 👍 one nitpick below, otherwise LGTM!

pkg/reconciler/managed/reconciler.go
```diff
@@ -654,6 +690,15 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
 		return reconcile.Result{RequeueAfter: r.shortWait}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
 	}
 
+	if observation.ResourceLateInitialized {
```
muvaf, line 693: When we call `client.Update`, it doesn't affect `status` sub-resource and the local object [is updated to reflect the latest state of the whole resource](https://github.com/kubernetes-sigs/controller-runtime/blob/bd97e08c43d5dd89adbe28222d695f9f37bb8037/pkg/client/typed_client.go#L70), meaning if there is any change in `status` subresource up until this point, it will be lost. If we call `client.Status().Update`, then spec & metadata changes [are lost](https://github.com/kubernetes-sigs/controller-runtime/blob/bd97e08c43d5dd89adbe28222d695f9f37bb8037/pkg/client/typed_client.go#L180).

That is the reason why we call `client.Update` in `Observe` right before status manipulating operations, like filling the `atProvider` section. From my reading of the code, in the reconcile passes where `observation.ResourceLateInitialized: true`, we will lose the status changes happened in that reconcile. I like the idea but I think we have to figure out a way to do it without losing information.

hasheddan: @muvaf won't status changes be populated on next reconcile?

muvaf: They will but there are operations like `Update` and `Delete` between the reconciles which might affect what's shown in the status. So, we'd lose some information there.

Another option could be to return early in `Observe` if `ResourceLateInitialized: true` (similar to returning early when resource is not found) and have this code snippet `if observation.ResourceLateInitialized {` right after `Observe`. So, at least we wouldn't make an effective operation until late-init is completed.

negz: @muvaf What you say sounds right to me based on our historical observations and a quick glance at the code you linked, but in practice it doesn't seem to be happening. 🤔 I added a few debug lines to the reconciler, so it looks like:

```go
if observation.ResourceLateInitialized {
	y, _ := yaml.Marshal(managed)
	fmt.Printf("Before: %s\n", y)
	if err := r.client.Update(ctx, managed); err != nil {
		log.Debug(errUpdateManaged, "error", err, "requeue-after", time.Now().Add(r.shortWait))
		record.Event(managed, event.Warning(reasonCannotUpdate, err))
		managed.SetConditions(v1alpha1.ReconcileError(errors.Wrap(err, errUpdateManaged)))
	}
	y, _ = yaml.Marshal(managed)
	fmt.Printf("After: %s\n", y)
}
```

In the logs I see:

```yaml
# Before:
apiVersion: database.gcp.crossplane.io/v1beta1
kind: CloudSQLInstance
metadata:                                    
  annotations:             
    crossplane.io/external-name: my-db-5x8bp-6nktq                                      
  creationTimestamp: "2020-10-14T02:31:03Z"
  finalizers:
  - finalizer.managedresource.crossplane.io
  generateName: my-db-5x8bp-     
  generation: 3           
  labels:                           
    crossplane.io/claim-name: my-db
    crossplane.io/claim-namespace: default
    crossplane.io/composite: my-db-5x8bp
  name: my-db-5x8bp-6nktq 
  ownerReferences:        
  - apiVersion: database.example.org/v1alpha1
    controller: true      
    kind: CompositePostgreSQLInstance
    name: my-db-5x8bp      
    uid: 9530b820-0862-4abe-b3a9-b2ae9f47f8e0
  resourceVersion: "147439"
  selfLink: /apis/database.gcp.crossplane.io/v1beta1/cloudsqlinstances/my-db-5x8bp-6nktq
  uid: 6b3027da-8a01-4f0f-a3ea-d9e9045130a7
spec:                     
  forProvider:                    
    databaseVersion: POSTGRES_9_6
    gceZone: us-central1-a  
    instanceType: CLOUD_SQL_INSTANCE
    region: us-central1                                             
    settings:                             
      activationPolicy: ALWAYS            
      backupConfiguration:
        startTime: "23:00"
      dataDiskSizeGb: 20     
      dataDiskType: PD_SSD                               
      ipConfiguration:          
        authorizedNetworks:
        - value: 0.0.0.0/0
        ipv4Enabled: true  
      locationPreference:                                              
        zone: us-central1-a
      pricingPlan: PER_USE        
      replicationType: SYNCHRONOUS                                                                                                                                                 
      storageAutoResize: true
      tier: db-custom-1-3840
      userLabels:                             
        crossplane-kind: cloudsqlinstance_database_gcp_crossplane_io
        crossplane-name: my-db-5x8bp-6nktq
        crossplane-providerconfig: default
  providerConfigRef:                          
    name: default           
  writeConnectionSecretToRef:
    name: 9530b820-0862-4abe-b3a9-b2ae9f47f8e0-postgresql
    namespace: crossplane-system
status:                                                                                                                                                                                                                                                                                                                                                               
  atProvider:                                                                                                                                                                      
    backendType: SECOND_GEN                                                                                                                                                                                                                                                                                                                                           
    connectionName: crossplane-playground:us-central1:my-db-5x8bp-6nktq                                                                                                                                                                                                                                                                                               
    gceZone: us-central1-a                                                                                                                                                                                                                                                                                                                                            
    project: crossplane-playground                                                                                                                                                                                                                                                                                                                                    
    selfLink: https://sqladmin.googleapis.com/sql/v1beta4/projects/crossplane-playground/instances/my-db-5x8bp-6nktq                                                                                                                                                                                                                                                  
    state: PENDING_CREATE                                                                                                                                                                                                                                                                                                                                             
  conditions:                                                                                                                                                                                                                                                                                                                                                         
  - lastTransitionTime: "2020-10-14T02:31:03Z"                                                                                                                                                                                                                                                                                                                        
    reason: Creating                                                                                                                                                                                                                                                                                                                                                  
    status: "False"                                                                                                                                                                                                                                                                                                                                                   
    type: Ready                                                                                                                                                                                                                                                                                                                                                       
  - lastTransitionTime: "2020-10-14T02:31:07Z"                                                                                                                                                                                                                                                                                                                        
    reason: ReconcileSuccess                                                                                                                                                                                                                                                                                                                                          
    status: "True"                                                                                                                                                                                                                                                                                                                                                    
    type: Synced  
```

```yaml
# After:
apiVersion: database.gcp.crossplane.io/v1beta1
kind: CloudSQLInstance
metadata:
  annotations:
    crossplane.io/external-name: my-db-5x8bp-6nktq
  creationTimestamp: "2020-10-14T02:31:03Z"
  finalizers:
  - finalizer.managedresource.crossplane.io
  generateName: my-db-5x8bp-
  generation: 4
  labels:
    crossplane.io/claim-name: my-db
    crossplane.io/claim-namespace: default
    crossplane.io/composite: my-db-5x8bp
  name: my-db-5x8bp-6nktq
  ownerReferences:
  - apiVersion: database.example.org/v1alpha1
    controller: true
    kind: CompositePostgreSQLInstance
    name: my-db-5x8bp
    uid: 9530b820-0862-4abe-b3a9-b2ae9f47f8e0
  resourceVersion: "147442"
  selfLink: /apis/database.gcp.crossplane.io/v1beta1/cloudsqlinstances/my-db-5x8bp-6nktq
  uid: 6b3027da-8a01-4f0f-a3ea-d9e9045130a7
spec:
  forProvider:
    databaseVersion: POSTGRES_9_6
    gceZone: us-central1-a
    instanceType: CLOUD_SQL_INSTANCE
    region: us-central1
    settings:
      activationPolicy: ALWAYS
      backupConfiguration:
        startTime: "23:00"
      dataDiskSizeGb: 20
      dataDiskType: PD_SSD
      ipConfiguration:
        authorizedNetworks:
        - value: 0.0.0.0/0
        ipv4Enabled: true
      locationPreference:
        zone: us-central1-a
      pricingPlan: PER_USE
      replicationType: SYNCHRONOUS
      storageAutoResize: true
      tier: db-custom-1-3840
      userLabels:
        crossplane-kind: cloudsqlinstance_database_gcp_crossplane_io
        crossplane-name: my-db-5x8bp-6nktq
        crossplane-providerconfig: default
  providerConfigRef:
    name: default
  writeConnectionSecretToRef:
    name: 9530b820-0862-4abe-b3a9-b2ae9f47f8e0-postgresql
    namespace: crossplane-system
status:
  atProvider:
    backendType: SECOND_GEN
    connectionName: crossplane-playground:us-central1:my-db-5x8bp-6nktq
    gceZone: us-central1-a
    project: crossplane-playground
    selfLink: https://sqladmin.googleapis.com/sql/v1beta4/projects/crossplane-playground/instances/my-db-5x8bp-6nktq
    state: PENDING_CREATE
  conditions:
  - lastTransitionTime: "2020-10-14T02:31:03Z"
    reason: Creating
    status: "False"
    type: Ready
  - lastTransitionTime: "2020-10-14T02:31:07Z"
    reason: ReconcileSuccess
    status: "True"
    type: Synced
```

If my mental model is correct, we're seeing something like:

1. New `CloudSQLnstance` MR. `Observe` returns early with `!ResourceExists`, so we `Create` it then requeue.
1. The external CloudSQL instance now exists. We return `ResourceExists` and `ResourceLateInitialized` because we observed some defaulted fields that we need to propagate back to the API server. We skip over the `Create` and `Delete` calls and find ourselves at the `if observation.ResourceLateInitialized` block. The `Observe` call mutated both the spec and status of the object, so we'd expect that calling `Update` would reset the status, but per the above debug output it does not seem to.

This meshes with the debug log ordering, which shows:

```
# First reconcile
2020-10-13T19:31:03.430-0700    DEBUG   provider-gcp    Reconciling     {"controller": "managed/cloudsqlinstance.database.gcp.crossplane.io", "request": "/my-db-5x8bp-6nktq"}
2020-10-13T19:31:07.560-0700    DEBUG   provider-gcp    Successfully requested creation of external resource    {"controller": "managed/cloudsqlinstance.database.gcp.crossplane.io", "request": "/my-db-5x8bp-6nktq", "uid": "6b3027da-8a01-4f0f-a3ea-d9e9045130a7", "version": "147412", "external-name": "", "requeue-after": "2020-10-13T19:31:37.560-0700"}
      
# Second reconcile                
2020-10-13T19:31:07.614-0700    DEBUG   provider-gcp    Reconciling     {"controller": "managed/cloudsqlinstance.database.gcp.crossplane.io", "request": "/my-db-5x8bp-6nktq"}     
Before: REDACTED FOR BREVITY
After: REDACTED FOR BREVITY
2020-10-13T19:31:07.807-0700    DEBUG   provider-gcp    External resource is up to date {"controller": "managed/cloudsqlinstance.database.gcp.crossplane.io", "request": "/my-db-5x8bp-6nktq", "uid": "6b3027da-8a01-4f0f-a3ea-d9e9045130a7", "version": "147439", "external-name": "my-db-5x8bp-6nktq", "requeue-after": "2020-10-13T19:32:07.807-0700"}             
```

negz: It occurred to me that my previous debug logs didn't actually prove that the status was not being reset to what was in the API server - it's possible the in-memory status just happened to match what was in the API server. I added some logic that changes a status field we'd expect the API server to reset when `Update()` was called, but I still can't reproduce the behaviour we'd expect (i.e. the field being reset).

```go
	// In the CloudSQL controller
	if !cmp.Equal(beforeLateInit, &cr.Spec.ForProvider) {
		// Simulate an updated status field that we'd expect to be reset when
		// Update() is called.
		cr.Status.AtProvider.State = "FAKE"
	}
```

In the logs I see:

```yaml
# Before
status:
  atProvider:                              
    backendType: SECOND_GEN                                                                                                                                                       
    connectionName: crossplane-playground:us-central1:my-db-h6ccr-cq6dg                                                                                                           
    gceZone: us-central1-a                                                                                                                                                        
    project: crossplane-playground                                                                                                                                                
    selfLink: https://sqladmin.googleapis.com/sql/v1beta4/projects/crossplane-playground/instances/my-db-h6ccr-cq6dg                                                              
    # Our pending status update that we'd expect to be reset.
    state: FAKE  

# After
atProvider:
    backendType: SECOND_GEN
    connectionName: crossplane-playground:us-central1:my-db-h6ccr-cq6dg
    gceZone: us-central1-a
    project: crossplane-playground
    selfLink: https://sqladmin.googleapis.com/sql/v1beta4/projects/crossplane-playground/instances/my-db-h6ccr-cq6dg
    # Not reset to the API server value
    state: FAKE
```

It's not clear to me why this would be - I'm guessing I must be missing something. Spelunking the code I've confirmed that:

1. If I make a `PUT` (i.e. update) call to the `/status` subresource via curl the entire object (meta, spec, and status) is returned, with any changes I had made to meta or spec unapplied.
1. 1. If I make a `PUT` call to the resource proper via curl the entire object (meta, spec, and status) is returned, with any changes I had made to the status unapplied.
1. Presuming the controller-runtime client is using the JSON decoder from runtime, that eventually boils down to a `jsoniter.Unmarshal` of the JSON returned by the API server into the `*CloudSQLInstance`, which I would expect to overwrite any pending changes as you mentioned.

negz: Aha. With a little more debugging, I think I see why this wasn't reproducing:

```go
	if observation.ResourceLateInitialized {
		o := managed.DeepCopyObject()
		_ = r.client.Get(ctx, types.NamespacedName{Name: managed.GetName()}, o)
		a, _ := yaml.Marshal(o)
		fmt.Printf("# Actual:\n%s\n", a)

		p, _ := yaml.Marshal(managed)
		fmt.Printf("# Pending:\n%s\n", p)
		if err := r.client.Update(ctx, managed); err != nil {
			log.Debug(errUpdateManaged, "error", err, "requeue-after", time.Now().Add(r.shortWait))
			record.Event(managed, event.Warning(reasonCannotUpdate, err))
			managed.SetConditions(v1alpha1.ReconcileError(errors.Wrap(err, errUpdateManaged)))
		}

		u, _ := yaml.Marshal(managed)
		fmt.Printf("# Updated:\n%s\n", u)
	}
```

I now see that when we call `Update()` to persist the late init `Actual` has `status.atProvider: {}`. This makes sense, because the flow in this case is:

1. Create the resource (with an empty status), then requeue.
1. Observe the resource, which populates fields in both `status.atProvider` and `spec.forProvider`.
1. Call `Update()` to late init the resource. This makes an API call that returns a JSON object with `status.atProvider: {}`. Unmarshalling that into our `*CloudSQLInstance` doesn't overwrite any fields.
1. Call `Status.Update()` to persist the pending status changes.

So this implementation works for the CloudSQLInstance controller, but that's because we're depending on the fact that the late init is always persisted before `status.atProvider` is ever set.

negz: > won't status changes be populated on next reconcile?

Right, I'd imagine the flow would then be:

1. At some point in the resource's lifecycle (most likely right after creation) it is late initialized in `Observe`. It's possible some status changes are queued up at this point too.
1. The reconciler calls `kube.Update()` to persist the late init, then either calls `external.Update()` or returns early. There's a risk that the `kube.Update()` call undoes the pending status updates.
1. We're requeued immediately because the `kube.Update()` call changed the resource. We `Observe` again. This time there's not likely to be anything further to late initialize, so we should successfully update our status.

> They will but there are operations like Update and Delete between the reconciles which might affect what's shown in the status. So, we'd lose some information there.

@muvaf How concerning do you think this information loss is? I'm wondering whether we should optimise for (what I suspect will be) simpler reconcile logic, and maybe include a comment near where we call `kube.Update()` to persist the late init clarifying that we may lose pending status updates, but that they would be persisted on the next reconcile.

muvaf: I'm not really too concerned about the data loss (while probably won't hold true in some cases, `status` is assumed to be reproducible) if we'll get it in the next reconcile (hoping it requeues immediately). My main concerns are around:
* Correctness of the returned value `ResourceLateInitialized`. I remember fixing a few bugs because the controller kept assuming a late init is necessary and called `Update()` in each reconcile hence never saving `status`. One could say implementation is buggy so it's on them but if `status` is never stored, then the effect of the bug is doubled.
* Managed reconciler is very low level and used by every managed resource in Crossplane ecosystem. A compromise on correctness of its implementation may cause unexpected bugs in higher levels. So, there is that risk but it's open-ended.

So, it's more about the fact that writers may not account for that loss to happen compared to the value or amount of that data. FWIW, I had the same urge a while ago and didn't pursue because of these reasons but I can also see that it could be worth the risk if it helps other things, like generation tooling. So, I'm OK either way.

muvaf: Oh and thanks for all the debugging!

negz: Got it - I'm thinking I'll try to address this using documentation for now. i.e. I'll update the `ResourceLateInitialized` comment (at least) to explain the risk that status updates may not be persisted.

hasheddan (approved):

PR #218, opened 2020-10-13
parser: symlinks should be followed

### LINKED ISSUE #217: Document that parser does not follow symbolic links
### What happened?

Parser doesn't include the content of the files pointed by the symbolic links. It seems [Walk](https://github.com/spf13/afero/blob/8902da1/path.go#L94) function we use doesn't follow symlinks.

### How can we reproduce it?

Create a directory to build a configuration and have one of the YAMLs to be a symlink to another file outside of that directory.

Crossplane version: 0.13

muvaf:

I was trying to have a symlink in a configuration package in https://github.com/crossplane/crossplane/pull/1836 and realized they are not followed.

This PR implements support to follow symlinks but it's not certain that we'd like to have that functionality because there are some cases where correctly following the symlink is not possible. For example, let's say I create a symlink using:
```
ln -s /Users/monus/go/src/github.com/crossplane/crossplane/docs/snippets/package/definition.yaml /Users/monus/go/src/github.com/crossplane/crossplane/docs/snippets/package/gcp/definition.yaml
```

Then even if the filesystems change, the returned `path` from symlink reader is `/Users/monus/go/src/github.com/crossplane/crossplane/docs/snippets/package/definition.yaml`. The symlink has to be relative to the path where the command is run, which is same with where the symlink exists right now but that may not be a safe assumption once we introduce a flag that will allow users to specify a directory path.

Another option is to ignore it and document this behavior. It's a useful feature but I'm kind of suspicious about its reliability and its effect on reproducibility of the packages. It seems Docker decided [not to support it](https://github.com/moby/moby/issues/1676#issuecomment-23878459) with the same reasons so I'm kind of leaning towards that option.

Fixes #217 

Fixes #

I have:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/google/go-cmp v0.4.0
 	github.com/hashicorp/go-getter v1.4.0
 	github.com/pkg/errors v0.9.1
-	github.com/spf13/afero v1.2.2
+	github.com/spf13/afero v1.4.1
 	golang.org/x/tools v0.0.0-20191018212557-ed542cd5b28a // indirect
 	k8s.io/api v0.18.6
 	k8s.io/apiextensions-apiserver v0.18.6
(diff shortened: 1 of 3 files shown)


hasheddan (commented):
@muvaf I would also lean towards not supporting it and was actually aware that symlinks would not be honored when implementing. I think this is common behavior for a CLI 👍

negz:
Thanks for looking into this @muvaf! I'm happy to avoid adding this functionality if we think it's not a good fit.

That said, I believe the linked Docker issue seems to state that Docker won't `ADD` the symlink itself (i.e. copy the symlink file itself into the image). That makes sense, but https://github.com/moby/moby/issues/1676#issuecomment-23891899 makes (to me) a fairly sound case that it would be useful for Docker to resolve the symlink and `ADD` its target. I feel like that functionality could be useful, and would fundamentally make the build unreproducible (e.g. in our case where the symlink target is part of the same repo as the symlink itself), though I can see that it might open up some security concerns (e.g. a symlink to `/etc/passwd`).

negz:
>  would not fundamentally make the build unreproducible

Whoops - I had originally written "would fundamentally make the build unreproducible" when I meant "would not". Edited.

PR #227, opened 2020-11-05
Allow XRs and XRCs to record when they last published connection details

negz:

See https://github.com/crossplane/crossplane/pull/1927 for details.

I have:

See https://github.com/crossplane/crossplane/pull/1927 for testing details.

### DIFF
--- a/pkg/resource/fake/mocks.go
+++ b/pkg/resource/fake/mocks.go
@@ -181,6 +181,20 @@ func (m *ComposedResourcesReferencer) SetResourceReferences(r []corev1.ObjectRef
 // GetResourceReferences gets the composed references.
 func (m *ComposedResourcesReferencer) GetResourceReferences() []corev1.ObjectReference { return m.Refs }
 
+// ConnectionDetailsLastPublishedTimer is a mock that implements the
+// ConnectionDetailsLastPublishedTimer interface.
+type ConnectionDetailsLastPublishedTimer struct{ Time *metav1.Time }
+
+// SetConnectionDetailsLastPublishedTime sets the published time.
+func (c *ConnectionDetailsLastPublishedTimer) SetConnectionDetailsLastPublishedTime(t *metav1.Time) {
+	c.Time = t
+}
+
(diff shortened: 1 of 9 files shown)


pkg/resource/fake/mocks.go
```diff
@@ -251,7 +265,9 @@ type Composite struct {
 	ComposedResourcesReferencer
 	ClaimReferencer
 	ConnectionSecretWriterTo
+
 	v1alpha1.ConditionedStatus
+	ConnectionDetailsLastPublishedTimer
```
muvaf, line 270: What do you think about having this as a `Condition`? I see that `Equal` check for condition doesn't include the last set time but maybe we can publish the time in the message. That could help keeping the schema simpler.

muvaf (approved):
LGTM! I left a comment suggesting to have this as condition but I don't strongly think it's better either way.

pkg/resource/fake/mocks.go
negz: I considered that - it would be nice but I decided not to because we need a timestamp of some kind (or at least a state change) and working that into a condition felt tricky from a UX perspective. Today we set `lastTransitionTime` on our conditions, but that represents the last time we transitioned between condition statuses so it's not a good fit. API conventions mention that [we could support `lastHeartbeatTime`](https://github.com/kubernetes/community/blob/46f114f1481fec20b08daef0c6e0923c164d5afd/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties) which is a better fit (it's described as the "last time we got an update on a given condition"), but still a little less explicit. i.e.

```yaml
status:
  connectionDetails:
    lastPublishedTime: T
```

Compared to:

```yaml
status:
  conditions:
  - type: PublishedConnectionDetails
    status: True
    lastTransitionTime: T
    lastHeartbeatTime: T+N
```

In the condition example it's not super clear to the user (IMO) whether the "last heartbeat time" is the time the connection details were last published, or just the last time we assessed whether we had published connection details.

Issue #221, opened 2020-10-21
Atomicity for Create call in managed reconciler


muvaf:
### What problem are you facing?

We weere suspecting that this could happen and [it did](https://github.com/crossplane/provider-aws/issues/392) - in case where creation results in a generated ID that we have to store in CR and if that update operation fails, it results in leaked external resource.

### How could Crossplane help solve your problem?

We need some kind of atomicity for that kind of `Create` operations. There are two ways I can think of:
* `ExternalCreation` can have `ResourceLateInitialized` and according to this value we can call spec Update. However, the additional thing we'd do is that if that call fails, we would call `r.managed.Delete` with the same local object.
* Let it be handled by the provider developers, i.e. they'd call `kube.Update` just like today and if that fails they can make deletion API call right there and move on.

muvaf:
In https://github.com/crossplane/provider-aws/issues/392 we have discovered that it's the composition controller getting event because of status update and if `CreateRouteTable` call takes a bit time, then composition controller applies a patch, which renders the local `cr` object in `RouteTable` `Create` call useless for an update. There are a few workarounds like doing a `Patch` instead of `Update` but we'll see the same problem if that operation fails for some other reason, like network problem during the HTTP call of `Update`/`Patch`.

PR #228, opened 2020-11-05
managed reconciler: more resilient update call for saving identifiers retrieved from provider

muvaf:

~~This PR makes sure external resource is deleted in case it's late initalized and we couldn't store its external identifier. I was thinking about using `Patch` to make sure we don't run into `object has changed` error but as @negz pointed out in Slack, `Patch` is also subject to `resourceVersion` constraint. Using our `APIPatchingApplicatior` reduces the chances of error, however, in case it fails we lose the identifier since it updates what you pass in with a `Get` call. So, I decided using usual `Update` with a rollback mechanism would be most future proof.~~

~~It's still not completely atomic but I'd say it's good enough to cover most cases. In case of a leak where `kube.Update` and `external.Delete` failed for some reason, we'll report it via an event.~~

We try our best to save the retrieved identifier by a force update and exponential backoff.

Fixes https://github.com/crossplane/crossplane-runtime/issues/221

I have:

I have tested it for success case and it's working. Then I used a fake error in the supplied func of `e.Run` like the following:
```
			if err := e.Run(ctx, func() error {
				_ = c.Apply(ctx, managed)
				err := errors.New("some error")
				if err != nil {
					log.Debug(errUpdateManaged, "error", err, "requeue-after", time.Now().Add(r.shortWait))
					record.Event(managed, event.Warning(reasonCannotUpdateManaged, errors.Wrap(err, fmt.Sprintf("error time: %s", time.Now().String()))))
				}
				return err
			}); err != nil {
				log.Debug(errUpdateAfterCreate, "error", err, "requeue-after", time.Now().Add(r.shortWait))
				record.Event(managed, event.Warning(reasonCannotUpdateManaged, errors.Wrap(err, errUpdateAfterCreate)))
				managed.SetConditions(v1alpha1.ReconcileError(errors.Wrap(err, errUpdateAfterCreate)))
			}
```

 Here is what happens in the worst case scenario:
```
Events:
  Type     Reason                       Age   From                               Message
  ----     ------                       ----  ----                               -------
  Normal   CreatedExternalResource      115s  managed/vpc.ec2.aws.crossplane.io  Successfully requested creation of external resource
  Warning  CannotUpdateManagedResource  115s  managed/vpc.ec2.aws.crossplane.io  err time: 2020-11-16 18:31:50.267872 +0300 +03 m=+8.895027127: some error
  Warning  CannotUpdateManagedResource  113s  managed/vpc.ec2.aws.crossplane.io  err time: 2020-11-16 18:31:52.273531 +0300 +03 m=+10.900636635: some error
  Warning  CannotUpdateManagedResource  109s  managed/vpc.ec2.aws.crossplane.io  err time: 2020-11-16 18:31:56.282473 +0300 +03 m=+14.909477639: some error
  Warning  CannotUpdateManagedResource  101s  managed/vpc.ec2.aws.crossplane.io  err time: 2020-11-16 18:32:04.289003 +0300 +03 m=+22.915804099: some error
  Warning  CannotUpdateManagedResource  85s   managed/vpc.ec2.aws.crossplane.io  err time: 2020-11-16 18:32:20.294885 +0300 +03 m=+38.921277832: some error
  Warning  CannotUpdateManagedResource  65s   managed/vpc.ec2.aws.crossplane.io  err time: 2020-11-16 18:32:40.301324 +0300 +03 m=+58.927204242: some error
  Warning  CannotUpdateManagedResource  45s   managed/vpc.ec2.aws.crossplane.io  err time: 2020-11-16 18:33:00.313363 +0300 +03 m=+78.938729299: some error
  Warning  CannotUpdateManagedResource  25s   managed/vpc.ec2.aws.crossplane.io  cannot update managed resource. this may have resulted in a leaked external resource: context deadline exceeded
  Normal   UpdatedExternalResource      22s   managed/vpc.ec2.aws.crossplane.io  Successfully requested update of external resource
```

It tries with exponential backoff but once context deadline is exceeded, it gives up. Then in the next reconcile, things work as usual as expected since Apply was being called.

### DIFF
--- a/pkg/reconciler/managed/reconciler.go
+++ b/pkg/reconciler/managed/reconciler.go
@@ -21,6 +21,9 @@ import (
 	"strings"
 	"time"
 
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/retry"
+
 	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -45,25 +48,27 @@ const (
 
 // Error strings.
 const (
-	errGetManaged       = "cannot get managed resource"
-	errReconcileConnect = "connect failed"
-	errReconcileObserve = "observe failed"
-	errReconcileCreate  = "create failed"
-	errReconcileUpdate  = "update failed"
-	errReconcileDelete  = "delete failed"
+	errGetManaged               = "cannot get managed resource"
+	errUpdateManagedAfterCreate = "cannot update managed resource. this may have resulted in a leaked external resource"
+	errReconcileConnect         = "connect failed"
+	errReconcileObserve         = "observe failed"
+	errReconcileCreate          = "create failed"
+	errReconcileUpdate          = "update failed"
+	errReconcileDelete          = "delete failed"
 )
 
 // Event reasons.
 const (
-	reasonCannotConnect     event.Reason = "CannotConnectToProvider"
-	reasonCannotInitialize  event.Reason = "CannotInitializeManagedResource"
-	reasonCannotResolveRefs event.Reason = "CannotResolveResourceReferences"
-	reasonCannotObserve     event.Reason = "CannotObserveExternalResource"
-	reasonCannotCreate      event.Reason = "CannotCreateExternalResource"
-	reasonCannotDelete      event.Reason = "CannotDeleteExternalResource"
-	reasonCannotPublish     event.Reason = "CannotPublishConnectionDetails"
-	reasonCannotUnpublish   event.Reason = "CannotUnpublishConnectionDetails"
-	reasonCannotUpdate      event.Reason = "CannotUpdateExternalResource"
+	reasonCannotConnect       event.Reason = "CannotConnectToProvider"
+	reasonCannotInitialize    event.Reason = "CannotInitializeManagedResource"
+	reasonCannotResolveRefs   event.Reason = "CannotResolveResourceReferences"
+	reasonCannotObserve       event.Reason = "CannotObserveExternalResource"
+	reasonCannotCreate        event.Reason = "CannotCreateExternalResource"
+	reasonCannotDelete        event.Reason = "CannotDeleteExternalResource"
+	reasonCannotPublish       event.Reason = "CannotPublishConnectionDetails"
+	reasonCannotUnpublish     event.Reason = "CannotUnpublishConnectionDetails"
+	reasonCannotUpdate        event.Reason = "CannotUpdateExternalResource"
+	reasonCannotUpdateManaged event.Reason = "CannotUpdateManagedResource"
 
 	reasonDeleted event.Reason = "DeletedExternalResource"
 	reasonCreated event.Reason = "CreatedExternalResource"
@@ -306,6 +311,12 @@ type ExternalObservation struct {
 
 // An ExternalCreation is the result of the creation of an external resource.
 type ExternalCreation struct {
+	// ExternalNameAssigned is true if the Create operation resulted in a change
+	// in the external name annotation. If that's the case, we need to issue a
+	// spec update and make sure it goes through so that we don't lose the identifier
+	// of the resource we just created.
+	ExternalNameAssigned bool
+
 	// ConnectionDetails required to connect to this resource. These details
 	// are a set that is collated throughout the managed resource's lifecycle -
 	// i.e. returning new connection details will have no affect on old details
@@ -676,6 +687,28 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error) 
 			return reconcile.Result{RequeueAfter: r.shortWait}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
 		}
 
+		if creation.ExternalNameAssigned {
+			en := meta.GetExternalName(managed)
+			// We will retry in all cases where the error comes from the api-server.
+			// At one point, context deadline will be exceeded and we'll get out
+			// of the loop. In that case, we warn the user that the external resource
+			// might be leaked.
+			err := retry.OnError(retry.DefaultRetry, resource.IsAPIError, func() error {
+				nn := types.NamespacedName{Name: managed.GetName()}
+				if err := r.client.Get(ctx, nn, managed); err != nil {
+					return err
+				}
+				meta.SetExternalName(managed, en)
+				return r.client.Update(ctx, managed)
+			})
+			if err != nil {
+				log.Debug("Cannot update managed resource", "error", err, "requeue-after", time.Now().Add(r.shortWait))
+				record.Event(managed, event.Warning(reasonCannotUpdateManaged, errors.Wrap(err, errUpdateManagedAfterCreate)))
+				managed.SetConditions(v1alpha1.ReconcileError(errors.Wrap(err, errUpdateManagedAfterCreate)))
+				return reconcile.Result{RequeueAfter: r.shortWait}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
+			}
+		}
+
 		if err := r.managed.PublishConnection(ctx, managed, creation.ConnectionDetails); err != nil {
 			// If this is the first time we encounter this issue we'll be
 			// requeued implicitly when we update our status with the new error
@@ -706,7 +739,7 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error) 
 		// and persist its status.
 		if err := r.client.Update(ctx, managed); err != nil {
 			log.Debug(errUpdateManaged, "error", err, "requeue-after", time.Now().Add(r.shortWait))
-			record.Event(managed, event.Warning(reasonCannotUpdate, err))
+			record.Event(managed, event.Warning(reasonCannotUpdateManaged, err))
 			managed.SetConditions(v1alpha1.ReconcileError(errors.Wrap(err, errUpdateManaged)))
 			return reconcile.Result{RequeueAfter: r.shortWait}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
 		}
--- a/pkg/reconciler/managed/reconciler_test.go
+++ b/pkg/reconciler/managed/reconciler_test.go
@@ -20,11 +20,10 @@ import (
 	"context"
 	"testing"
 
-	"sigs.k8s.io/controller-runtime/pkg/client"
-
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/pkg/errors"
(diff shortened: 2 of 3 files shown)


pkg/reconciler/managed/reconciler.go
```diff
@@ -306,6 +308,15 @@ type ExternalObservation struct {
 
 // An ExternalCreation is the result of the creation of an external resource.
 type ExternalCreation struct {
+	// ResourceLateInitialized should be true if the managed resource's spec was
+	// updated during its creation. A Crossplane provider may update a
```
hasheddan, line 312: nitpick: this includes annotations (e.g. external-name) and potentially other things outside of the spec right?

pkg/reconciler/managed/reconciler.go
```diff
@@ -306,6 +308,15 @@ type ExternalObservation struct {
...
+	// managed resource's spec fields or metadata in order to store the persistent
+	// ID of the external resource if it's generated by the Provider API.
+	// Note that changes to the spec will be persisted to api-server but status
+	// changes will be lost in this reconciliation if ResourceLateInitialized is
+	// true.
+	ResourceLateInitialized bool
```
hasheddan, line 318: I wonder if it would be worth defining a new concept that is separate from "Late Initialization" for this case that is specific to identification information. However, it does fall into the category of late initialization, so maybe we should just be consistent

pkg/reconciler/managed/reconciler.go
```diff
@@ -675,6 +686,34 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
 			managed.SetConditions(v1alpha1.ReconcileError(errors.Wrap(err, errReconcileCreate)))
 			return reconcile.Result{RequeueAfter: r.shortWait}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
 		}
+		log.Debug("Successfully requested creation of external resource", "requeue-after", time.Now().Add(r.shortWait))
+		record.Event(managed, event.Normal(reasonCreated, "Successfully requested creation of external resource"))
+
+		if creation.ResourceLateInitialized {
+			err := r.client.Update(ctx, managed)
+			if err != nil {
```
negz, line 694: I'd prefer to be able to avoid making superfluous calls to the external API if possible (i.e. create, delete, then create again). Do we have a good idea of why updates fail here? Is it due to write contention with the XR controller, i.e. having a stale resource version?

If we're seeing updates fail for some reason other than a stale resource version perhaps this is one niche case in which retrying the update inside the same reconcile loop would work; we could use exponential backoff with a small retry cap.

If we are seeing updates fail mostly due to stale resource versions things get trickier, but there are some options that don't require us to delete and recreate the resource in the provider's API, i.e.:

* Patch the resource after creation instead of updating it. This could be sufficient if we're confident that we'll only ever really be calling update to persist the resource's newly allocated external name annotation.
* The nuclear option; do something like the `APIUpdatingApplicator` does and retry the update with a manually overridden fresh resource version.
* Require that resources that generate their own IDs be somehow annotated by the Create call such that we could find them again on subsequent creates if they were orphaned? i.e. Tag VPCs with the `kube-system` namespace's UID and the name of the managed resource.
* Create some new type that acts as a temporary store of leaked external resources? Something like:

```yaml
apiVersion: crossplane.io/v1alpha1
kind: LeakedResource
metadata:
  name: example
managedResourceRef:
  apiVersion: ec2.crossplane.io/v1beta1
  kind: VPC
  name: example
externalName: vpc-a828k2dd9
```

pkg/reconciler/managed/reconciler.go
negz: I had the same thought. Intuitively I initially considered this to be different from late initialisation, though I can see the case for it being considered another form of late initialisation. Are we aware of cases in which the create call "late inits" a managed resource in any way other than setting its external name?

pkg/reconciler/managed/reconciler.go
muvaf: > If we are seeing updates fail mostly due to stale resource versions things get trickier.

I believe this is the case.

> The nuclear option; do something like the APIUpdatingApplicator does and retry the update with a manually overridden fresh resource version.

This sounds good to me. One risk is the case where we're actually stale, i.e. accidentally override some user input. We can use `APIUpdatingApplicator` to get the resource, change the annotation and let it update, i.e. do not touch any other thing meanwhile. Alternatively, we could do that via patching where the patch is manually curated in the code to have only the annotation so that we're sure that's the only thing we touch.

We can have an option in the applicator we use for exponential backoff so that we cover the case where api-server itself is down for a while (might be limited to 30 secs which is what we have in context for non-external-client operations).

> Require that resources that generate their own IDs be somehow annotated by the Create call such that we could find them again on subsequent creates if they were orphaned? i.e. Tag VPCs with the kube-system namespace's UID and the name of the managed resource.

I'm not sure if that's possible for all cases. A lot of resources have separate APIs for tagging, which could fail during creation of the resource and also brings in its own complications like what to do for cases where we want to import the resource.

> Create some new type that acts as a temporary store of leaked external resources?

I'm not sure. Nuke option seems simpler to manage to me.

pkg/reconciler/managed/reconciler.go
muvaf: > Intuitively I initially considered this to be different from late initialisation, though I can see the case for it being considered another form of late initialisation.

I feel like it falls under late-init. There is a default assigned by the API and we fetch it to update the not-given user input. One difference from the late-init we do in `Observe` is that, we override the user's input even if they did assign a value. Though if we go with the nuke option that [strictly updates only the external-name annotation](https://github.com/crossplane/crossplane-runtime/pull/228#discussion_r524164123), calling this late-init would be misleading.

> Are we aware of cases in which the create call "late inits" a managed resource in any way other than setting its external name?

I'm not sure but it might be possible where there are more than one string used to identify the resource and both could be generated by the API.

muvaf:
I have pushed two commits. The first one includes the implementation of nuke option and the second one wraps the call with an runner that tries the call with exponential backoff period. Let me know what you guys think, I don't feel strongly about the exponential backoff TBH.

pkg/reconciler/managed/reconciler.go
negz: > One difference from the late-init we do in Observe is that, we override the user's input even if they did assign a value.

I'll let you make the call here, but this doesn't feel like a late init to me.

pkg/reconciler/managed/reconciler.go
```diff
@@ -675,6 +686,19 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
 			managed.SetConditions(v1alpha1.ReconcileError(errors.Wrap(err, errReconcileCreate)))
 			return reconcile.Result{RequeueAfter: r.shortWait}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
 		}
+		log.Debug("Successfully requested creation of external resource", "requeue-after", time.Now().Add(r.shortWait))
+		record.Event(managed, event.Normal(reasonCreated, "Successfully requested creation of external resource"))
+
+		if creation.ResourceLateInitialized {
+			// We use APIUpdatingApplicator to enforce the update to go through
+			// by setting object's resourceVersion to the latest.
+			if err := resource.NewAPIUpdatingApplicator(r.client).Apply(ctx, managed); err != nil {
```
negz: I'd prefer us to have this entire reconciler use a `resource.ClientApplicator` instead of a `client.Client` that we can mock out rather than explicitly creating a `NewAPIUpdatingApplicator` inline.

pkg/reconciler/managed/reconciler.go
```diff
@@ -690,9 +690,18 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
...
 			// We use APIUpdatingApplicator to enforce the update to go through
 			// by setting object's resourceVersion to the latest.
-			if err := resource.NewAPIUpdatingApplicator(r.client).Apply(ctx, managed); err != nil {
+			c := resource.NewAPIUpdatingApplicator(r.client)
+			if err := e.Run(ctx, func() error {
+				err := c.Apply(ctx, managed)
+				if err != nil {
+					log.Debug(errUpdateManaged, "error", err, "requeue-after", time.Now().Add(r.shortWait))
+					record.Event(managed, event.Warning(reasonCannotUpdateManaged, errors.Wrap(err, errUpdateManaged)))
+				}
```
negz, line 702: How does this debugging info correspond to the debugging info below at line 705? Do we need both?

pkg/reconciler/managed/reconciler.go
```diff
@@ -690,9 +690,18 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
...
 
 		if creation.ResourceLateInitialized {
+			e := resource.NewRunnerWithExponentialBackoff(16 * time.Second)
 			// We use APIUpdatingApplicator to enforce the update to go through
 			// by setting object's resourceVersion to the latest.
-			if err := resource.NewAPIUpdatingApplicator(r.client).Apply(ctx, managed); err != nil {
+			c := resource.NewAPIUpdatingApplicator(r.client)
+			if err := e.Run(ctx, func() error {
+				err := c.Apply(ctx, managed)
+				if err != nil {
```
negz, line 699: Nit: This could be `if err := c.Apply(ctx, managed); err != nil {}`. It could also just be `return c.Apply(...)` if we don't need the debugging info inside this function.

pkg/resource/api.go
```diff
@@ -204,6 +207,47 @@ func (a *APIUpdatingApplicator) Apply(ctx context.Context, o runtime.Object, ao
 	return errors.Wrap(a.client.Update(ctx, m), "cannot update object")
 }
 
+// ExponentialBackoff runs the given function with retries in exponentially increasing
+// backoff periods with a given maximum.
+type ExponentialBackoff struct {
+	// MaximumBackoff will be used as waiting period if the calculated backoff
+	// exceeds it.
+	MaximumBackoff time.Duration
+}
```
negz, line 216: https://godoc.org/k8s.io/client-go/util/retry#RetryOnConflict

I went to take a look whether client-go had utilities for exponential backoff and found the above function, which seems at first glance like it could replace most of this PR. I'd prefer to use `RetryOnConflict` rather than an updating applicator if possible.

negz (changes requested):

pkg/resource/api.go
muvaf: Thanks for finding that, it's exactly what I was looking for.

I got it implemented and it's working. However, it does a retry if and only if the error is a conflict. Under the hood, the function is actually a wrap around `OnError`:
```golang
func RetryOnConflict(backoff wait.Backoff, fn func() error) error {
	return OnError(backoff, errors.IsConflict, fn)
}
```

As you can see, we're able to give an error detection function and I think we should keep retrying for all kinds of api-server errors, like `ServerTimeout` or `ServiceUnavailable` etc. that could happen during either in `Get` or `Update` calls so that we don't have to come back again to this problem in other cases where api-server is down for instance. So, I found an interface in `kerrors` opackage that covers all api-server errors that we can use for filtering. As a result, until the context deadline is exceeded (which is an `error` that doesn't satisfy `APIStatus` interface), it'll keep retrying and the total context timeout is max 1.30min, so we'll do our actual best to store the external name.

negz (approved):
Awesome, thank you for fixing this!

hasheddan (approved):
LGTM!

Issue #229, opened 2020-11-16
Provide more informative errors in package parser


hasheddan:
### What problem are you facing?

Currently the package parser will return generic, "object-level" errors, such as not recognizing an object when there is an issue with the schema. It also has no concept of files, which can make it difficult to find where the error resides when building a large package.

### How could Crossplane help solve your problem?

The parser should provider more informative errors, which will likely require passing down logic from the parser backends to provide additional context (e.g. file path from the `FsBackend`).

PR #232, opened 2020-12-08
Support AnnotatedReaders in package parser

hasheddan:

Adds the `AnnotatedReader` interface and an implementation in `FsReadCloser`. This allows a parser backend to provide additional context about the data being read if an error is encountered.

xref #229 

I have:

See https://github.com/crossplane/crossplane/pull/2014 for usage.

### DIFF
--- a/pkg/parser/fsreader.go
+++ b/pkg/parser/fsreader.go
@@ -24,6 +24,14 @@ import (
 	"github.com/spf13/afero"
 )
 
+var _ AnnotatedReadCloser = &FsReadCloser{}
+
+// FsReadCloserAnnotation annotates data for an FsReadCloser.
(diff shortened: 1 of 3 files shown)


negz (approved):

pkg/parser/fsreader.go
```diff
@@ -56,6 +65,16 @@ func SkipDirs() FilterFn {
 	}
 }
 
+// SkipEmpty skips empty files.
+func SkipEmpty() FilterFn {
+	return func(path string, info os.FileInfo) (bool, error) {
+		if info.Size() == 0 {
+			return true, nil
+		}
+		return false, nil
```
negz, line 74: Nit: I believe this could be `return info.Size() == 0, nil`

PR #235, opened 2021-01-12
Introduce IgnoreAny to be able to ignore more than one type of error

muvaf:

I needed this to be able to append more `ErrorIs` functions to generated call from another file. I think even the original `Ignore` could be deprecated since this one covers it as well but I know it's used in many places so I kept it. Let me know of your thoughts about whether we should deprecate it and use this everywhere instead.

I have:

Unit tests.


hasheddan (approved):
LGTM :+1: I would be fine with deprecating `Ignore`

negz:
I'd slightly prefer to keep the original `Ignore`. It's simple enough that I don't think it's a maintenance burden to keep, and `IgnoreAny` as a name makes more sense when you actually pass more than one error.

PR #240, opened 2021-01-28
Support additional ProviderConfig cred sources and provide tooling for extraction

### LINKED ISSUE #236: Support Filesystem and Environment CredentialsSource in ProviderConfig

Crossplane providers all currently support a `ProviderConfig` type that specifies how the provider should obtain credentials for a given reconcile operation. These `ProviderConfig`s typically embed the [ProviderConfigSpec](https://github.com/crossplane/crossplane-runtime/blob/c674db318cdb5bd326bab5595fd94fa11293cb3a/apis/common/v1/resource.go#L178) struct, which has common credentials sources, such as `Secret`, `InjectedIdentity`, and `None`. However, there are other ways that credentials may be injected into a provider `Pod` that do not currently fall into these categories. While providers are not required to implement this exact `ProviderConfigSpec`, it should include common credentials sources to allow for more providers to use it directly.

Two of the most common sources for credentials are injection in the container's filesystem and setting of environment variables. These may be done directly using something like [`ContollerConfig`](https://doc.crds.dev/github.com/crossplane/crossplane/pkg.crossplane.io/ControllerConfig/v1alpha1@v1.0.0), or by an external system, such as the [Vault agent's mutating webhook](https://learn.hashicorp.com/tutorials/vault/agent-kubernetes). The `ProviderConfigSpec` should support `CredentialsSource` of `Filesystem` and `Environment` for these use-cases, as well as embed `fs` and `env` structs for specifying how to extract the credentials. Examples for each are shown below:
```yaml
apiVersion: aws.crossplane.io/v1beta1
kind: ProviderConfig
metadata:
  name: from-fs
spec:
  credentials:
    source: Filesystem
    fs:
      path: /crossplane/secrets/aws-creds.txt
---
apiVersion: aws.crossplane.io/v1beta1
kind: ProviderConfig
metadata:
  name: from-env
spec:
  credentials:
    source: Environment
    env:
      name: MY_AWS_CREDS
``` 

`crossplane-runtime` should also provide common functions for extracting these values, such that the `Connect` methods can be simplified for reconciliation loops written in each provider. This would likely take the form of an engine that accepted a `ProviderConfigSpec` and had handlers for extracting `Filesystem`, `Environment`, and `Secret` credentials sources, then accepted additional handlers for provider-specific sources, such as `InjectedIdentity`, or overriding the previously mentioned handlers.

One area this design is still slightly rough is when a provider wants to go outside the bounds of the common `ProviderConfigSpec` functionality. A good example of this is if a `provider-aws` user wanted to use an injected SA token for IRSA for two separate roles (see example [here](https://github.com/aws/amazon-eks-pod-identity-webhook/issues/32#issuecomment-588113740)). It would be nice to be able to provide a custom struct such as `irsaOverrides`:

```yaml
apiVersion: aws.crossplane.io/v1beta1
kind: ProviderConfig
metadata:
  name: other-irsa
spec:
  credentials:
    source: InjectedIdentity
    irsaOverrides:
      awsRoleArnEnvVar: AWS_ROLE_ARN_WEST
```

However, this would not be possible with the current `ProviderConfigSpec` if used directly (this is not a new restriction, but rather the current status of the `ProviderConfigSpec` design). As mentioned above, there is always the opportunity for a provider to not use the `ProviderConfigSpec`, but it is highly encouraged. Options to make this a more seamless experience could be to mix and match source and struct:

```yaml
apiVersion: aws.crossplane.io/v1beta1
kind: ProviderConfig
metadata:
  name: other-irsa
spec:
  credentials:
    source: InjectedIdentity
    env:
      name: AWS_ROLE_ARN_2 # provider-aws knows this overrides AWS_ROLE_ARN when InjectedIdentity is specified
```

Or to encourage only the embedding of the [`ProviderCredentials` struct](https://github.com/crossplane/crossplane-runtime/blob/c674db318cdb5bd326bab5595fd94fa11293cb3a/apis/common/v1/resource.go#L204), have that passed to the aforementioned "handler engine", then allow the provider to define other fields in the spec:

```yaml
apiVersion: aws.crossplane.io/v1beta1
kind: ProviderConfig
metadata:
  name: other-irsa
spec:
  credentials:
    source: InjectedIdentity
  irsaOverrides: # note this now lives at same level as credentials in the spec as opposed to nested under it
    awsRoleArnEnvVar: AWS_ROLE_ARN_WEST
```

I think this model would be my preference, and could be done without actually impacting the current API of `ProviderConfig` in each provider.

cc @negz @muvaf @kasey @prasek

hasheddan:

To allow for more flexible credential methods, we provide common
selectors but do not enforce source enum or restrict from adding
additional selectors. The CredentialsSelectors are meant to be emdbedded
inline in a ProviderConfig's spec.credentials object.

Fixes #236 

I have:

Unit testing is present, but will link to usage in an existing provider so the changes required to consume can be observed before moving forward.


apis/common/v1/resource.go
```diff
@@ -161,55 +161,31 @@ type ResourceStatus struct {
...
-// A ProviderSpec defines the common way to get to the necessary objects to
-// connect to the provider.
-// Deprecated: Please use ProviderConfigSpec.
-type ProviderSpec struct {
-	// CredentialsSecretRef references a specific secret's key that contains
-	// the credentials that are used to connect to the provider.
-	// +optional
-	CredentialsSecretRef *SecretKeySelector `json:"credentialsSecretRef,omitempty"`
+// CredentialsSelectors provides common selectors for extracting credentials.
+type CredentialsSelectors struct {
```
hasheddan, line 165: maybe just:
```suggestion
type CredentialSelectors struct {
```

pkg/resource/providerconfig.go
```diff
@@ -46,6 +56,119 @@ func IsMissingReference(err error) bool {
...
+// WithOverride constructs a new HandlerOverride.
+func WithOverride(key string, handler ExtractionHandlerFn) *HandlerOverride {
+	return &HandlerOverride{
+		key:     key,
+		handler: handler,
+	}
+}
+
+// NewProviderConfigCredentialExtractor returns a new credential extractor for ProviderConfigs.
+func NewProviderConfigCredentialExtractor(client client.Client, overrides ...*HandlerOverride) *ProviderConfigCredentialExtractor {
```
hasheddan, line 149: we pass the client here so that we can have `ExtractSecretFn` be registered by default, but it does feel kind of weird.

pkg/resource/providerconfig.go
```diff
@@ -46,6 +56,119 @@ func IsMissingReference(err error) bool {
...
+type ExtractionHandlerFn func(ctx context.Context, c interface{}) ([]byte, error)
+
+// EnvLookupFn looks up an environment variable.
+type EnvLookupFn func(string) string
+
+// OsEnvLookup looks up an environment variable from the OS.
+func OsEnvLookup(s string) string {
+	e, _ := os.LookupEnv(s)
+	return e
+}
```
negz, line 75: As far as I can tell this function is identical to `os.Getenv`. Could we use that instead?

apis/common/v1/resource.go
```diff
@@ -161,55 +161,31 @@ type ResourceStatus struct {
...
-	Credentials ProviderCredentials `json:"credentials"`
+// EnvSelector selects an environment variable.
+type EnvSelector struct {
+	// Name is the name of an environment variable.
+	Name string `json:"name,omitempty"`
 }
 
-// A CredentialsSource is a source from which provider credentials may be
-// acquired.
-type CredentialsSource string
```
negz, line 185: Per our discussion yesterday, I still prefer to keep this as an enum for UX purposes. Perhaps a good compromise would be to:

1. Keep this type definition here, along with a few of the most common credential sources that providers may (optionally) choose to support.
2. Remove the generic `ProviderConfigSpec`, and instead have each provider implement their own.
3. Define standard `EnvSelector`, `FsSelector`, `SecretKeySelector` etc structs here that providers may also choose to support.
4. Define simple helper functions for each type - i.e. a type that takes a `SecretKeySelector` and a client and returns the secret data. These functions wouldn't be attempting to satisfy a consistent type or interface - providers would be responsible for switching on the CredentialsSource and calling the appropriate function with the appropriate inputs (i.e. the appropriate selector struct).

apis/common/v1/resource.go
```diff
@@ -161,55 +161,31 @@ type ResourceStatus struct {
...
-// provider config may embed this type in its spec in order to support standard
-// fields. Provider configs may choose to avoid embedding this type as
-// appropriate, but are encouraged to follow its conventions.
-type ProviderConfigSpec struct {
-	// Credentials required to authenticate to this provider.
-	Credentials ProviderCredentials `json:"credentials"`
+// EnvSelector selects an environment variable.
+type EnvSelector struct {
+	// Name is the name of an environment variable.
+	Name string `json:"name,omitempty"`
```
negz, line 182: Should these be omitempty?

apis/common/v1/resource.go
hasheddan: @negz I think we can still have the advantages the enum provides with the approach I have here by letting each provider define their `Source` as an enum of the types they support, right? This makes it such that we can still provide the default handlers and avoid forcing provider authors to implement their own switch. More info here https://github.com/crossplane/crossplane-runtime/issues/236#issuecomment-769062449

apis/common/v1/resource.go
hasheddan: Nope! Thanks for catching!

pkg/resource/providerconfig.go
hasheddan: It serves the exact same purpose, just makes it more testable. I could be convinced to remove it and set the env var in the test directly if desired.

hasheddan: Either way, it should be updated to just return `os.GetEnv` rather than using `os.LookupEnv`, but I like the quality of wrapping it nonetheless.

apis/common/v1/resource.go
hasheddan: @negz I opened a PR to show how this would look on a ProviderConfig CRD in practice. I think my extractors are going to have to be modified at least partially because we are going to have to pass a different struct for those defined as defaults vs additional selectors added by provider implementation, but the source enum works pretty well IMO. Let me know your thoughts https://github.com/crossplane/provider-aws/pull/518/files#diff-1201d66f172379e6b70b7ffed32ee04906876a8c748c6397e278ee3bf933e28aR107

negz: Thanks - that illustration helps a lot. This approach LGTM, though I'd suggest we define the `CredentialsSource` type and the common sources (none, secret, env, fs) in this package, rather than defining them in each provider given that AFAICT all providers will be expected to know about those sources (because they embed CredentialsSelectors). Each provider could still define and use additional sources.

pkg/resource/providerconfig.go
```diff
@@ -46,6 +56,119 @@ func IsMissingReference(err error) bool {
...
 }
 
+// A CredentialExtractor extracts a provider's credentials.
+type CredentialExtractor interface {
+	Extract(context.Context, string, interface{})
+}
+
+// An ExtractionHandlerFn is a function that can extract a specific type of
+// credentials.
+type ExtractionHandlerFn func(ctx context.Context, c interface{}) ([]byte, error)
```
negz, line 66: Does this take `interface{}` instead of `xpv1.CredentialsSelectors` because we expect that some `ExtractionHandlerFn` functions would be built using a different kind of struct as input? If so we might consider narrowing it down a bit by using our own interface. e.g.:

```go
type CredentialSelector interface {
    // A dummy method that types can implement to indicate they're a credential selector.
    CredentialSelector() bool
}
```

I think this could help with readability/code documentation.

pkg/resource/providerconfig.go
```diff
@@ -46,6 +56,119 @@ func IsMissingReference(err error) bool {
...
+			"Filesystem":  ExtractFsFn(afero.NewOsFs()),
+			"Environment": ExtractEnvFn(OsEnvLookup),
+			"Secret":      ExtractSecretFn(client),
+		},
+	}
+	for _, o := range overrides {
+		o.Override(e.handlers)
+	}
+	return e
+}
```
negz, line 161: This seems like a bit of a roundabout way to build and return a map of handlers. Presuming we want to be able to extend the base set of handlers perhaps it would be more straightforward to go with something like:

```go
type ProviderConfigCredentialHandlers map[CredentialsSource]ExtractionHandlerFn

func NewProviderConfigCredentialHandlers(c client.Client) ProviderConfigCredentialHandlers {
	return ProviderConfigCredentialHandlers{
		CredentialsSourceFs:  ExtractFsFn(afero.NewOsFs()),
		CredentialsSourceEnv: ExtractEnvFn(OsEnvLookup),
		CredentialsSourceSecret:      ExtractSecretFn(c),
	},
}
```

For the caller this would look like:

```go
h := NewProviderConfigCredentialHandlers(client)
h[CredentialsSourceCustom] = ExtractCustomFn()
```

Or if you want to keep the ability to do it all in one call you might consider reducing the `HandlerOverride` to a simple struct - e.g.

```go
type ProviderConfigCredentialHandlers map[CredentialsSource]ExtractionHandlerFn

type Handler struct {
	Source CredentialsSource
	Fn ExtractionHandlerFn
}

func NewProviderConfigCredentialHandlers(c client.Client, h ...Handler) ProviderConfigCredentialHandlers {
	hds := ProviderConfigCredentialHandlers{
		CredentialsSourceFs:  ExtractFsFn(afero.NewOsFs()),
		CredentialsSourceEnv: ExtractEnvFn(OsEnvLookup),
		CredentialsSourceSecret:      ExtractSecretFn(c),
	},
	for _, hd := range h {
		hds[hd.Source] = hd.Fn
	}
	return hds
}
```

pkg/resource/providerconfig.go
hasheddan: Good point, didn't see that until I updated to `os.Getenv` from `os.LookupEnv` :+1:

apis/common/v1/resource.go
hasheddan: That sounds fine to me :+1:

pkg/resource/providerconfig.go
hasheddan: I ended up just passing the `CommonCredentialSelectors` to the builder functions, which feels a little weird, but enables folks to define their custom functions with other struct types as well without having to use the interface method defined above. Would be interested to hear if you have a preference one way or the other?

pkg/resource/providerconfig.go
hasheddan: I ended up with this, which feels kind of like we are wrapping a map for no reason, but I do like the idea of being able to substitute a different extractor in the future if needed:

```
// ProviderConfigCredentialExtractor extracts credentials as specified by a ProviderConfig.
type ProviderConfigCredentialExtractor map[xpv1.CredentialsSource]ExtractionHandlerFn

// NewProviderConfigCredentialExtractor returns a new credential extractor for ProviderConfigs.
func NewProviderConfigCredentialExtractor(client client.Client, c xpv1.CommonCredentialSelectors) ProviderConfigCredentialExtractor {
	return ProviderConfigCredentialExtractor{
		xpv1.CredentialsSourceFilesystem:  ExtractFsFn(afero.NewOsFs(), c),
		xpv1.CredentialsSourceEnvironment: ExtractEnvFn(os.Getenv, c),
		xpv1.CredentialsSourceSecret:      ExtractSecretFn(client, c),
	}
}

// Register adds additional handlers to the extractor.
func (p ProviderConfigCredentialExtractor) Register(source xpv1.CredentialsSource, fn ExtractionHandlerFn) {
	p[source] = fn
}

// Extract extracts credentials as specified by a ProviderConfig.
func (p ProviderConfigCredentialExtractor) Extract(ctx context.Context, source xpv1.CredentialsSource) ([]byte, error) {
	fn, ok := p[source]
	if !ok {
		return nil, errors.Errorf(errNoHandlerForSourceFmt, source)
	}
	return fn(ctx)
}
```

Thoughts?

hasheddan: Been thinking about / experimenting with this more and I think I am leaning more towards just putting these common sources in a `switch` statement in something like `CommonSourceExtactor(source xpv1.CredentialsSource, client client.Client)`, then allowing consumers to use it directly if they only use the common methods, or otherwise just make it the `default` in their larger `switch`:

```
switch s := pc.Spec.Credentials.Source; s { //nolint:exhaustive
	case v1beta1.CredentialSourceCustomOne:
		// handle custom source one
	case v1beta1.CredentialSourceCustomTwo:
		// handle custom source two
	default:
		resource.CommonSourceExtractor(s, client)
	}
```

They could override one of the common extractors by having a separate case for it earlier in the `switch`, or just support the extraction methods directly by using the individual ones embedded in `resource.CommonSourceExtractor`.

hasheddan:
@negz I need to clean up commit history but should be ready to go otherwise :+1:

apis/common/v1/resource.go
```diff
@@ -198,20 +179,44 @@ const (
...
-// ProviderCredentials required to authenticate.
-type ProviderCredentials struct {
-	// Source of the provider credentials.
-	// +kubebuilder:validation:Enum=None;Secret;InjectedIdentity
-	Source CredentialsSource `json:"source"`
+// CommonCredentialSelectors provides common selectors for extracting
+// credentials.
+type CommonCredentialSelectors struct {
+	// Fs is a reference to a filesystem location that contains credentials that
+	// must be used to connect to the provider. +optional
```
negz, line 196: Looks like the `+optional` tags got wrapped here.

pkg/resource/providerconfig.go
```diff
@@ -46,6 +55,50 @@ func IsMissingReference(err error) bool {
...
+// CommonCredentialExtractor extracts credentials from common sources.
+func CommonCredentialExtractor(ctx context.Context, source xpv1.CredentialsSource, client client.Client, selector xpv1.CommonCredentialSelectors) ([]byte, error) {
+	switch source { // nolint:exhaustive
+	case xpv1.CredentialsSourceEnvironment:
+		return ExtractEnv(ctx, os.Getenv, selector)
+	case xpv1.CredentialsSourceFilesystem:
+		return ExtractFs(ctx, afero.NewOsFs(), selector)
+	case xpv1.CredentialsSourceSecret:
+		return ExtractSecret(ctx, client, selector)
+	}
```
negz, line 98: Should we add a case for `None` that just does nothing (i.e. doesn't return creds or an error)?

negz (approved):

pkg/resource/providerconfig.go
hasheddan: I was a little concerned with whether we would want to return `nil` bytes with `nil` error if users were just passing that along to some sort of client constructor. They could always override though so it is likely fine.

PR #241, opened 2021-02-02
Remove deprecated ProviderReference

hasheddan:

Removes the deprecated ProviderReference field in favor of moving to
only using ProviderConfigReference.

I have:

n/a


negz:
Should we just leave this here, given this package is `v1`? I forget where we landed with regard to `common/v1` following proper Kubernetes versioning semantics (since it just includes subtypes), but presumably leaving the type in place can't hurt too much (even if most or all providers cease to use it).

hasheddan:
@negz providers will be unable to use the encompassing struct if they don't want to have a `providerRef` on resources from now on. Are you thinking that providers will continue to support `providerRef` in perpetuity?

negz:
@hasheddan Ah my bad, for some reason I was thinking this was a type definition rather than a field in a struct.

I don't think we want to continue to support `providerRef` in perpetuity, but you could definitely argue removing it is a breaking change (that line is a little blurry for runtime and providers tough). I suspect we have a few breaking changes on the horizon as part of our intending to take a critical eye to the XRM. It would be ideal to batch this with those breaking changes, but if it's particularly onerous to maintain support while you add our new env and fs credential sources I could be convinced to drop it now.

hasheddan:
@negz I think we can wait to batch them together :) I finished up the update in provider-aws and it was fine without this update :+1:

muvaf:
What would be the way to handle this breaking change? We now have the conversion webhook abilities so we can bump the API schemas, however, we can't create a `ProviderConfig` object and mutate `providerConfigRef` in those conversion functions. It seems like we'll have to break and provide manual instructions even if the resources are v1beta1 or v1?

negz:
I do think we need to do this one day, but this PR probably won't be it. 😄

PR #242, opened 2021-02-11
managed reconciler: set deleting and creating conditions before calling respective functions

muvaf:

We set these conditions in `ExternalClient` implementations but they are easy to forget and there isn't a scenario I can think of that we'd not want to set them.
I have:

Unit tests.


negz:
I think originally the idea behind letting the ExternalClients handle this was because I speculated that concepts like "deleting" and "creating" might have different names for some resources (e.g. "provisioning"). In practice that hasn't really been the case so this seems reasonable.

negz (approved):

hasheddan (approved):

Issue #40, opened 2019-10-08
Requeue with exponential backoff


negz:
### What problem are you facing?

Both the resource claim and managed resource reconcilers requeue a reconcile after exactly 30 seconds when they encounter an error (unless the error is known to be unrecoverable). Ideally we'd retry a little sooner with a reasonably capped exponential backoff.

Note that controller-runtime applies exponential backoff when a reconciler returns `{Requeue: true}`, and/or an error. Unfortunately the backoff starts at 5ms and doubles until it reaches a little over 16 minutes, and this is not configurable. This leads to situations like https://github.com/crossplaneio/crossplane/issues/241, in which a resource could take much longer to reconcile than expected, for reasons that are not obvious to users.

### How could Crossplane help solve your problem?

The various reconcilers we ship in crossplane-runtime could support exponential backoff. I believe it's important that this backoff be:

* Bounded appropriately for the controller in question..
* Discoverable by users who do not have access to Crossplane and/or Stack pod logs

We could implement this ourselves by manipulating `RequeueAfter`, but ideally we'd avoid reinventing controller-runtime's wheels. I've raised https://github.com/kubernetes-sigs/controller-runtime/issues/631 requesting controller-runtime allow us to plug in our own backoff logic.

negz:
Note that configurable exponential backoff (https://github.com/kubernetes-sigs/controller-runtime/pull/731) made it into https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.5.1.

hasheddan:
`controller-runtime` takes a pretty simple approach based on what you return in `reconcile.Result`. You can view the source [here](https://github.com/kubernetes-sigs/controller-runtime/blob/16bf3ad036b908d897543c415fcc0bafc5cec711/pkg/internal/controller/controller.go#L297), but it essentially boils down to (in order):
1. If you return an error it will add the request to the queue with the passed rate limiting interface.
2. If you return `RequeueAfter: <vaule>`, the request is added to the queue with the specified delay.
3. If you return `Requeue: true`, the request is added to the queue with the passed rate limiting interface.

[...]

Issue #198, opened 2020-09-10
Deprecate the 'Synced' status condition


negz:
### What problem are you facing?

Most Crossplane resources (managed, claims, composites, etc) typically expose two types of status condition:

* `Ready` indicates whether the thing being reconciled (e.g. the external resource) is expected to be functional.
* `Synced` indicates whether the most recent reconcile attempt succeeded or failed, including any error.

Sometimes `Ready` goes by different names - e.g. `Established` for XRDs.

Originally Crossplane only had a `Ready` condition, but it was conceptually conflated with "synced" and attempted to reflect both whether the underlying resource was ready and whether Crossplane was successfully able to reconcile the underlying resource. The `Synced` condition has served us well for some time, but I think it should be removed because:

* It's duplicated by events. Anecdotally, we emit an event corresponding to ~95% of our `SetConditions` calls.
* It's duplicated by debug logs in the other ~5% of cases, which are 'internal' errors like being unable to remove a finalizer.
* It's a footgun when writing multiple controllers that reconcile the same kind of resource.

The apiextensions 'definition' and 'offered' controllers are examples of the latter issue. Both controllers collaborate to reconcile a `CompositeResourceDefinition`. The `definition` controller manages the lifecycle of the defined composite resource and its controller, while the `offered` controller manages the lifecycle of the offered resource claim (if any) and its controller. Currently both reconcilers attempt to set the `Synced` condition of the XRD; if one controller was reconciling successfully and the other was encountering an error they would get stuck in a tight reconcile loop.

### How could Crossplane help solve your problem?

I suggest we deprecate and stop using the `Synced` condition in any reconciler that has been updated to emit events and logs.

muvaf:
I believe conditions are most useful in machine to machine communication, which is not really what events/logs are for. However, `Synced` doesn't really mean much for other controllers or systems, at least I haven't seen us writing code to check `Synced` condition, it's mostly the `Ready` one. So, in the places we're able to communicate that `there is a misconfiguration problem` to the user via another mechanism like event, I agree that we don't need `Synced` condition.

muvaf:
Hmm, actually UI systems are usually built using YAML output of the resources, which do not have the events. So that's a machine usage of the `Synced` condition; to show user that there is a misconfiguration through the UI. I believe Upbound Cloud is built this way. I went to look at `Pod` to see what conditions it has:
```yaml
  conditions:
  - lastProbeTime: null
[...]

PR #243, opened 2021-02-11
Make use of rate limiting in managed reconciler

hasheddan:

Updates managed reconciler to return Requeue: true instead of
RequeueAfter when encountering non-status update errors. This allows the
reconciler to make use of the controller rate limiter rather use a
constant value for requeuing after errors.

This also renames longWait to pollInterval to more accurately reflect
its behavior, and adds default rate limiters to be used at both the provider and
controller level. The provider rate limiter is a configurable token
bucket and the controller limiter is a max of limiter that uses the
provider limiter and a per-item exponential backoff.

xref #40 

I have:

Unit tests.

### DIFF
--- a/pkg/ratelimiter/default.go
+++ b/pkg/ratelimiter/default.go
@@ -1,0 +1,48 @@
+/*
+Copyright 2021 The Crossplane Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package ratelimiter
+
+import (
+	"time"
+
+	"golang.org/x/time/rate"
+	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/ratelimiter"
+)
+
+// DefaultProviderRPS is the recommended default average requeues per second
+// tolerated by a provider's controller manager.
+const DefaultProviderRPS = 1
+
+// NewDefaultProviderRateLimiter returns a token bucket rate limiter meant for
+// limiting the number of average total requeues per second for all controllers
+// registered with a controller manager. The bucket size is a linear function of
+// the requeues per second.
+func NewDefaultProviderRateLimiter(rps int) *workqueue.BucketRateLimiter {
+	return &workqueue.BucketRateLimiter{Limiter: rate.NewLimiter(rate.Limit(rps), rps*10)}
+}
+
+// NewDefaultManagedRateLimiter returns a rate limiter that takes the maximum
+// delay between the passed providerRateLimiter and a per-item exponential
+// backoff limiter. The exponential backoff limiter has a base delay of 1s and a
+// maximum of 60s.
+func NewDefaultManagedRateLimiter(providerRateLimiter ratelimiter.RateLimiter) ratelimiter.RateLimiter {
+	return workqueue.NewMaxOfRateLimiter(
+		workqueue.NewItemExponentialFailureRateLimiter(1*time.Second, 60*time.Second),
+		providerRateLimiter,
+	)
+}
--- a/pkg/ratelimiter/default.go
+++ b/pkg/ratelimiter/default.go
@@ -37,12 +37,11 @@ func NewDefaultProviderRateLimiter(rps int) *workqueue.BucketRateLimiter {
 }
 
 // NewDefaultManagedRateLimiter returns a rate limiter that takes the maximum
-// delay between the passed providerRateLimiter and a per-item exponential
-// backoff limiter. The exponential backoff limiter has a base delay of 1s and a
-// maximum of 60s.
-func NewDefaultManagedRateLimiter(providerRateLimiter ratelimiter.RateLimiter) ratelimiter.RateLimiter {
+// delay between the passed provider and a per-item exponential backoff limiter.
+// The exponential backoff limiter has a base delay of 1s and a maximum of 60s.
+func NewDefaultManagedRateLimiter(provider ratelimiter.RateLimiter) ratelimiter.RateLimiter {
 	return workqueue.NewMaxOfRateLimiter(
 		workqueue.NewItemExponentialFailureRateLimiter(1*time.Second, 60*time.Second),
-		providerRateLimiter,
+		provider,
 	)
 }
--- a/pkg/reconciler/managed/reconciler.go
+++ b/pkg/reconciler/managed/reconciler.go
@@ -530,8 +530,8 @@ func (r *Reconciler) Reconcile(_ context.Context, req reconcile.Request) (reconc
 
 	if err := r.managed.Initialize(ctx, managed); err != nil {
 		// If this is the first time we encounter this issue we'll be requeued
-		// implicitly when we update our status with the new error condition.
-		// If not, we want to try again after a short wait.
+		// implicitly when we update our status with the new error condition. If
+		// not, we requeue explicitly, which will trigger backoff.
 		log.Debug("Cannot initialize managed resource", "error", err)
 		record.Event(managed, event.Warning(reasonCannotInitialize, err))
 		managed.SetConditions(xpv1.ReconcileError(err))
@@ -550,9 +550,10 @@ func (r *Reconciler) Reconcile(_ context.Context, req reconcile.Request) (reconc
 	if !meta.WasDeleted(managed) {
 		if err := r.managed.ResolveReferences(ctx, managed); err != nil {
 			// If any of our referenced resources are not yet ready (or if we
-			// encountered an error resolving them) we want to try again after a
-			// short wait. If this is the first time we encounter this situation
-			// we'll be requeued implicitly due to the status update.
+			// encountered an error resolving them) we want to try again. If
+			// this is the first time we encounter this situation we'll be
+			// requeued implicitly due to the status update. If not, we want
+			// requeue explicitly, which will trigger backoff.
 			log.Debug("Cannot resolve managed resource references", "error", err)
 			record.Event(managed, event.Warning(reasonCannotResolveRefs, err))
 			managed.SetConditions(xpv1.ReconcileError(err))
@@ -565,7 +566,8 @@ func (r *Reconciler) Reconcile(_ context.Context, req reconcile.Request) (reconc
 		// We'll usually hit this case if our Provider or its secret are missing
 		// or invalid. If this is first time we encounter this issue we'll be
 		// requeued implicitly when we update our status with the new error
-		// condition. If not, we want to try again after a short wait.
+		// condition. If not, we requeue explicitly, which will trigger
+		// backoff.
 		log.Debug("Cannot connect to provider", "error", err)
 		record.Event(managed, event.Warning(reasonCannotConnect, err))
 		managed.SetConditions(xpv1.ReconcileError(errors.Wrap(err, errReconcileConnect)))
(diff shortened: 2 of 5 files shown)


pkg/reconciler/managed/reconciler.go
```diff
@@ -510,6 +511,19 @@ func NewReconciler(m manager.Manager, of resource.ManagedKind, o ...ReconcilerOp
...
 
+// maybeBackoff updates the managed resource status and returns an error in the
+// case that backoff is enabled and the passed error is not nil. Otherwise, the
+// request is requeued with specified wait time, and the error result of the
+// update operation is returned. If the update operation error result is not
+// nil, backoff will also be triggered implicitly.
+func (r *Reconciler) maybeBackoff(ctx context.Context, m resource.Managed, wait time.Duration, err error) (reconcile.Result, error) {
+	updateErr := errors.Wrap(r.client.Status().Update(ctx, m), errUpdateManagedStatus)
+	if r.useBackoff && err != nil {
+		return reconcile.Result{}, err
```
hasheddan, line 522: We could concatenate the update error in the case where we both pass an error, backoff is enabled, and the status update fails.

pkg/reconciler/managed/reconciler.go
```diff
@@ -510,6 +511,19 @@ func NewReconciler(m manager.Manager, of resource.ManagedKind, o ...ReconcilerOp
...
+	}
+	return reconcile.Result{RequeueAfter: wait}, updateErr
+}
+
```
negz, line 526: I'm a little averse to wrappers that produce return values like this, but I can't think of a better approach if we want to allow folks to pick between backoff or fixed requeuafter values.

Do you think we could just straight-up switch to using backoff? i.e. If/when folks bump to this version of runtime they'd be using backoff? Providers could cut over at their own leisure.

hasheddan: I would feel pretty good about just switching over if we include a default implementation (and potentially some variations) that are recommended because we are essentially just using an extremely naive rate limiting function right now as it is. Would just need to explicitly call out in release notes though or else consumers would be defaulting to backing off to ~16 minutes with the controller-runtime implementation. The long wait would remain 1 minute thought so likely not too big of a problem.

muvaf: > If/when folks bump to this version of runtime they'd be using backoff? Providers could cut over at their own leisure.

That sounds good to me.

> I would feel pretty good about just switching over if we include a default implementation (and potentially some variations) that are recommended because we are essentially just using an extremely naive rate limiting function right now as it is.

I agree that a simple default rate limiter would go a long way here. I believe the big three providers would just opt for using the default, maybe except provider-azure since its requests take longer than others.

pkg/reconciler/managed/reconciler.go
```diff
@@ -42,8 +42,7 @@ const (
 	reconcileGracePeriod = 30 * time.Second
 	reconcileTimeout     = 1 * time.Minute
 
-	defaultManagedShortWait = 30 * time.Second
-	defaultManagedLongWait  = 1 * time.Minute
+	defaultManagedLongWait = 1 * time.Minute
```
negz, line 45: Nit: Maybe long doesn't make sense in context anymore, since we don't have a short wait? I tend to think of this as the "speculative" or "happy path" poll interval these days.

pkg/reconciler/managed/reconciler.go
```diff
@@ -545,10 +545,10 @@ func (r *Reconciler) Reconcile(_ context.Context, req reconcile.Request) (reconc
 		// If this is the first time we encounter this issue we'll be requeued
 		// implicitly when we update our status with the new error condition.
 		// If not, we want to try again after a short wait.
-		log.Debug("Cannot initialize managed resource", "error", err, "requeue-after", time.Now().Add(r.shortWait))
+		log.Debug("Cannot initialize managed resource", "error", err)
```
negz, line 548: We don't need to bring this into scope for this PR, but it would be great if we could eventually make our backoff state inspectable such that we could keep something like the `requeue-after` context. I'd feel more comfortable with longer backoffs (e.g. multiple minutes) if we could provide users with context as to how long our backoff was and why.

pkg/reconciler/managed/reconciler.go
```diff
@@ -545,10 +545,10 @@ func (r *Reconciler) Reconcile(_ context.Context, req reconcile.Request) (reconc
...
 		record.Event(managed, event.Warning(reasonCannotInitialize, err))
 		managed.SetConditions(xpv1.ReconcileError(err))
-		return reconcile.Result{RequeueAfter: r.shortWait}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
+		return reconcile.Result{Requeue: true}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
```
negz, line 551: Would we be able to just `return reconcile.Result, err` in these situations if we deprecated the `Synced` status per https://github.com/crossplane/crossplane-runtime/issues/198? Just curious - we don't need to bring that into scope with this PR.

pkg/reconciler/managed/reconciler.go
hasheddan: @negz I was considering that as well, but didn't know if we want to the `WithLongWait()` on update. I think most providers don't actually use that option (i.e. they default to the 1 minute), and it might also be good for the break to be exposed to give a stronger indication that something has changed behind the scenes.

pkg/reconciler/managed/reconciler.go
hasheddan: Yeah unfortunately we don't have access to the rate limiting interface within the reconciler (it is set at the controller level) and the controller-runtime default actually doesn't allow any sort of "peek" method that doesn't also increment the failure count used for backoff. This was part of my motivation for having a "layer 2" rate limiter because we could:
1. Set a default without folks having to explicitly pass it in to the controller.
2. Decouple the external API backoff from the k8s API server one.
3. Do things like log what the backoff value is currently at.

pkg/reconciler/managed/reconciler.go
hasheddan: I believe we still need to call the status update in some scenarios (such as #242). One thing we could do it concatenate the `err` from the method that was called with the status update. I played around with this a bit initially though and it felt pretty sloppy.

negz: Ah, that's true. We also need to assume that at the `Observe` method (and possibly also `Create` and `Update`) could be mutating the managed resource's status.

That said, perhaps it's acceptable for us to only persist these status updates in the happy path? i.e. In the return statements that follow any place we'd currently log "Successfully Xed the external resource". When we hit an error case we'd return the error without persisting the managed resource's status. When we didn't hit an error we'd call `r.client.Status().Update()` as we do today, and return that error if there was one.

Without having thought through it too much I feel like this is probably acceptable? Managed resource viewers should be able to tell that we're trying to create or delete a managed resource because we'll be emitting warning events like `CannotCreateExternalResource`. Any status fields that would be set by the `ExternalClient` methods would presumably just get set once the error cleared and we were able to proceed.

pkg/reconciler/managed/reconciler.go
negz: We'd be making a fairly equivalent breaking change in removing `WithShortWait()`, right? We could also have WithLongWait become an alias for `WithPollInterval` or whatever we called this now.

pkg/reconciler/managed/reconciler.go
negz: I'd prefer to tackle this (someday) at the controller-runtime level if possible. It would be nice if, for example, the `controller.Manager` had a method we could call to ask what the soonest time we'd be requeued at was.

hasheddan:
Example usage: https://github.com/crossplane/provider-aws/pull/544

pkg/reconciler/managed/reconciler.go
```diff
@@ -400,24 +398,14 @@ func WithTimeout(duration time.Duration) ReconcilerOption {
...
-// new reconciliation in 'short wait' scenarios. The Reconciler requeues after a
-// short wait when it knows it is waiting for an external operation to complete,
-// or when it encounters a potentially temporary error.
-func WithShortWait(after time.Duration) ReconcilerOption {
+// WithpollInterval specifies how long the Reconciler should wait before queueing
+// a new reconciliation after a successful reconcile. The Reconciler requeues
+// after a specified duration when it is not actively waiting for an external
+// operation, but wishes to check whether an existing external resource needs to
+// be synced to its Crossplane Managed resource.
+func WithpollInterval(after time.Duration) ReconcilerOption {
```
muvaf, line 406:
```suggestion
func WithPollInterval(after time.Duration) ReconcilerOption {
```

pkg/reconciler/managed/reconciler.go
muvaf: > perhaps it's acceptable for us to only persist these status updates in the happy path? i.e. In the return statements that follow any place we'd currently log "Successfully Xed the external resource". When we hit an error case we'd return the error without persisting the managed resource's status.

Would that result in errors not appearing in conditions, hence shown only as event?

muvaf (approved):
LGTM! Thanks for tackling this long-running issue @hasheddan ! Solid solution.

pkg/reconciler/managed/reconciler.go
hasheddan: @muvaf yep, I don't think we can do this currently

pkg/ratelimiter/default.go
```diff
@@ -0,0 +1,48 @@
...
+// DefaultProviderRPS is the recommended default average requeues per second
+// tolerated by a provider's controller manager.
+const DefaultProviderRPS = 1
+
+// NewDefaultProviderRateLimiter returns a token bucket rate limiter meant for
+// limiting the number of average total requeues per second for all controllers
+// registered with a controller manager. The bucket size is a linear function of
+// the requeues per second.
+func NewDefaultProviderRateLimiter(rps int) *workqueue.BucketRateLimiter {
+	return &workqueue.BucketRateLimiter{Limiter: rate.NewLimiter(rate.Limit(rps), rps*10)}
```
hasheddan, line 36: Using `10` as a linear constant here to determine bucket size is likely not suitable for every provider. However, providers may use their own `BucketRateLimiter` with a different value. Using the number of controllers under management is not an awful proxy, but it really depends on actual runtime number of objects / user preference. I feel pretty good about having this as the default and letting providers adjust / expose configuration as needed.

Note: if a user wanted a strict cap of requeues per second they would likely need to set bucket size = 1 here.

hasheddan (commented):
@negz I adjusted the managed reconciler comments as you pointed out. Would love for you to get a final look at this before merge.

pkg/ratelimiter/default.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+// NewDefaultManagedRateLimiter returns a rate limiter that takes the maximum
+// delay between the passed providerRateLimiter and a per-item exponential
+// backoff limiter. The exponential backoff limiter has a base delay of 1s and a
+// maximum of 60s.
+func NewDefaultManagedRateLimiter(providerRateLimiter ratelimiter.RateLimiter) ratelimiter.RateLimiter {
```
negz, line 43:
```suggestion
func NewDefaultManagedRateLimiter(provider ratelimiter.RateLimiter) ratelimiter.RateLimiter {
```

Nit: `RateLimiter` is a bit redundant here given the type.

pkg/reconciler/managed/reconciler.go
```diff
@@ -563,13 +550,14 @@ func (r *Reconciler) Reconcile(_ context.Context, req reconcile.Request) (reconc
 	if !meta.WasDeleted(managed) {
 		if err := r.managed.ResolveReferences(ctx, managed); err != nil {
 			// If any of our referenced resources are not yet ready (or if we
-			// encountered an error resolving them) we want to try again after a
-			// short wait. If this is the first time we encounter this situation
-			// we'll be requeued implicitly due to the status update.
-			log.Debug("Cannot resolve managed resource references", "error", err, "requeue-after", time.Now().Add(r.shortWait))
+			// encountered an error resolving them) we want to try again. If
+			// this is the first time we encounter this situation we'll be
+			// requeued implicitly due to the status update. // If not, we want
```
negz, line 555: Looks like a line wrap issue here.

negz (approved):

PR #244, opened 2021-02-25
add pointer helper functions for slices for referencers

smcavallo:
Signed-off-by: smcavallo <smcavallo@hotmail.com>

`FromPtrValue` and `ToPtrValue` are extremely convenient to use when using `Resolve` as part of referencers in providers
This PR adds `FromPtrValues` and `ToPtrValues`  for the same convenience when using `ResolveMultiple` - for example referencers with a slice of SubnetIDs or SecurityGroupIDs
ex.
```
	mrsp, err = r.ResolveMultiple(ctx, reference.MultiResolutionRequest{
		CurrentValues: FromPtrValues(mg.Spec.ForProvider.VPCConfig.SecurityGroupIDs)
```

I have:

* Unit test
* Run in a provider directly


pkg/reference/reference.go
```diff
@@ -56,6 +56,22 @@ func ToPtrValue(v string) *string {
 	return &v
 }
 
+// FromPtrValues adapts a slice of string pointer fields for use as CurrentValues.
+func FromPtrValues(v []*string) (res []string) {
```
hasheddan, line 60: Since length is known based on the passed slice we can pre-allocate the return slice before setting its values at each index. Would you mind updating to do so?

hasheddan (commented):
Thanks for this addition @smcavallo! One nitpick below.

pkg/reference/reference.go
muvaf: Wondering if we should return `""` or skip the element when the value is `nil`. I guess putting `""` would be more expected by the users of this function, compared to giving 10 elements and retrieving, say, 3.

hasheddan: Looks like we are now returning `""` which seems right to me because we want to error on that case in validation right? https://github.com/crossplane/crossplane-runtime/blob/8c9b0752894993fb456c2a438270c6d02dcf5b66/pkg/reference/reference.go#L159

muvaf (approved):

hasheddan (approved):

PR #246, opened 2021-02-25
Set namespace when looking up resource with externalName

### LINKED ISSUE #245: Namespaced resources with externalName fails to update
### What happened?

I'm using crossplane-runtime with a namespaced CR wanted to use the externalName feature.

### How can we reproduce it?

Use a namespaced CR and return [`ExternalNameAssigned = true`](https://github.com/crossplane/crossplane-runtime/blob/master/pkg/reconciler/managed/reconciler.go#L317) in `Create`. This results in the following error:

```
2021-02-24T16:59:21.926+0100    DEBUG   account_test    Cannot update managed resource  {"controller": "managed/account.example.org", "request": "default/test-acc-b5f78", "uid": "cb3d3cd8-90ca-4d61-a394-b04bc932e1fb", "version": "94", "external-name": "", "error": "Account.example.org \"test-acc-b5f78\" not found", "requeue-after": "2021-02-24T16:59:26.926+0100"}
```

The fix is quite simple, I will submit a PR shortly.

Crossplane-runtime version: v0.12.0 and also on master

ctrox:

Set namespace when looking up resource with externalName.

Fixes #245 

I have:

### DIFF
--- a/pkg/reconciler/managed/reconciler.go
+++ b/pkg/reconciler/managed/reconciler.go
@@ -687,7 +687,7 @@ func (r *Reconciler) Reconcile(_ context.Context, req reconcile.Request) (reconc
 			// of the loop. In that case, we warn the user that the external resource
 			// might be leaked.
 			err := retry.OnError(retry.DefaultRetry, resource.IsAPIError, func() error {
-				nn := types.NamespacedName{Name: managed.GetName()}
+				nn := types.NamespacedName{Name: managed.GetName(), Namespace: managed.GetNamespace()}
 				if err := r.client.Get(ctx, nn, managed); err != nil {
 					return err
 				}


pkg/reconciler/managed/reconciler.go
```diff
@@ -687,7 +687,7 @@ func (r *Reconciler) Reconcile(_ context.Context, req reconcile.Request) (reconc
 			// of the loop. In that case, we warn the user that the external resource
 			// might be leaked.
 			err := retry.OnError(retry.DefaultRetry, resource.IsAPIError, func() error {
-				nn := types.NamespacedName{Name: managed.GetName()}
+				nn := types.NamespacedName{Name: managed.GetName(), Namespace: managed.GetNamespace()}
```
hasheddan, line 690: Managed resources are intended to be cluster-scoped and the machinery here is intended to accommodate that fact. Using namespace-scoped managed resources could impact other aspects of the managed reconciler, such as cross-resource references. Do you have an example of where you are currently using the managed reconciler successfully for a namespaced resource?

ctrox: I was not aware of that :) I know that all included Crossplane providers only use cluster-scoped resources for their managed resources but so far nothing had stopped us using namespaced resources in our own provider. So far everything has worked but tbh we have not had to use cross-resource refs yet. If this is completely unsupported, I think there should be a more obvious way to tell us (docs or maybe even some check on controller start).

But of course my preference would be that this is *somewhat* supported so we can keep using it. I'm aware of all the other constructs that Crossplane provides to have namespaced resources (compositions, publications etc.) but sometimes that's just a bit too much configuration overhead for simple resources. So anyway, I don't have an example ready right now but there is not much to it besides setting `+kubebuilder:resource:scope=Namespaced`. I would prefer it if crossplane-runtime did not limit this use-case but maybe instead pointed at the limitations of it, like for example that cross-resource refs would just work if the other resource is cluster-scoped.

ctrox:
This turned out not to be a bug, I addressed the wider issue here: https://github.com/crossplane/crossplane-runtime/issues/251

PR #248, opened 2021-02-26
Fix panic in debug log in managed reconciler

hasheddan:
### Description of your changes

The debug logger panics with mismatched keys and values. One of the
requeue-after keys was not cleaned up when rate limiting was enabled.

Signed-off-by: hasheddan <georgedanielmangum@gmail.com>

Fixes #247

cc @enderv

I have:

- [x] Read and followed Crossplane's [contribution process].
- [x] Run `make reviewable test` to ensure this PR is ready for review.

### How has this code been tested

Exercised code path with debug logger enabled.

[contribution process]: https://git.io/fj2m9

Issue #247, opened 2021-02-26
Panic when error reconciling with debug turned


enderv:
### What happened?
Attempting to delete with debug logging on can cause a panic if there is an error deleting the resource

### How can we reproduce it?
I was using provider-aws and created a vpc

```
apiVersion: ec2.aws.crossplane.io/v1beta1
kind: VPC
metadata:
  name: crossplane-test-reconcile-vpc
spec:
  forProvider:
    region: us-east-2
    cidrBlock: 172.16.0.0/16
    enableDnsSupport: true
    enableDnsHostNames: true
  providerConfigRef:
    name: aws
```
Manually add a security group in the console and then attempt to delete the vpc resource. 

```
E0225 19:12:28.135300   79698 runtime.go:76] Observed a panic: odd number of arguments passed as key-value pairs for logging
goroutine 1825 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic(0x3aef020, 0xc0008814d0)
	/Users/cvodak/go/pkg/mod/k8s.io/apimachinery@v0.20.1/pkg/util/runtime/runtime.go:74 +0x134
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
	/Users/cvodak/go/pkg/mod/k8s.io/apimachinery@v0.20.1/pkg/util/runtime/runtime.go:48 +0xdf
panic(0x3aef020, 0xc0008814d0)
	/usr/local/Cellar/go/1.15.3/libexec/src/runtime/panic.go:975 +0x4c6
go.uber.org/zap/zapcore.(*CheckedEntry).Write(0xc000c9adc0, 0xc000e3dcc0, 0x1, 0x1)
	/Users/cvodak/go/pkg/mod/go.uber.org/zap@v1.15.0/zapcore/entry.go:230 +0x7aa
go.uber.org/zap.(*Logger).DPanic(0xc0016766c0, 0x41ab0d4, 0x3d, 0xc000e3dcc0, 0x1, 0x1)
	/Users/cvodak/go/pkg/mod/go.uber.org/zap@v1.15.0/logger.go:215 +0x7d
github.com/go-logr/zapr.handleFields(0xc0016766c0, 0xc001131830, 0x3, 0x3, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
	/Users/cvodak/go/pkg/mod/github.com/go-logr/zapr@v0.2.0/zapr.go:100 +0x6cb
[...]

PR #249, opened 2021-02-26
remove odd numbered key:value on debugs

enderv:
Signed-off-by: cvodak1 <cvodak@ea.com>

Fixes # https://github.com/crossplane/crossplane-runtime/issues/247

I have:

Unit tests run only

### DIFF
--- a/pkg/reconciler/managed/reconciler.go
+++ b/pkg/reconciler/managed/reconciler.go
@@ -600,7 +600,7 @@ func (r *Reconciler) Reconcile(_ context.Context, req reconcile.Request) (reconc
 				// this issue we'll be requeued implicitly when we update our
 				// status with the new error condition. If not, we want requeue
 				// explicitly, which will trigger backoff.
-				log.Debug("Cannot delete external resource", "error", err, "requeue-after")
+				log.Debug("Cannot delete external resource", "error", err)
 				record.Event(managed, event.Warning(reasonCannotDelete, err))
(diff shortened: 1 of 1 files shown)


hasheddan:
Ah thanks @enderv! Wasn't aware you were tackling this, I'll close #248 in favor of this one :)

hasheddan (approved):

negz:
@hasheddan @enderv What do you think about passing a debug logger to the unit tests to catch this?

hasheddan:
@negz I tried that out, but wasn't sure if we wanted actual debug logs in test output. That being said, it might actually be nice to see those and they would only show up on test failures I believe.

@enderv you should be able to add this by adding a `WithLogger(logging.NewLogrLogger(zap.New(zap.UseDevMode(true))))` reconciler option to the various tests in the table.

enderv:
@hasheddan  yeah sorry was feeling ambitious and they said it was just one line change, didn't know you saw it as well. 

trying the logger bit now

enderv:
@hasheddan @negz added. Verified locally that they fail by rebreaking the debug log.

hasheddan:
Hmm looks like this does dump the debug logs into stdout on successful runs too. Mind dropping the second commit for the time-being @enderv and we can evaluate how we want to handle this across all the projects?

enderv:
@hasheddan sounds good. removed it for now

PR #252, opened 2021-03-11
Use OpenAPI to set the default ProviderConfig

negz:

This allows us to avoid an extra update call to the API server when new managed resources are created.

I have:

I've tested this by building provider-gcp against this commit, and ensuring that `spec.providerConfigRef.name` is defaulted even when the provider is not running.

### DIFF
--- a/apis/common/v1/resource.go
+++ b/apis/common/v1/resource.go
@@ -139,6 +139,7 @@ type ResourceSpec struct {
 	// ProviderConfigReference specifies how the provider that will be used to
 	// create, observe, update, and delete this managed resource should be
 	// configured.
+	// +kubebuilder:default={"name": "default"}
 	ProviderConfigReference *Reference `json:"providerConfigRef,omitempty"`
 
 	// ProviderReference specifies the provider that will be used to create,
--- a/pkg/reconciler/managed/reconciler.go
+++ b/pkg/reconciler/managed/reconciler.go
@@ -367,11 +367,8 @@ func defaultMRManaged(m manager.Manager) mrManaged {
 	return mrManaged{
 		ConnectionPublisher: NewAPISecretPublisher(m.GetClient(), m.GetScheme()),
 		Finalizer:           resource.NewAPIFinalizer(m.GetClient(), managedFinalizerName),
-		Initializer: InitializerChain{
-			NewDefaultProviderConfig(m.GetClient()),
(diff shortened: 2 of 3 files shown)


pkg/reconciler/managed/reconciler.go
```diff
@@ -367,11 +367,8 @@ func defaultMRManaged(m manager.Manager) mrManaged {
 	return mrManaged{
 		ConnectionPublisher: NewAPISecretPublisher(m.GetClient(), m.GetScheme()),
 		Finalizer:           resource.NewAPIFinalizer(m.GetClient(), managedFinalizerName),
-		Initializer: InitializerChain{
-			NewDefaultProviderConfig(m.GetClient()),
```
negz, line 371: I considered leaving this out of an abundance of caution, because:

* It's a no-op when the ProviderConfigReference is already set.
* Many providers assume the ProviderConfigReference will never be nil
* It's possible that a provider will be built against this library, but forget to generate/apply new CRDs.

I decided to just remove it, given that it's very unlikely anyone using the package manager and our standard build tooling would ever end up accidentally deploying new provider code without also updating their CRDs.

apis/common/v1/resource.go
```diff
@@ -139,6 +139,7 @@ type ResourceSpec struct {
 	// ProviderConfigReference specifies how the provider that will be used to
 	// create, observe, update, and delete this managed resource should be
 	// configured.
+	// +kubebuilder:default={"name": "default"}
```
muvaf, line 142: I didn't know you could give an object here, nice.

muvaf (approved):
LGTM! Note that many managed resources that opt out of `NewNameAsExternalName` manually set `NewDefaultProviderConfig` as initializer. Though I don't think it's necessary to address that now.

hasheddan (approved):

negz:
> Note that many managed resources that opt out of NewNameAsExternalName manually set NewDefaultProviderConfig as initializer. Though I don't think it's necessary to address that now.

I agree. I'm pretty sure for those managed resources the NewDefaultProviderConfig logic will just become a no-op because `spec.providerConfigRef.name` will never be unset, so I think we should just surface this deprecation in the release notes then let providers update at their leisure. I think our linter config complains when folks try to use a function with a `// Deprecated` comment, which will help bug folks to update.

PR #259, opened 2021-03-28
Unpublish connection details and remove finalizer without attempting to connect to external client when deletion policy is Orphan

### LINKED ISSUE #256: Allow deletion of managed resources with orphan policy when cannot connect to provider

The managed reconciler will error [early in the reconcile loop](https://github.com/crossplane/crossplane-runtime/blob/fe13ff9a1c2173dfb9d8db7dd4bbada7a33626d6/pkg/reconciler/managed/reconciler.go#L561) if we are unable to connect to the external provider using the specified `ProviderConfig`. This means situations like https://github.com/crossplane/crossplane/issues/2212 can occur where we don't need to take any external action, but we still cannot clean up the resources in the cluster because the provider API (in this case an Azure SQL database) no longer exists.

The managed reconciler should check if `meta.WasDeleted()` and `deletionPolicy: Orphan` upon failing to connect to the external provider. If so, it should not call delete on the external client. This will require some refactoring of the `Connect()` methods of most providers as they typically will return a `nil` external client if there is an error connecting. One option to accommodate clients that don't want to support this behavior is to just not update their `Connect()` methods and maintain the existing behavior for providers where the external client is `nil` on `err != nil`. This would allow providers such as `provider-sql` to support this, but others that wouldn't encounter this issue as frequently (such as all the cloud providers), to not.

Note that this should not take the place of https://github.com/crossplane/crossplane/issues/1782, but could supplement the behavior.

hasheddan:

This PR adds the following functionality to the managed reconciler (see individual commits for more detail):
- Modifies managed reconciler to skip attempting to connect to external
client as well as initialize and resolve references when managed
resource has deletion timestamp and deletion policy is orphan.
- Updates the status of managed resource to be set to deleting in the case
where an external resource does not exist, but we fail to unpublish
connection or remove finalizer.
- Now that the managed reconciler will handle deletion with policy Orphan
early, there is no need to check that deletion policy is not orphan when
handling a managed resource with deletion timestamp set and external
resource exists.

Fixes #256 

I have:

Unit tests.

### DIFF
--- a/pkg/reconciler/managed/reconciler.go
+++ b/pkg/reconciler/managed/reconciler.go
@@ -629,7 +629,9 @@ func (r *Reconciler) Reconcile(_ context.Context, req reconcile.Request) (reconc
 		log = log.WithValues("deletion-timestamp", managed.GetDeletionTimestamp())
 		managed.SetConditions(xpv1.Deleting())
 
-		if observation.ResourceExists && managed.GetDeletionPolicy() != xpv1.DeletionOrphan {
+		// We'll only reach this point if deletion policy is not orphan, so we
+		// are safe to call external deletion if external resource exists.
+		if observation.ResourceExists {
 			if err := external.Delete(externalCtx, managed); err != nil {
 				// We'll hit this condition if we can't delete our external
 				// resource, for example if our provider credentials don't have
--- a/pkg/reconciler/managed/reconciler.go
+++ b/pkg/reconciler/managed/reconciler.go
@@ -525,6 +525,46 @@ func (r *Reconciler) Reconcile(_ context.Context, req reconcile.Request) (reconc
 		"external-name", meta.GetExternalName(managed),
 	)
 
+	// If managed resource has a deletion timestamp and and a deletion policy of
+	// Orphan, we do not need to observe the external resource before attempting
+	// to unpublish connection details and remove finalizer.
+	if meta.WasDeleted(managed) && managed.GetDeletionPolicy() == xpv1.DeletionOrphan {
+		log = log.WithValues("deletion-timestamp", managed.GetDeletionTimestamp())
+		managed.SetConditions(xpv1.Deleting())
+
+		// Empty ConnectionDetails are passed to UnpublishConnection because we
+		// have not retrieved them from the external resource. In practice we
+		// currently only write connection details to a Secret, and we rely on
+		// garbage collection to delete the entire secret, regardless of the
+		// supplied connection details.
+		if err := r.managed.UnpublishConnection(ctx, managed, ConnectionDetails{}); err != nil {
+			// If this is the first time we encounter this issue we'll be
+			// requeued implicitly when we update our status with the new error
+			// condition. If not, we requeue explicitly, which will trigger
+			// backoff.
+			log.Debug("Cannot unpublish connection details", "error", err)
+			record.Event(managed, event.Warning(reasonCannotUnpublish, err))
+			managed.SetConditions(xpv1.ReconcileError(err))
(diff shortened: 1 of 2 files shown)


pkg/reconciler/managed/reconciler.go
```diff
@@ -525,6 +525,41 @@ func (r *Reconciler) Reconcile(_ context.Context, req reconcile.Request) (reconc
...
 	)
 
+	// If managed resource has a deletion timestamp and and a deletion policy of
+	// Orphan, we do not need to observe the external resource before attempting
+	// to unpublish connection details and remove finalizer.
+	if meta.WasDeleted(managed) && managed.GetDeletionPolicy() == xpv1.DeletionOrphan {
+		log = log.WithValues("deletion-timestamp", managed.GetDeletionTimestamp())
+		managed.SetConditions(xpv1.Deleting())
+
+		if err := r.managed.UnpublishConnection(ctx, managed, ConnectionDetails{}); err != nil {
```
hasheddan, line 535: Passing empty `ConnectionDetails{}` here is a little strange, but the only "unpublish" implementation we have right now is a no-op and allows secrets to be handled with GC. However, it is unclear whether we would ever really need to pass connection details to an unpublish connection, as the contents of the secret are not really important to unpublishing the entire secret. I presume the reason why the connection publisher interface accepts this argument is to allow an implementation that partially unpublishes connection details or verifies that details match before unpublishing. We could consider just removing the `ConnectionDetails` argument from `UnpublishConnection`, but that would technically be a breaking API change (though I doubt any providers would be significantly impacted if at all).

negz: The intent of the `UnpublishConnection` interface is supposed to be "remove this subset of connection details", without any opinion as to where those details are being unpublished from (e.g. a secret). Roughly the inverse of `PublishConnection`.

> Passing empty ConnectionDetails{} here is a little strange, but the only "unpublish" implementation we have right now is a no-op and allows secrets to be handled with GC.

This seems like a great thing to include in a comment on this call.

hasheddan: @negz done :+1:

pkg/reconciler/managed/reconciler.go
```diff
@@ -525,6 +525,43 @@ func (r *Reconciler) Reconcile(_ context.Context, req reconcile.Request) (reconc
...
 
+	// If managed resource has a deletion timestamp and and a deletion policy of
+	// Orphan, we do not need to observe the external resource before attempting
+	// to unpublish connection details and remove finalizer.
+	if meta.WasDeleted(managed) && managed.GetDeletionPolicy() == xpv1.DeletionOrphan {
+		log = log.WithValues("deletion-timestamp", managed.GetDeletionTimestamp())
+		managed.SetConditions(xpv1.Deleting())
+
+		// Empty ConnectionDetails are passed to UnpublishConnection because we
+		// have not retrieved them from the external resource.
```
negz, line 536:
```suggestion
		// Empty ConnectionDetails are passed to UnpublishConnection because we
		// have not retrieved them from the external resource. In practice we currently
		// only write connection details to a Secret, and we rely on garbage collection
		// to delete the entire secret, regardless of the supplied connection details.
```

negz (approved):
LGTM, though I'd like to include a little more detail in the comment.

An aside - this `Reconcile` method has a cyclomatic complexity of 26 in master at the moment. I think this PR will bring it up to 27 or 28. I want to voice that I'm concerned about the slowly increasing branchiness of this core reconciler as we account for more edge conditions like this. No idea how (or whether we even should) address this. It's definitely not something we need to address in this PR, but it does come to mind every time I review "new functionality" in this reconciler.

muvaf (commented):
Sorry, seems I forgot my review in pending state. Submitting it just in case you find it useful.

PR #264, opened 2021-04-28
Add Retry Logic for APISecretPublisher

### LINKED ISSUE #263: Add Retry Logic for APISecretPublisher

As discussed [here](https://github.com/crossplane/provider-gcp/issues/298), if a managed resource publishes a connection secret via the `APISecretPublisher` and if we cannot persist it in etcd at the time of external resource provisioning, we may leak the external resource. For example, for Google Cloud API iam.ServiceAccountKey resources, private keys provisioned by GCP are not stored in GCP, and thus [are only available as the result of a `keys.create` call](https://cloud.google.com/iam/docs/reference/rest/v1/projects.serviceAccounts.keys). If because of a temporary failure the connection secret is not persisted and the external name annotation is not set, then we have a leaking Cloud API resource.

For certain external resources, we might be able to associate metadata with the external resource for registering ownership to crossplane. In this case, eventually we can clean up leaking resources. However, for instance, Google Cloud API's `ServiceAccountKey` seems not to allow adding such metadata. 

For such resource, there seems no perfect solution as we cannot put provisioning of the external resource and provisioning of the connection secret into the same transactional boundary, but for now, we may considering adding a retry logic to `APISecretPublisher.PublishConnection` for transient failures to decrease the probability of leaking certain external resources.

ulucinar:

This PR adds `resource.ApplicatorWithRetry` that can retry application of a nested `resource.Applicator` on designated errors with the designated schedule. And it further uses a `resource.ApplicatorWithRetry` to introduce retries during connection secret publication.

Fixes #263 

I have:

- I have manually tested this change in crossplane-runtime together with a custom build of provider-gcp & `v1alpha1.ServiceAccountKey`, which published a connection secret.
- Unit tests are also introduced with this PR.

### DIFF
--- a/pkg/resource/resource.go
+++ b/pkg/resource/resource.go
@@ -27,6 +27,8 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
@@ -212,6 +214,11 @@ func IsAPIError(err error) bool {
 	return ok
 }
 
+// IsAPIErrorWrapped returns true if err is a K8s API error, or recursively wraps a K8s API error
+func IsAPIErrorWrapped(err error) bool {
+	return IsAPIError(errors.Cause(err))
+}
+
 // IsConditionTrue returns if condition status is true
 func IsConditionTrue(c xpv1.Condition) bool {
 	return c.Status == corev1.ConditionTrue
@@ -220,6 +227,39 @@ func IsConditionTrue(c xpv1.Condition) bool {
 // An Applicator applies changes to an object.
 type Applicator interface {
 	Apply(context.Context, client.Object, ...ApplyOption) error
+}
+
+type shouldRetryFunc func(error) bool
+
+// An ApplicatorWithRetry applies changes to an object, retrying on transient failures
+type ApplicatorWithRetry struct {
+	Applicator
+	shouldRetry shouldRetryFunc
+	backoff     wait.Backoff
+}
+
+// Apply invokes nested Applicator's Apply retrying on designated errors
+func (awr *ApplicatorWithRetry) Apply(ctx context.Context, c client.Object, opts ...ApplyOption) error {
+	return retry.OnError(awr.backoff, awr.shouldRetry, func() error {
+		return awr.Applicator.Apply(ctx, c, opts...)
+	})
+}
+
+// NewApplicatorWithRetry returns an ApplicatorWithRetry for the specified
+// applicator and with the specified retry function.
+//   If backoff is nil, then retry.DefaultRetry is used as the default.
+func NewApplicatorWithRetry(applicator Applicator, shouldRetry shouldRetryFunc, backoff *wait.Backoff) *ApplicatorWithRetry {
+	result := &ApplicatorWithRetry{
+		Applicator:  applicator,
+		shouldRetry: shouldRetry,
+		backoff:     retry.DefaultRetry,
+	}
+
+	if backoff != nil {
+		result.backoff = *backoff
+	}
+
+	return result
 }
 
 // A ClientApplicator may be used to build a single 'client' that satisfies both
--- a/pkg/resource/resource_test.go
+++ b/pkg/resource/resource_test.go
@@ -22,12 +22,17 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/util/retry"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 	"github.com/crossplane/crossplane-runtime/pkg/resource/fake"
@@ -38,13 +43,19 @@ const (
 	namespace = "coolns"
 	name      = "cool"
 	uid       = types.UID("definitely-a-uuid")
+	testSteps = 3
 )
 
-var MockOwnerGVK = schema.GroupVersionKind{
-	Group:   "cool",
-	Version: "large",
-	Kind:    "MockOwner",
-}
+var (
+	MockOwnerGVK = schema.GroupVersionKind{
+		Group:   "cool",
+		Version: "large",
+		Kind:    "MockOwner",
+	}
+
+	testBackoff = wait.Backoff{}
+	errTest     = errors.New("test-error")
+)
 
 func TestLocalConnectionSecretFor(t *testing.T) {
 	secretName := "coolsecret"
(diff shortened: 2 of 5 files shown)


pkg/resource/resource.go
```diff
@@ -212,6 +214,11 @@ func IsAPIError(err error) bool {
 	return ok
 }
 
+// IsAPIErrorWrapped returns true if err is a K8s API error, or recursively wraps a K8s API error
+func IsAPIErrorWrapped(err error) bool {
```
ulucinar, line 218: I do not want to change the behavior of existing resource.IsAPIError because I don't know all the contexts it's used.

pkg/resource/resource.go
```diff
@@ -222,6 +229,39 @@ type Applicator interface {
...
+// An ApplicatorWithRetry applies changes to an object, retrying on transient failures
+type ApplicatorWithRetry struct {
+	Applicator
+	shouldRetry shouldRetryFunc
+	backoff     wait.Backoff
+}
+
+// Apply invokes nested Applicator's Apply retrying on designated errors
+func (awr *ApplicatorWithRetry) Apply(ctx context.Context, c client.Object, opts ...ApplyOption) error {
+	return retry.OnError(awr.backoff, awr.shouldRetry, func() error {
```
hasheddan, line 243: @ulucinar just double checking my understanding here: This will retry according to policy specified as long as we are running into k8s API errors, but if we continue to be unable to publish, we will eventually fail on the underlying `Apply` due to context deadline exceeded, at which point we will exit reconciliation. Am I correct on that?

ulucinar: Hi @hasheddan,
Yes, correct. If retries do not help, what's returned to the managed reconciler is the underlying error, which could be context deadline exceeded or even a K8s API error because we retry with a limit as specified by the back-off configuration.

hasheddan: @ulucinar thanks! I wasn't sure exactly how retry limit was being handled, but I took a look at the underlying [exponential backoff implementation](https://github.com/kubernetes/apimachinery/blob/8daf28983e6ecf28bd8271925ee135c1179ad13a/pkg/util/wait/wait.go#L402) and see that it breaks on `steps == 1`, with steps being decremented on every iteration and a default of [4 steps](https://github.com/kubernetes/client-go/blob/8c8fa70f7a2acc191d4787327639621f69992efd/util/retry/util.go#L39) :+1:

ulucinar:
In codecov check here:
https://github.com/crossplane/crossplane-runtime/pull/264/checks?check_run_id=2458129341

there seems to be [an unexpected coverage change](https://docs.codecov.io/docs/unexpected-coverage-changes). According to the codecov report, it seems like we have increased code coverage in both of the two (non-ignored) files modified:
https://codecov.io/gh/crossplane/crossplane-runtime/commit/b5db8c1c9e354f0f399e09d6a608d093846e514b/

But codecov says that the overall coverage has dropped 1.3%. And I did not delete tests. Can anyone with experience on codecov bring an explanation? 

At least, I may add some new tests to increase coverage over already existing code to get rid of this, correct?

hasheddan:
> At least, I may add some new tests to increase coverage over already existing code to get rid of this, correct?

@ulucinar that would be fine, but you can also ignore for now if you like. Codecov is notoriously flaky, which is why we don't require it as a check :)

ulucinar:
@hasheddan, thank you for the insight. If there are some easy-to-cover lines, I will try.

ulucinar:
> > At least, I may add some new tests to increase coverage over already existing code to get rid of this, correct?
> 
> @ulucinar that would be fine, but you can also ignore for now if you like. Codecov is notoriously flaky, which is why we don't require it as a check :)

Hi @hasheddan,
I have added some small tests, let's see if things improve on the codecov side. If not, I also think we should move on as you suggested because even the "coverage changes" tab of codecov actually shows an increase in coverage:
https://app.codecov.io/gh/crossplane/crossplane-runtime/compare/264/changes

Thank you @hasheddan!

pkg/resource/resource.go
```diff
@@ -222,6 +229,40 @@ type Applicator interface {
...
+		return awr.Applicator.Apply(ctx, c, opts...)
+	})
+}
+
+// NewApplicatorWithRetry returns an ApplicatorWithRetry for the specified
+// applicator and with the specified retry function.
+//   If backoff is nil, then retry.DefaultRetry is used as the default.
+func NewApplicatorWithRetry(applicator Applicator, shouldRetry shouldRetryFunc, backoff *wait.Backoff) Applicator {
```
muvaf, line 251:
```suggestion
func NewApplicatorWithRetry(applicator Applicator, shouldRetry shouldRetryFunc, backoff *wait.Backoff) ApplicatorWithRetry {
```

Since the interface is programatically enforced already, I usually return just the struct itself so that if at some point a modification on the struct is needed, users can do so. If they require only the interface, which is the case here, then that'd work as well.

pkg/resource/resource.go
```diff
@@ -222,6 +229,40 @@ type Applicator interface {
...
+	result := &ApplicatorWithRetry{
+		Applicator:  applicator,
+		shouldRetry: shouldRetry,
+	}
+
+	if backoff != nil {
+		result.backoff = *backoff
+	} else {
+		result.backoff = retry.DefaultRetry
+	}
```
muvaf, line 261:
```suggestion
	result := &ApplicatorWithRetry{
		Applicator:  applicator,
		shouldRetry: shouldRetry,
		backoff: retry.DefaultRetry
	}

	if backoff != nil {
		result.backoff = *backoff
	}
```

nitpick

pkg/resource/resource.go
```diff
@@ -212,6 +214,11 @@ func IsAPIError(err error) bool {
 	return ok
 }
 
+// IsAPIErrorWrapped returns true if err is a K8s API error, or recursively wraps a K8s API error
```
muvaf, line 217: From my reading of the source code, it doesn't seem like we'd get more than one level below, i.e. if it's wrapped two times, we won't actually see the first error because `Wrap` treats the incoming error as cause, not the cause of that incoming error.

If that's the case, I wonder if we should make this APIError check closer to where the error is returned without relying on whether the wrapping is done only once.

pkg/resource/resource.go
muvaf:
```suggestion
func NewApplicatorWithRetry(applicator Applicator, shouldRetry shouldRetryFunc, backoff wait.Backoff) Applicator {
```

nitpick: I'd feel fine about requiring a backoff to be supplied since that's the interface of the underlying `retry.OnError`

ulucinar: Hi @muvaf,
Thank you for the comments.

>  If they require only the interface, which is the case here, then that'd work as well.
Yes, I would like to hide the `resource.applicatorWithRetry` and only expose a constructor method with it. That's why the signature uses the interface type. I have also un-exported the type itself.

ulucinar: > nitpick: I'd feel fine about requiring a backoff to be supplied since that's the interface of the underlying retry.OnError

I agree. We could get rid of providing a default for the back-off configuration. Why I preferred to use a default value is as follows:
The `retry` package provides two defaults back-off configurations with different semantics. The first one (`retyr.DefaultRetry`) is suggested to be used in contexts where there are competing modifiers for a resource (i.e., during resolution of HTTP-409 errors). The other one (`retry.DefaultBackoff`) is suggested to be used in contexts with unrelated changes. Because I expect, in general, a `resource.APIPatchingApplicator` to be wrapped by the `applicatorWithRetry`, I preferred to have the clients of `applicatorWithRetry` use `retry.DefaultRetry` as the default. 

But I don't have a strong opinion here, we can also remove a default value.

pkg/resource/resource.go
ulucinar: `resource.IsAPIErrorWrapped` uses `errors.Cause` which recursively looks for the first non-causer error. So, even if the original error is wrapped more than once, we are safe to get the original non-causer error.

pkg/resource/resource.go
muvaf: > Yes, I would like to hide the resource.applicatorWithRetry and only expose a constructor method with it.

Why though? I'm thinking of the case where a consumer of `NewApplicatorWithRetry` gets an `Applicator`. If they'd like to change configuration after initialization or make use of new features we introduce to that struct that's not covered by the interface, they have to do a type casting. But if we return exported `ApplicatorWithRetry`, interface would still be satisfied and they'd get to work with the real object.

pkg/resource/resource.go
muvaf: Got it, seems I misinterpreted the source code. Thanks for the explanation!

pkg/resource/resource.go
muvaf: Regarding the default backoff, I think user supplying `nil` and getting a default could be a bit misleading. In this context, `nil` is not a valid option so they might figure that out easily. But for defaulting, we usually accept [options](https://github.com/crossplane/crossplane-runtime/blob/f2418dc/pkg/reconciler/managed/reconciler.go#L386) for configuration and let the default overridden by the options if the user wants to. So it'd be consistent to expose options here as well if you'd like to have a default, but if you don't want to (since it's a single config parameter), requiring it to be supplied wouldn't be much of a hassle for users anyway IMO.

muvaf (approved):
I have some small comments but overall LGTM! Thanks @ulucinar !

pkg/resource/resource.go
ulucinar: @muvaf believes that we may utilize the applicatorWithRetry type in core crossplane repos. So he proposes to expose it. 
Agreed.

muvaf:
@ulucinar could you squash the commits before I merge?

ulucinar:
> @ulucinar could you squash the commits before I merge?

Done @muvaf. Thanks.

pkg/resource/resource_test.go
```diff
@@ -638,3 +649,241 @@ func TestGetExternalTags(t *testing.T) {
 		})
 	}
 }
+
+// single test case => not using tables
+func Test_errNotControllable_NotControllable(t *testing.T) {
```
negz, line 654: Very late commentary here, but in future I'd like to avoid these `_` separated long test names. I prefer shorter less descriptive names that include a `reason` string that is printed if the test fails with more context.

ulucinar: Hi @negz,
Would you like me to open a PR changing these test names?

negz: Yes please! It's not urgent at all, but I would prefer to avoid the convention creeping in. :)

PR #266, opened 2021-05-30
Implement a LateInitializer struct for use by providers

muvaf:

`LateInitializer` will help providers figure out whether a change has been made and spec update is needed after the late initialization without using `DeepCopy` of the original and changed objects.

I wanted to include only the bare minimum I need for generated ACK resources where the source and target field types are the same to be conservative about what we expose. We can add more as needed.

Example usage:
```go
func lateInitialize(cr *svcapitypes.Table, resp *svcsdk.DescribeTableOutput) bool {
  li := awsclients.NewLateInitializer()
  cr.Spec.ForProvider.Property = li.LateInitializeStringPtr(cr.Spec.ForProvider.Property, resp.Table.Property)
  // .. many calls using the same li object ...
  return li.IsChanged()
}
```

I have:

Unit tests.

### DIFF
--- a/pkg/resource/late_initializer.go
+++ b/pkg/resource/late_initializer.go
@@ -41,12 +41,18 @@ func (li *LateInitializer) IsChanged() bool {
 	return li.changed
 }
 
+// SetChanged marks the LateInitializer such that users can tell whether any
+// of the late initialization calls returned the non-original argument.
+func (li *LateInitializer) SetChanged() {
+	li.changed = true
+}
+
 // LateInitializeStringPtr implements late initialization for *string.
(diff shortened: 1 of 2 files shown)


hasheddan (approved):
@muvaf this looks good to me! One thing I was considering is whether we should add a `SetChanged()` method in case someone is using this with an unsupported type, but don't want to track if any changes happen in a separate variable, any thoughts on that?

muvaf:
@hasheddan sounds good. I started with having a lot of `LateInitialize*` functions and then reduced to these few, thinking providers could extend it if they want non-conventional late initialize functions but missed the fact that they can't access `changed` field. Thanks!

PR #270, opened 2021-07-08
Support merging in addition to replacing objects in patches

ulucinar:

Please refer to https://github.com/crossplane/crossplane/issues/2211 for more background on this PR.
This PR aims to add merge support to the `fieldpath` package and moves some code from the `crossplane` repo to the `crossplane-runtime` repo to implement common functionality & prevent code duplication. For merging objects (`map[string]interface{}`), the [mergo](https://github.com/imdario/mergo) library is used. Certain aspects of the behavior of the `mergo` library can be controlled via the new `fieldpath.MergeOptions` struct. Merging of a value at a given path of the desired `runtime.Object` into the actual (current) `runtime.Object` is achieved via a `resource.ApplyOption` that's passed as an argument to `resource.Applicator.Apply`. This new kind of `ApplyOption` can be obtained using the new `resource.WithMergeOptions` constructor. `crossplane` repo PR https://github.com/crossplane/crossplane/pull/2409 depends on this PR.

I have:

- [x] Added automated tests

Together with https://github.com/crossplane/crossplane/pull/2409, I have manually tested this PR using the following manifests. I have kubectl edited resources to test & observe the behavior (e.g., when `keepMapValues` is `false` or is missing in the composition for the `metadata` patch, observe that claim's annotations overwrite composed's annotations with the same keys. However, when `keepMapValues` is true, existing values in the composed annotations are preserved even if the claim (and correspondingly the XR) has a different value for the same key):
```yaml
---
kind: CompositeResourceDefinition
metadata:
  name: compositepostgresqlinstances.database.example.org
spec:
  group: database.example.org
  names:
    kind: CompositePostgreSQLInstance
    plural: compositepostgresqlinstances
  claimNames:
    kind: PostgreSQLInstance
    plural: postgresqlinstances
  connectionSecretKeys:
    - username
    - password
    - endpoint
    - port
  versions:
  - name: v1alpha1
    served: true
    referenceable: true
    schema:
      openAPIV3Schema:
        type: object
        properties:
          spec:
            type: object
            properties:
              parameters:
                type: object
                properties:
                  vpcSecurityGroupIDSelector:
                    description: VPCSecurityGroupIDSelector selects references to VPCSecurityGroups used to set the VPCSecurityGroupIDs.
                    properties:
                      matchControllerRef:
                        description: MatchControllerRef ensures an object with the same controller reference as the selecting object is selected.
                        type: boolean
                      matchLabels:
                        additionalProperties:
                          type: string
                        description: MatchLabels ensures an object with matching labels is selected.
                        type: object
                    type: object
                  storageGB:
                    type: integer
                required:
                  - storageGB
            required:
              - parameters
---
kind: Composition
metadata:
  name: composition-new
  labels:
    provider: aws
    guide: quickstart
    vpc: default
spec:
  writeConnectionSecretsToNamespace: crossplane-system
  compositeTypeRef:
    apiVersion: database.example.org/v1alpha1
    kind: CompositePostgreSQLInstance
  patchSets:
  - name: selector
    patches:
    - fromFieldPath: spec.parameters.vpcSecurityGroupIDSelector
      toFieldPath: spec.forProvider.vpcSecurityGroupIDSelector
      policy:
        mergeOptions:
          appendSlice: true
          keepMapValues: true
  - name: metadata
    patches:
    - fromFieldPath: metadata.annotations
      policy:
        mergeOptions:
          keepMapValues: true
      type: FromCompositeFieldPath
  resources:
    - name: rdsinstance
      base:
        apiVersion: database.aws.crossplane.io/v1beta1
        kind: RDSInstance
        spec:
          forProvider:
            region: us-east-1
            dbInstanceClass: db.t2.small
            masterUsername: masteruser
            engine: postgres
            engineVersion: "9.6"
            skipFinalSnapshotBeforeDeletion: true
            publiclyAccessible: true
            allocatedStorage: 20
      patches:
         - type: PatchSet
           patchSetName: selector
         - type: PatchSet
           patchSetName: metadata
         - fromFieldPath: "spec.writeConnectionSecretToRef.namespace"
           toFieldPath: "spec.writeConnectionSecretToRef.namespace"
         - fromFieldPath: "metadata.uid"
           toFieldPath: "spec.writeConnectionSecretToRef.name"
           transforms:
             - type: string
               string:
                 fmt: "%s-postgresql"
         - fromFieldPath: "spec.parameters.storageGB"
           toFieldPath: "spec.forProvider.allocatedStorage"
      connectionDetails:
        - fromConnectionSecretKey: username
        - fromConnectionSecretKey: password
        - fromConnectionSecretKey: endpoint
        - fromConnectionSecretKey: port
---
apiVersion: database.example.org/v1alpha1
kind: PostgreSQLInstance
metadata:
  name: my-db
  namespace: default
  annotations:
    a: a-from-claim
    b: b-from-claim
spec:
  parameters:
    vpcSecurityGroupIDSelector:
      matchLabels:
        l1: l1-from-claim
        l2: l2-from-claim
    storageGB: 20
  compositionSelector:
    matchLabels:
      provider: aws
      vpc: default
```


pkg/fieldpath/merge.go
```diff
@@ -0,0 +1,156 @@
...
+	errInvalidMerge = "failed to merge values"
+)
+
+// MergeOptions Specifies merge options on a field path
+type MergeOptions struct { // TODO(aru): add more options that control merging behavior
+	// Specifies that already existing values in a merged map should be preserved
+	KeepMapValues bool `json:"keepMapValues,omitempty"`
+	// Specifies that already existing elements in a merged slice should be preserved
+	AppendSlice bool `json:"appendSlice,omitempty"`
+}
```
negz, line 37: Is the intent that this type would be embedded in API types, as in https://github.com/crossplane/crossplane/pull/2409/files#diff-40b5c705a7ba08a514b14b156d26120fcd0c6a8b4326096ceecbb7d3bd971f29R283?

If so, I would prefer it to be defined along with all other API types in https://github.com/crossplane/crossplane-runtime/tree/master/apis/common/v1. I'd also suggest we make sure it complies with API conventions. Specifically:

* Are we sure boolean fields are the best option here? API conventions [warns to think twice about them](https://github.com/kubernetes/community/blob/0d0129/contributors/devel/sig-architecture/api-conventions.md#primitive-types).
* Are these fields optional? If so they should be marked `// +optional` and be `*bool`.

pkg/fieldpath/merge.go
```diff
@@ -0,0 +1,156 @@
...
+	if u, ok := o.(interface{ UnstructuredContent() map[string]interface{} }); ok {
+		return Pave(u.UnstructuredContent()), false, nil
+	}
+
+	oMap, err := runtime.DefaultUnstructuredConverter.ToUnstructured(o)
+	if err != nil {
+		return nil, false, err
+	}
+	return Pave(oMap), true, nil
+}
```
negz, line 70: Is `merge.go` the right place for this function? It does not seem specific to merging.

pkg/resource/api.go
```diff
@@ -152,6 +153,15 @@ func (a *APIPatchingApplicator) Apply(ctx context.Context, o client.Object, ao .
 	return errors.Wrap(a.client.Patch(ctx, o, &patch{desired}), "cannot patch object")
 }
 
+// WithMergeOptions returns an ApplyOption for merging the value at the given
+// fieldPath of desired object onto the current object with
+// the given merge options.
```
negz, line 158: I wonder whether we should consider moving to server side apply rather than building our own merge implementation client side?

https://kubernetes.io/docs/reference/using-api/server-side-apply/#merge-strategy

pkg/fieldpath/merge_test.go
```diff
@@ -0,0 +1,504 @@
...
+import (
+	"fmt"
+	"reflect"
+	"runtime"
+	"sort"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/imdario/mergo"
+	v1 "k8s.io/api/core/v1"
```
negz, line 28: Nit: We typically import this as corev1

pkg/fieldpath/merge.go
```diff
@@ -0,0 +1,156 @@
...
+	// replace desired object's value at fieldPath with
+	// the computed (merged) current value at the same path
+	return MergePath(fieldPath, desired, copyCurrent, nil)
+}
+
+// merges the given src onto the given dst.
+// If a nil merge options is supplied, the default behavior is MergeOptions'
+// default behavior. If dst or src is nil, src is returned
+// (i.e., dst replaced by src).
+func merge(dst, src interface{}, mergeOptions *MergeOptions) (interface{}, error) {
```
negz, line 141: Does this function require that `dst` and `src` are of the same underlying type?

pkg/fieldpath/merge.go
```diff
@@ -0,0 +1,156 @@
...
+		return err
+	}
+
+	return PatchFieldValueToObject(fieldPath, val, dst, mergeOptions)
+}
+
+// MergeReplace merges the value at fieldPath from desired into
+// a copy of current and then replaces the value at fieldPath of
+// desired with the merged value. current object is not modified.
+func MergeReplace(fieldPath string, current, desired runtime.Object, mergeOptions *MergeOptions) error {
```
negz, line 127: This API is a little unintuitive to me - possibly just due to the `current` and `desired` argument names. "Desired" reads more like the input (e.g. "I desire it to be this way") rather than something I would expect to be mutated.

pkg/fieldpath/merge.go
```diff
@@ -0,0 +1,156 @@
...
+
+	if copied {
+		return runtime.DefaultUnstructuredConverter.FromUnstructured(paved.object, to)
+	}
+	return nil
+}
+
+// MergePath merges the value at the given field path of the src object into
+// the dst object.
+func MergePath(fieldPath string, dst, src runtime.Object, mergeOptions *MergeOptions) error {
```
negz, line 106: I'm not sure how I feel about making this package deal more directly with `runtime.Object` - i.e. taking them as an argument. So far the established pattern has been to convert a `runtime.Object` (or anything else) to a `*Paved`, then operate on that. What do you think about implementing `MergePath` as a method on `*Paved` similar to the current `SetValue`? For example:

```go
func (p *Paved) MergeValue(path string, value interface{}, o ...MergeOption) error
```

pkg/resource/api.go
ulucinar: I will take a look.

pkg/fieldpath/object/object.go
```diff
@@ -0,0 +1,94 @@
...
+
+	v1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
+	"github.com/crossplane/crossplane-runtime/pkg/fieldpath"
+)
+
+// ToPaved tries to convert a runtime.Object into a *Paved via
+// runtime.DefaultUnstructuredConverter if needed. Returns the paved if
+// the conversion is successful along with whether the
+// runtime.DefaultUnstructuredConverter has been employed during the
+// conversion.
```
negz, line 30: I see that when you call this function you name the `bool` return value `copied`. It would be useful to include this information in the docstring. From my reader's perspective it is not obvious what the implications of `runtime.DefaultUnstructuredConverter` being employed is - I can only infer from the function below that it means the returned value is a copy when it would otherwise not be.

negz: Also, note that there is already a `fieldpath.PaveObject` function that has very similar functionality to this one. The key difference being that it always invokes the `DefaultUnstructuredConverter`. Do we need both?

pkg/fieldpath/object/object.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	oMap, err := runtime.DefaultUnstructuredConverter.ToUnstructured(o)
+	if err != nil {
+		return nil, false, err
+	}
+	return fieldpath.Pave(oMap), true, nil
+}
+
+// PatchFieldValueToObject applies the value to the "to" object at the given
```
negz, line 43: I'm still a little uncomfortable with the fact that we're adding what appear to be convenience functions that operate on `runtime.Object` for patching, but not for any other part of the `fieldpath` package.

My concern is not in thinking that `runtime.Object` is the wrong level to operate at, but rather that merging now appears to be a special case that is asymmetrical from the rest of the API. If the caller wants to set or get a value they handle conversion to a `*Paved` and use `SetValue` and `GetValue`. However if the caller wants to patch, they use this special package that operates on `runtime.Object`.

Perhaps this makes sense if you imagine this package would be expanded to handle more than just merging with time?

pkg/fieldpath/object/object.go
ulucinar: If we decide to keep `ToPaved`, I will update the documentation as you suggested above.

pkg/fieldpath/object/object.go
ulucinar: I agree. That's why I had also suggested that we can maybe move these convenience functions to core crossplane. The `Paved.MergeValue` is coherent with the rest of the `fieldpath` package, but we also need some higher level functions that will extract the value to be merged into the destination from a source `runtime.Object`, as you also acknowledge. 

What do you think about moving these convenience functions to `crossplane/crossplane` (I'm calling this as "core crossplane", not sure if it conveys the right meaning)? I'm also fine if they stay under `fieldpath/object` as they act as a bridge between `runtime.Object`s and `Paved`s. Maybe a little less for `object.MergeReplace`, but then even `object.MergeReplace` accepts a `path` parameter.

pkg/resource/api.go
ulucinar: Hi @negz,
Took a look at server-side apply. My understanding is that it is mainly addressing a somewhat different kind of a problem than the feature we are trying to implement, though we can employ conflict resolution mechanics of server-side apply to achieve merging/replacing fields as you suggested. And this does *not* mean that we cannot change the feature formulation (or that my understanding does not have flaws). 

My understanding is that server-side apply primarily aims a "more stable object lifecycle" by keeping track of appliers (aka fieldManagers) as the managers of a set of fields, and giving a conflict if a field is being modified by another applier than the current fieldManager. We can control how field ownership will be defined, at "compile" time, via merge strategies, e.g., define a merge-strategy of `granular` for a map if we would like to allow different appliers to control different keys of a map, or a merge-strategy of `atomic` if would like the map to be managed by a single applier with all its keys.

With the current implementation here, we are allowing a composition to specify the merge strategy at runtime and it works bidirectionally, i.e., while merging from composite into composed or from composed into composite. One composition can request a different merge strategy than another. 

I'm not sure but server-side apply sounded to me like something of a broader context than adding this feature. For example, I think we may want to reconsider `resource.APIPatchingApplicator` if we decide to employ apply patches and the reconcilers themselves (though we may also choose to use server-side apply only in specific contexts).

pkg/fieldpath/object/object.go
negz: Thanks! I don't have a strong opinion either. Usually I would bias for API simplicity over performance optimisations though, at least until we experience noticeable performance issues. In this case I'm happy for you to choose - we can either deprecate the old function or not add the new one.

pkg/fieldpath/object/object.go
negz: Hrm I can see a case either way. Perhaps we should start with these in core Crossplane - we can move them here if we find our library of functions that operate on objects grows.

ulucinar: I've moved the functions in `fieldpath.object` to core Crossplane including `resource.WithMergeOptions` and unexported them. If we need to use them elsewhere, we can consider to export them but for now, no need to export them. Please see the dependent [PR](https://github.com/crossplane/crossplane/pull/2409).

pkg/fieldpath/object/object.go
ulucinar: Removed `fieldpath.ToPaved` and kept `fieldpath.PaveObject`.

ulucinar:
Hi @negz,
Thank you very much for the review. I really appreciate it. I see it as a great learning opportunity. 

I realized that when `MergeOptions.AppendSlice` is set and the intersection between the destination and the source slices is not empty, we get duplicate items in the destination slice.  I believe this is not the UX we would want and I've added [fieldpath.removeSourceDuplicates](https://github.com/crossplane/crossplane-runtime/blob/4d6ec580bef8d7e9576e4ed954d44a6b38026948/pkg/fieldpath/merge.go#L80) to handle duplicates by default. I have also added test cases in `pkg/fieldpath/merge_test.go` for this case and various others.

pkg/fieldpath/merge.go
```diff
@@ -0,0 +1,108 @@
...
+// If a nil merge options is supplied, the default behavior is MergeOptions'
+// default behavior. If dst or src is nil, src is returned
+// (i.e., dst replaced by src).
+func merge(dst, src interface{}, mergeOptions *xpv1.MergeOptions) (interface{}, error) {
+	const keyArg = "arg"
+	argWrap := func(arg interface{}) map[string]interface{} {
+		return map[string]interface{}{
+			keyArg: arg,
+		}
+	}
```
negz, line 61: Could you add a comment explaining why this `argWrap` function is needed? I can't quite figure it out.

negz (approved):
Thanks for persisting with this @ulucinar! One last thing, but otherwise this looks good to me.

pkg/fieldpath/merge.go
ulucinar: Done.

PR #275, opened 2021-07-23
Add event when resource need reconcile

### LINKED ISSUE #274: Add event in resource on observe out of sync with diff in message
Like verbose logging was added in #271. Also add diff message in resource events.

More easy to debug resources via `kubectl describe` without reading logs. 
Common users of provider can do `kubectl describe` to found problem, when read logs can only provider admins.

vaspahomov:

Added diff message in resources events.

Fixes #274

I have:

I've built provider-azure with this patch. Changed one of resource and looked in `kubectl describe` event section.

```
  Normal   NeedReconcileManagedResource  7s (x5 over 27s)      managed/virtualnetwork.network.azure.crossplane.io  AddressSpace:   &network.AddressSpace{
-          AddressPrefixes: &[]string{"192.168.0.1/16"},
+          AddressPrefixes: &[]string{"192.168.0.0/16"},
  }
```

On larger diff looks like:
```
  Normal   NeedReconcileManagedResource  5m19s (x16 over 15m)  managed/virtualnetwork.network.azure.crossplane.io  AddressSpace:   &network.AddressSpace{
-          AddressPrefixes: &[]string{"192.168.0.1/16"},
+          AddressPrefixes: &[]string{"192.168.0.0/16"},
  }
  Normal  NeedReconcileManagedResource  12s (x4 over 2m51s)  managed/virtualnetwork.network.azure.crossplane.io  AddressSpace:   &network.AddressSpace{
-         AddressPrefixes: &[]string{"192.168.0.1/16"},
+         AddressPrefixes: &[]string{"192.168.0.0/16"},
  }
EnableDdosProtection:   &bool(
-   true,
+   false,
  )
EnableVMProtection:   (*bool)(
-   &true,
+   nil,
  )
```

### DIFF
--- a/pkg/reconciler/managed/reconciler.go
+++ b/pkg/reconciler/managed/reconciler.go
@@ -72,6 +72,8 @@ const (
 	reasonDeleted event.Reason = "DeletedExternalResource"
 	reasonCreated event.Reason = "CreatedExternalResource"
 	reasonUpdated event.Reason = "UpdatedExternalResource"
+
+	reasonNeedReconcile event.Reason = "NeedReconcileManagedResource"
 )
 
 // ControllerName returns the recommended name for controllers that use this
@@ -797,6 +799,7 @@ func (r *Reconciler) Reconcile(_ context.Context, req reconcile.Request) (reconc
 
 	if observation.Diff != "" {
 		log.Debug("External resource differs from desired state", "diff", observation.Diff)
+		record.Event(managed, event.Normal(reasonNeedReconcile, observation.Diff))
 	}
 
 	update, err := external.Update(externalCtx, managed)


pkg/reconciler/managed/reconciler.go
```diff
@@ -797,6 +799,7 @@ func (r *Reconciler) Reconcile(_ context.Context, req reconcile.Request) (reconc
 
 	if observation.Diff != "" {
 		log.Debug("External resource differs from desired state", "diff", observation.Diff)
+		record.Event(managed, event.Normal(reasonNeedReconcile, observation.Diff))
```
negz, line 802:
```suggestion
		record.Event(managed, event.Normal(reasonNeedReconcile, "Diff:\n" + observation.Diff))
```

What do you think about this change? I figure since the diff will always be multi-line it might read more clearly if it starts on its own line.

pkg/reconciler/managed/reconciler.go
```diff
@@ -72,6 +72,8 @@ const (
 	reasonDeleted event.Reason = "DeletedExternalResource"
 	reasonCreated event.Reason = "CreatedExternalResource"
 	reasonUpdated event.Reason = "UpdatedExternalResource"
+
+	reasonNeedReconcile event.Reason = "NeedReconcileManagedResource"
```
negz, line 76: I'm not sure about this event name - technically we're already reconciling the managed resource. 🤔 What about something like `DriftDetected` or `UpdateNeeded`?

negz (commented):
Thanks for putting this together @vaspahomov! I'm still quite on the fence about this functionality. I can see it helping with debugging, and the readability doesn't seem too bad though I do wonder what it would look like with larger objects. My key concern is regarding the number and verbosity of events we're emitting - we already have some feedback that larger Crossplane deployments are abusing the API server due to the number of events we're emitting.

pkg/reconciler/managed/reconciler.go
negz: Another idea might be to include this with the `UpdatedExternalResource` event we already send.

hasheddan (commented):
Thanks so much for having an eye for improving this user experience @vaspahomov! I am pretty strongly against including this information in the events because I view events as informing what happened, and potentially providing information on how to investigate, while this feels like giving all of this information to the user directly, in a potentially very verbose manner.

However, putting my user hat on, I know that making this info readily accessible to users would be a huge feature that many folks would appreciate. I wonder if there is an alternate way that we could surface this easily to users. I hesitate to venture outside of the k8s API to do so, but I can't think of a great way to provide data given the APIs it offers.

One option could be to have controllers write this type of data to a `ConfigMap` when a given field was set on the object. This would be a bit of a gross solution, but you can imagine something similar to `writeConnectionSecretToRef`, maybe `writeDiagnosticsToRef`. This could eventually be a pluggable sink, similar to what is proposed for secrets in https://github.com/crossplane/crossplane/issues/2366

smcavallo:
Here is how I would prefer this to work for crossplane:

There is a similar feature in argocd where you can set the sync options to false, in which case it will not clobber/update the resource if `selfHeal` is disabled.
https://argoproj.github.io/argo-cd/user-guide/auto_sync/#automatic-self-healing
```
      syncPolicy:
        automated:
          selfHeal: false
```
image


There are also annotations for deleting resources, similar to `deletionPolicy: Orphan`
https://argoproj.github.io/argo-cd/user-guide/sync-options/
```
  annotations:
    argocd.argoproj.io/sync-options: Prune=false
```
image

And you can control the criteria argocd uses to determine whether an object is out of sync or needs to be synced:
https://argoproj.github.io/argo-cd/user-guide/diffing/
```
spec:
  ignoreDifferences:
  - group: apps
    kind: Deployment
    name: guestbook
    namespace: default
    jsonPointers:
    - /spec/replicas
```

What I would prefer from crossplane:
- spec is the API representation of the object ONLY
- sync interval and options (ex. deletion policy) are set via annotations or syncOptions property on each object
- something similar to `ignoreDifferences` with jsonPath can be used to ignore differences (and ignore updates from being called if a value in the `cr.Spec.ForProvider` doesn't match the API)
- a representation of the actual API object stored somewhere (configmap?) so it can be diffed against the crossplane representation of the object

This would put a lot more control in the hands of the end users, especially around controlling what syncs, when it syncs, how it syncs.
It could enable a really nice UI, especially around visualizing differences.

To achieve this, it would be great if crossplane was much more opinionated about `IsUpToDate.`  A big painpoint today is implementing `IsUpToDate` when writing a new resource.  It requires the most code, the most tests, and it's also where the most bugs seem to be.  Each resource has it's own unique implementation which has been decided by the implementer and they are very prone to errors.
I would prefer if there were a consistent way to diff.  Ex. each resource forced you to create a 2 instances of the same class/struct, one populated with the data from crossplane and one populated with the data from the remote service (maybe with DeepCopy helper?)  It would always be an apples to apples comparison, and consistent across all providers and resources.
An example would be - copy `cr.Spec.ForProvider` into `S3CreateBucket`, copy the results of `DescribeBucket` into `S3CreateBucket` object, and then compare the 2 `S3CreateBucket` objects.
Crossplane could even `json.Marshal` both objects and expose a json diff helper for the comparison.  This would DRY up a lot of the resources and unit tests.

Long story short, instead of logging verbose diffs and json, i'd prefer to set `selfHeal: false` to stop the bleeding for resources which are constantly syncing and to be able to then be able to compare the 2 objects using some other method.

benagricola:
The implementation on this looks fine but I somewhat agree with @hasheddan and @negz - Crossplane is already fast and loose with the number of events it submits (particularly with misbehaving resources), and I think adding to that by logging potentially very large diffs (or the same small diff repeatedly) will just exacerbate the issue. 

If it could be enabled when necessary (e.g. via an annotation) then that would be less of a problem, but I don't think it should be on by default in its' current form.

On the `upToDate` points, I agree - it has been the source of most of the bugs I've found (both in my own code and other providers) and I think the provider ecosystem would really benefit from having some more boilerplate for performing update checks in a safe and tested manner. 

Myself and colleagues went through various implementations of `upToDate` when working on `provider-cloudflare` and the least bug-prone approach I found was to convert the cloud resource definition to the relevant `Parameters` struct and then use `cmp.Equal()` (ex: [here](https://github.com/benagricola/provider-cloudflare/blob/be1e11a31946ab2cc3ffa0c6e8e4758dcacf7fe1/internal/clients/sslsaas/customhostnames/custom_hostnames.go#L142)), which I think is what @smcavallo is suggesting above - lifting that approach out in a manner that could be used cross-provider would be a big help for further provider dev.

chlunde:
I have also worked on many upToDate bugs and would find it valuable to log the full diff in debug mode.

For an event, I think it would be sufficient and probably acceptable to log only _field names_ of the non-up-to-date resource, probably just the top level elements. This would probably remove the need for debug-logging in 9/10 cases for me, and would be enough hint for a non-developer to find a workaround in many cases, and good bug reports / issues here on github. It would also avoid leaking data into events and other systems.

I also agree on the number of events is a bit too much, so if we could log changed keys as part of the `UpdatedExternalResource` event, I would be very happy.

luebken:
I would think controlling the debug level in events is a common issue in K8s. Any other prior art in addition to the ArgoCD example.

negz:
> I have also worked on many upToDate bugs and would find it valuable to log the full diff in debug mode.

Keep in mind we already have _logging_ of these diffs, just not events. So today folks who are debugging a managed resource (e.g. in development or in a particularly tricky production environment) have the option to restart a provider in `--debug` mode to see these diffs. The question is whether we should expose them as events also. Put otherwise, do we want this information to be available (with a little work) in debugging situations, or just available all the time?

> I would think controlling the debug level in events is a common issue in K8s.

@luebken This is not the case, at least as far as I know. I believe the verbosity of events is typically fixed - not tunable - and that they tend to be succinct as @hasheddan points out. That said, there's not a lot of guidance on their use. The API conventions document has [a small section](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#events).

negz:
I think there's a bunch of great feedback in this issue WRT implementing `IsUptoDate` functions being painful and error prone, as well as some interesting suggestions as to how we can deal with constant update cycles. We may want to track these in distinct issues. In particular the idea of having a way to stop reconciles has come up a few times, and I know other projects (e.g. ArgoCD, Config Connector) support variations of this pattern. I've always been a little wary of it personally (couldn't it be misleading for some resources to _not_ actually be authoritative?) but it seems to be a common enough approach that I'm coming around to it. :)

negz:
This is a tough one, but after discussion at the last community meeting the consensus among the maintainers is that we'd prefer to keep this at the debug logging level for now, so I am going to close this. We do appreciate the suggestion though, and the discussion it spurred around the trickiness of IsUpToDate.

PR #280, opened 2021-08-16
Add machinery for tracking external resource create time

negz:

Adds plumbing to fix at least one variant of https://github.com/crossplane/provider-aws/issues/802.

Per https://github.com/crossplane/provider-aws/issues/802 some external APIs (including some AWS APIs) appear to experience some delay between the time a new resource is successfully created and the time at which that resource appears in queries.

This commit adds a new 'crossplane.io/external-create-time' annotation and a new 'ResourcePending' field in the Observation struct. Together these can be used by an Observe method to allow for a small grace period before it determines that a resource does not exist. For example:

```go
func Observe(ctx context.Context, mg resource.Managed) (managed.ExternalObservation, error) {
  err := api.Get(AsInput(mg))
  if IsNotFound(err) {
    if t := meta.GetExternalCreateTime(); t != nil && time.Since(t.Time) < 1 * time.Minute {
      // We're in the grace period - wait a bit longer for the resource to appear.
      return managed.ExternalObservation{ResourcePending: true}, nil
    }
    // The resource does not exist.
    return managed.ExternalObservation{ResourceExists: false}, nil
  }
  if err != nil {
    return managed.ExternalObservation{}, err
  }
  return managed.ExternalObervation{ResourceExists: true}
}

func Create(ctx context.Context, mg resource.Managed) (managed.ExternalCreation, error) {
  _ := api.Create(AsInput(mg))
  meta.SetExternalCreateTime()
  return managed.ExternalCreation{ExternalCreateTimeSet: true}, nil
}
```

I have:

Per https://github.com/crossplane/crossplane-runtime/pull/279 I've updated the AWS `RouteTable` and `InternetGateway` controllers to use the above pattern and run the script in https://gist.github.com/negz/e1f2e74f18802d15440214a1a1abc981 for 12+ hours. I did not observe any leaked route tables or internet gateways during my testing, whereas without this fix I reliably see 2-3 leaked resources after 2-3 hours.

### DIFF
--- a/pkg/meta/meta.go
+++ b/pkg/meta/meta.go
@@ -270,9 +270,9 @@ func GetExternalCreateTime(o metav1.Object) *metav1.Time {
 }
 
 // SetExternalCreateTime sets the time at which the external resource was most
-// recently created to the current time.
-func SetExternalCreateTime(o metav1.Object) {
-	AddAnnotations(o, map[string]string{AnnotationKeyExternalCreateTime: metav1.Now().Format(time.RFC3339)})
+// recently created to the supplied time.
+func SetExternalCreateTime(o metav1.Object, t metav1.Time) {
+	AddAnnotations(o, map[string]string{AnnotationKeyExternalCreateTime: t.Format(time.RFC3339)})
 }
 
 // AllowPropagation from one object to another by adding consenting annotations
--- a/pkg/meta/meta.go
+++ b/pkg/meta/meta.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"hash/fnv"
 	"strings"
+	"time"
 
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
@@ -31,9 +32,21 @@ import (
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 )
 
-// AnnotationKeyExternalName is the key in the annotations map of a resource for
-// the name of the resource as it appears on provider's systems.
-const AnnotationKeyExternalName = "crossplane.io/external-name"
+const (
+	// AnnotationKeyExternalName is the key in the annotations map of a
+	// resource for the name of the resource as it appears on provider's
+	// systems.
+	AnnotationKeyExternalName = "crossplane.io/external-name"
+
+	// AnnotationKeyExternalCreateTime is the key in the annotations map of
+	// a resource that represents the last time the external resource was
+	// created successfully. Its value must be an RFC3339 timestamp, which
+	// can be used to determine how long ago a resource was created. This is
+	// useful for eventually consistent APIs that may take some time before
+	// the API called by Observe will report that a recently created
+	// external resource exists.
+	AnnotationKeyExternalCreateTime = "crossplane.io/external-create-time"
+)
 
 // Supported resources with all of these annotations will be fully or partially
 // propagated to the named resource of the same kind, assuming it exists and
@@ -243,6 +256,23 @@ func GetExternalName(o metav1.Object) string {
 // SetExternalName sets the external name annotation of the resource.
 func SetExternalName(o metav1.Object, name string) {
 	AddAnnotations(o, map[string]string{AnnotationKeyExternalName: name})
+}
+
+// GetExternalCreateTime returns the time at which the external resource was
+// most recently created.
+func GetExternalCreateTime(o metav1.Object) *metav1.Time {
+	a := o.GetAnnotations()[AnnotationKeyExternalCreateTime]
+	t, err := time.Parse(time.RFC3339, a)
+	if err != nil {
+		return nil
+	}
+	return &metav1.Time{Time: t}
+}
+
+// SetExternalCreateTime sets the time at which the external resource was most
+// recently created to the current time.
+func SetExternalCreateTime(o metav1.Object) {
+	AddAnnotations(o, map[string]string{AnnotationKeyExternalCreateTime: metav1.Now().Format(time.RFC3339)})
 }
 
 // AllowPropagation from one object to another by adding consenting annotations
--- a/pkg/reconciler/managed/reconciler.go
+++ b/pkg/reconciler/managed/reconciler.go
@@ -48,7 +48,7 @@ const (
 // Error strings.
 const (
 	errGetManaged               = "cannot get managed resource"
-	errUpdateManagedAfterCreate = "cannot update managed resource. this may have resulted in a leaked external resource"
+	errUpdateManagedAnnotations = "cannot update managed resource. this may have resulted in a leaked external resource"
 	errReconcileConnect         = "connect failed"
 	errReconcileObserve         = "observe failed"
 	errReconcileCreate          = "create failed"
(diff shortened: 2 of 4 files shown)


pkg/meta/meta.go
```diff
@@ -245,6 +258,23 @@ func SetExternalName(o metav1.Object, name string) {
...
+		return nil
+	}
+	return &metav1.Time{Time: t}
+}
+
+// SetExternalCreateTime sets the time at which the external resource was most
+// recently created to the current time.
+func SetExternalCreateTime(o metav1.Object) {
+	AddAnnotations(o, map[string]string{AnnotationKeyExternalCreateTime: metav1.Now().Format(time.RFC3339)})
+}
```
negz, line 276: I wonder whether this function should take the time as an argument, rather than assuming it to be the call time? I feel like in 99% of cases it would be called as `meta.SetExternalCreateTime(o, metav1.Now())` but perhaps that's not so bad for a little extra flexibility.

pkg/reconciler/managed/reconciler.go
```diff
@@ -316,12 +326,19 @@ type ExternalObservation struct {
...
+	// needed for external resource's with unpredictable external names that
+	// are returned from the API at create time.
 	ExternalNameAssigned bool
 
+	// ExternalCreateTimeSet should be true if the Create operation resulted
+	// in the resource's external create time being set. This is typically
+	// only needed to accommodate APIs that have some delay between when an
+	// external resource is created and when the external resource is
+	// reported to exist.
+	ExternalCreateTimeSet bool
```
negz, line 340: In retrospect it might have been nicer for this API to return these values rather than signal that they'd been set, but since we've established a pattern with `ExternalNameAssigned` I figure we may as well stick with it.

negz: Hrm - another approach here would be to just always set the `crossplane.io/external-create-time` annotation in the `managed.Reconciler` rather than relying on `Create` implementations to optionally set it. I could imagine the time at which the external resource was created being of some use to folks even in a purely informational sense for debugging etc.

I think the only downside here would be that we'd always `Update` the managed resource immediately after calling `Create`, which resets any pending managed resource status updates. Today the `Update` is effectively opt-in, though to be fair these implications aren't really obvious to `ExternalClient` authors as is.

turkenh: I would agree with this approach.

Regarding the downside, couldn't we fix that by storing status to a temp var before running `Update()` and then run a `Status.Update()` using the stored status?

negz: > Regarding the downside, I believe we could fix that by storing status to a temp var before running Update() and then run a Status.Update() using the stored status?

Proooobably - the challenge would be extracting it. The `managed.Reconciler` deals in the `resource.Managed` interface, which has no awareness of the `Spec` and `Status` of the type it contains. Off the top of my head the only way to do this would be to convert to an `*Unstructured`, which I'm a little hesitant to do for such a core library.

I'd perhaps lean toward just advising folks that they should not update the `Status` during the `Create` call. I believe the existing convention is to modify as little of the managed resource as possible during `Create`, then use `Observe` to set `Status` and late-init `Spec`.

turkenh: I remember that I've dealt with this situation by:
1. Taking a deepcopy of the object with the desired status, 
2. Do the `Update`
3. Set resource version returned by update call
4. and do a status update. 

Something like:

```
++ mgWithDesiredStatus := managed.DeepCopyObject()

err := retry.OnError(retry.DefaultRetry, resource.IsAPIError, func() error {
	nn := types.NamespacedName{Name: managed.GetName()}
	if err := r.client.Get(ctx, nn, managed); err != nil {
		return err
	}
	meta.SetExternalName(managed, en)
	return r.client.Update(ctx, managed)
})
if err != nil {
	log.Debug("Cannot update managed resource", "error", err)
	record.Event(managed, event.Warning(reasonCannotUpdateManaged, errors.Wrap(err, errUpdateManagedAfterCreate)))
	managed.SetConditions(xpv1.ReconcileError(errors.Wrap(err, errUpdateManagedAfterCreate)))
}

++ mgWithDesiredStatus.(metav1.Object).SetResourceVersion(managed.GetResourceVersion())
++ r.client.Status().Update(ctx, mgWithDesiredStatus)
```

Feels a bit hacky, but worked well so far :)

negz: This sounds a little like what our "updating" applicator does. I think the only risk there is that copying the resource version defeats the API server's concurrency control (i.e. we'd blindly overwrite anything that happened between our deepcopy and our status update). I suppose that's not a _huge_ issue when it comes to managed resource status, given we don't really expect write contention with other controllers on the status field, and we also don't rely on any state being persisted to that object.

pkg/meta/meta.go
negz: I decided to update this to take the create time as an argument.

pkg/reconciler/managed/reconciler.go
negz: I've pushed a commit that moves responsibility for setting the `crossplane.io/external-create-time` annotation into the `managed.Reconciler`, and thus updates annotations after every successful create. For the time being I've opted not to worry about persisting status. It occurred to me that the current implementation actually discards _all_ updates made during `Create` with the exception of annotation changes.

To my knowledge we don't have any `Create` implementations that expected spec or status updates to be persisted automatically, so I've documented this expectation.

pkg/reconciler/managed/reconciler.go
```diff
@@ -631,6 +650,12 @@ func (r *Reconciler) Reconcile(_ context.Context, req reconcile.Request) (reconc
 		return reconcile.Result{Requeue: true}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
 	}
 
+	if observation.ResourcePending {
```
muvaf, line 653: Is there a way to calculate this here instead of expecting controller to return? I wonder if we can say `ignore ResourceExists: false if the creation time annotation is set and X seconds delay has not passed yet` where X is given in `Setup` function of the controller. I guess the downside is that retries will have some delay (mostly 1-2s maybe?) but it'd let us not have to worry about eventually consistent APIs one by one when people open issues about leaking.

pkg/reconciler/managed/reconciler.go
```diff
@@ -725,26 +750,27 @@ func (r *Reconciler) Reconcile(_ context.Context, req reconcile.Request) (reconc
...
+		meta.SetExternalCreateTime(managed, metav1.Now())
+
+		// We try to persist annotation several times because if we fail
+		// to persist the external name annotation (if necessary) we'll
+		// leak our newly created external resource. Any other changes
+		// made during Create will be reverted; at the time of writing
+		// Create implementations are advised not to alter status, but
+		// we may revisit this in future.
+		a := managed.GetAnnotations()
+		if err := retry.OnError(retry.DefaultRetry, resource.IsAPIError, func() error {
```
muvaf, line 762: Note that we run this update with retries only in case external name assigned because we lose the status when we do this operation and there are cases where information is only available as result of creation output. One possible solution could be to save the creation time before we call `Create` though that would require two updates with retries, the one after create being inside conditional, which increases the complexity here.

negz:
@muvaf / @chlunde I like your suggestion to avoid `ExternalClient` authors needing to deal with (much of) this. I'm working on a new PR that combines this one with https://github.com/crossplane/crossplane-runtime/pull/279 (per @turkenh's suggestion). I think it should be possible to make your changes there.

negz:
Closing in favor of https://github.com/crossplane/crossplane-runtime/pull/283.

PR #283, opened 2021-08-31
Account for two different kinds of consistency issues

negz:

Fixes https://github.com/crossplane/provider-aws/issues/802
Closes https://github.com/crossplane/crossplane-runtime/pull/279
Closes https://github.com/crossplane/crossplane-runtime/pull/280

This PR is intended to address two issues that we diagnosed while investigating https://github.com/crossplane/provider-aws/issues/802. It addresses feedback from and supercedes https://github.com/crossplane/crossplane-runtime/pull/279 and https://github.com/crossplane/crossplane-runtime/pull/280. Both issues are known to cause 'leaked' and duplicate managed resources. I have only been able to personally reproduce the second issue but we have evidence of the first happening in the wild.

The first issue is that controller-runtime does not guarantee reads from cache will return the freshest version of a resource. It's possible we could create an external resource in one reconcile, then shortly after trigger another in which it appears that the managed resource was never created because we didn't record its external-name. This only affects the subset of managed resources with non-deterministic external-names that are assigned during creation.

The second issue is that some external APIs are eventually consistent. A newly created external resource may take some time before our ExternalClient's observe call can confirm it exists. AWS EC2 is an example of one such API.

This PR attempts to address the first issue by making an Update to a managed resource immediately before Create it called. This Update call will be rejected by the API server if the managed resource we read from cache was not the latest version.

It attempts to address the second issue by allowing managed resource controller authors to configure an optional grace period that begins when an external resource is successfully created. During this grace period we'll requeue and keep waiting if Observe determines that the external resource doesn't exist, rather than (re)creating it.

I have:

https://gist.github.com/negz/e1f2e74f18802d15440214a1a1abc981

I've run the script from the above gist for about 12 hours against provider-aws built with this PR and observed zero leaked resources.

### DIFF
--- a/pkg/meta/meta.go
+++ b/pkg/meta/meta.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"hash/fnv"
 	"strings"
+	"time"
 
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
@@ -31,9 +32,33 @@ import (
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 )
 
-// AnnotationKeyExternalName is the key in the annotations map of a resource for
-// the name of the resource as it appears on provider's systems.
-const AnnotationKeyExternalName = "crossplane.io/external-name"
+const (
+	// AnnotationKeyExternalName is the key in the annotations map of a
+	// resource for the name of the resource as it appears on provider's
+	// systems.
+	AnnotationKeyExternalName = "crossplane.io/external-name"
+
+	// AnnotationKeyExternalCreatePending is the key in the annotations map
+	// of a resource that indicates the last time creation of the external
+	// resource was pending. Its value must be an RFC3999 timestamp. This
+	// annotation is removed once we record the result of external resource
+	// creation.
+	AnnotationKeyExternalCreatePending = "crossplane.io/external-create-pending"
+
+	// AnnotationKeyExternalCreateSucceeded is the key in the annotations
+	// map of a resource that represents the last time the external resource
+	// was created successfully. Its value must be an RFC3339 timestamp,
+	// which can be used to determine how long ago a resource was created.
+	// This is useful for eventually consistent APIs that may take some time
+	// before the API called by Observe will report that a recently created
+	// external resource exists.
+	AnnotationKeyExternalCreateSucceeded = "crossplane.io/external-create-succeeded"
+
+	// AnnotationKeyExternalCreateFailed is the key in the annotations map
+	// of a resource that indicates the last time creation of the external
+	// resource failed. Its value must be an RFC3999 timestamp.
+	AnnotationKeyExternalCreateFailed = "crossplane.io/external-create-failed"
+)
 
 // Supported resources with all of these annotations will be fully or partially
 // propagated to the named resource of the same kind, assuming it exists and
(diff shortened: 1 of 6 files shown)


pkg/reconciler/managed/reconciler.go
```diff
@@ -581,6 +628,18 @@ func (r *Reconciler) Reconcile(_ context.Context, req reconcile.Request) (reconc
...
+	// to persist our annotations after the ExternalClient's Create method
+	// was called. If that is the case we may have lost a critical update to
+	// the external name and leaked a resource. The safest thing to do is to
+	// refuse to proceed.
+	if meta.GetExternalCreatePending(managed) != nil {
+		log.Debug(errCreatePending)
+		record.Event(managed, event.Warning(reasonCannotInitialize, errors.New(errCreatePending)))
+		managed.SetConditions(xpv1.ReconcileError(errors.New(errCreatePending)))
+		return reconcile.Result{Requeue: true}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
+	}
```
negz, line 641: Heh, I just kicked off a test and every managed resource immediately got stuck on this clause. I see that all managed resources have both a create pending and a successful create timestamp set at roughly the same time, for example:

```yaml
kind: InternetGateway                                                                    
metadata:           
  annotations:     
    crossplane.io/composition-resource-name: internetgateway
    crossplane.io/external-create-pending: "2021-08-31T07:29:04Z"
    crossplane.io/external-create-succeeded: "2021-08-31T07:29:04Z"
    crossplane.io/external-name: igw-08657141fbbe34faf
```

Adding the successful creation annotation should remove the creation pending annotation, so I guess the pending annotation is somehow being immediately re-added? I need to sign off for the evening, so I'll work through this tomorrow.

negz: I figured out the problem - I was using `meta.AddAnnotations(o, a)` rather than `o.SetAnnotations(a)`. I'm a little worried that using `SetAnnotations` in our tight retry loop might 'undo' annotations being added by another controller, so I've refactored how the annotations are expected to work in a separate commit.

negz:
```
2021-09-01T13:18:18.202Z        DEBUG   provider-aws    cannot determine creation result - remove the crossplane.io/external-create-pending annotation if it is safe to proceed {"controller": "managed/routetable.ec2.aws.crossplane.io", "request": "/test-2-ff7wr", "uid": "1afdaa6a-82ec-4dfd-9d44-4f269f9d5eaa", "version": "444726", "external-name": ""}
```

Interestingly I did see the above case once in the logs from my 12-13 hours of testing. It seems to have resolved itself (the RouteTable was cleaned up successfully. I'm currently experiencing some login issues so I can't actually confirm whether or not the RouteTable was leaked, but I imagine it wasn't because if it was it would have blocked deletion of the VPC managed resource it existed in.

pkg/reconciler/managed/reconciler.go
```diff
@@ -721,30 +804,49 @@ func (r *Reconciler) Reconcile(_ context.Context, req reconcile.Request) (reconc
...
+
+		// We handle annotations specially here because it's critical
+		// that they are persisted to the API server. If we don't remove
+		// the external-create-pending annotation the reconciler will
+		// refuse to proceed, because it won't know whether or not it
+		// created an external resource. This is important in cases
+		// where we must record an external-name annotation set by the
+		// Create call. Any other changes made during Create will be
+		// reverted when annotations are updated; at the time of writing
+		// Create implementations are advised not to alter status, but
```
muvaf, line 842: We could possibly do something like `managed.GetStatus().DeepCopy` and after the critical update, `managed.SetStatus(copied)` but that'd require change in `Managed` interface and the code generator, so maybe in the future. But I'll open an issue just to have a place to see how big of a problem this will be.

pkg/meta/meta.go
```diff
@@ -31,9 +32,32 @@ import (
...
+	// AnnotationKeyExternalName is the key in the annotations map of a
+	// resource for the name of the resource as it appears on provider's
+	// systems.
+	AnnotationKeyExternalName = "crossplane.io/external-name"
+
+	// AnnotationKeyExternalCreatePending is the key in the annotations map
+	// of a resource that indicates the last time creation of the external
+	// resource was pending (i.e. about to happen). Its value must be an
+	// RFC3999 timestamp.
+	AnnotationKeyExternalCreatePending = "crossplane.io/external-create-pending"
```
muvaf, line 45:
```suggestion
	AnnotationKeyExternalCreatePending = "crossplane.io/external-creation-pending"
```

nitpick

pkg/reconciler/managed/reconciler.go
```diff
@@ -412,6 +437,16 @@ func WithPollInterval(after time.Duration) ReconcilerOption {
 	}
 }
 
+// WithCreationGracePeriod configures an optional period during which we will
+// wait for the external API to report that a newly created external resource
+// exists. This allows us to tolerate eventually consistent APIs that do not
+// immediately report that newly created resources exist when queried.
+func WithCreationGracePeriod(d time.Duration) ReconcilerOption {
+	return func(r *Reconciler) {
+		r.creationGracePeriod = d
```
muvaf, line 446: What do you think about giving this a default like 10 seconds? I think that'd cover most (all?) cases we want to cover without having to think about which resources need it and update their code.

muvaf (approved):
LGTM! Thanks for fixing this hard problem with the least code change necessary in providers!

pkg/reconciler/managed/reconciler.go
```diff
@@ -721,30 +804,49 @@ func (r *Reconciler) Reconcile(_ context.Context, req reconcile.Request) (reconc
 			// the new error condition. If not, we requeue explicitly, which will trigger backoff.
 			log.Debug("Cannot create external resource", "error", err)
 			record.Event(managed, event.Warning(reasonCannotCreate, err))
+
+			// We handle annotations specially here because it's
+			// critical that they are persisted to the API server.
+			// If we don't remove the external-create-pending
```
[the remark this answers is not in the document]
negz: Good catch - this comment was based on a previous implementation.

pkg/reconciler/managed/reconciler.go
negz: Yeah @turkenh suggested the same thing on a previous iteration of this PR. I'd prefer to avoid it unless there's a use case for modifying spec and/or status during `Create`.

pkg/meta/meta.go
negz: Unless you're super opinionated here I'm just going to stick with `create` mostly out of laziness. 😬

pkg/reconciler/managed/reconciler.go
negz: I went with 30.

chlunde:
Nice, looks good!

negz:
/backport

(Now that the commands workflow actually exists in this repo.)

negz:
/backport

This time for the v0.13 and v0.14 releases.

PR #292, opened 2021-09-15
Set `Creating` and `Deleting` conditions close to `Status().Update()` calls 

### LINKED ISSUE #285: Status updates in `ExternalClient.Create` are omitted

With https://github.com/crossplane/crossplane-runtime/pull/283, we issue a standard update in all cases and that causes the `status` to be lost since managed resources treat it as subresource. But there are cases where some information about the resource is available only in the response of the creation call to the API.

We could possibly expand `resource.Managed` interface to have `GetStatus() runtime.Object` and `SetStatus(runtime.Object)` and then `DeepCopy` the status before the update operation and later call `SetStatus` on the returned object and issue a status update this time.

negz:

Relates to #285

This approach causes us to repeat ourselves a bit, but prevents issues like the above, where refactoring caused us to accidentally overwrite a pending status update that we hadn't committed.

I have:

I haven't yet.

### DIFF
--- a/pkg/reconciler/managed/reconciler.go
+++ b/pkg/reconciler/managed/reconciler.go
@@ -584,7 +584,6 @@ func (r *Reconciler) Reconcile(_ context.Context, req reconcile.Request) (reconc
 	// to unpublish connection details and remove finalizer.
 	if meta.WasDeleted(managed) && managed.GetDeletionPolicy() == xpv1.DeletionOrphan {
 		log = log.WithValues("deletion-timestamp", managed.GetDeletionTimestamp())
-		managed.SetConditions(xpv1.Deleting())
 
 		// Empty ConnectionDetails are passed to UnpublishConnection because we
 		// have not retrieved them from the external resource. In practice we
@@ -598,7 +597,7 @@ func (r *Reconciler) Reconcile(_ context.Context, req reconcile.Request) (reconc
 			// backoff.
 			log.Debug("Cannot unpublish connection details", "error", err)
 			record.Event(managed, event.Warning(reasonCannotUnpublish, err))
-			managed.SetConditions(xpv1.ReconcileError(err))
+			managed.SetConditions(xpv1.Deleting(), xpv1.ReconcileError(err))
 			return reconcile.Result{Requeue: true}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
 		}
 		if err := r.managed.RemoveFinalizer(ctx, managed); err != nil {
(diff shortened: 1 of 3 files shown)


pkg/reconciler/managed/reconciler.go
```diff
@@ -607,7 +606,7 @@ func (r *Reconciler) Reconcile(_ context.Context, req reconcile.Request) (reconc
 			// condition. If not, we requeue explicitly, which will trigger
 			// backoff.
 			log.Debug("Cannot remove managed resource finalizer", "error", err)
-			managed.SetConditions(xpv1.ReconcileError(err))
+			managed.SetConditions(xpv1.Deleting(), xpv1.ReconcileError(err))
 			return reconcile.Result{Requeue: true}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
 		}
```
muvaf, line 612: I think we need to set `xpv1.Creating()` in line 638 as well, which is called `if meta.ExternalCreateIncomplete(managed)`

muvaf: Line 699 as well, in `if !observation.ResourceExists && meta.ExternalCreateSucceededDuring(managed, r.creationGracePeriod)`

muvaf: Line 793, where `if err := r.client.Update(ctx, managed); err != nil`

muvaf (approved):
Need a few additions. The rest looks good to me but I think this fixes only one part of #285 . We still lose the status set in `external.Create`.

negz:
> We still lose the status set in external.Create so we shouldn't close that issue.

Oh - I missed that. I didn't think we actually had any managed resources that modified status during `Create` (apart from setting the `Creating` status condition, which we now handle in the reconciler). Do you have an example of resources that do this? Now that I think about it I'm vaguely remembering some of the Azure resources recording operations there?

muvaf:
I had not thought about Azure but yes and in fact, it could be showstopper since [the creation process](https://github.com/crossplane/provider-azure/blob/master/pkg/controller/database/mysqlserver/managed.go#L167) is being tracked with that ID. I was thinking about AWS, where the response of `Create` operation is [stored](https://github.com/crossplane/provider-aws/blob/master/pkg/controller/rds/dbinstance/zz_controller.go#L117) in status, which in some cases is different than `Describe`, though I couldn't find an example now.

FWIW, there is this [KEP](https://github.com/kubernetes/enhancements/pull/2537) by Tim Hockin that removes the litmus test of `If I erased the whole status struct, would everything in it be reconstructed (or at least be reconstructible) from observation?` So, `status` won't be treated as ephemeral as it used to be.

PR #293, opened 2021-09-20
Add a `controller.Options` type

negz:

This type is intended to be passed as the argument to most Crossplane Setup functions, for example:

```go
func Setup(mgr ctrl.Manager, o controller.Options) error
```

This allows us to add new options to be plumbed down to all or most controllers without increasing the number of arguments provided to each Setup function. Sets of controllers that require additional arguments (e.g. the pkg controllers from crossplane/crossplane) can define their own Options struct that embeds this one.

This PR also migrates the `*feature.Flags` type from crossplane/crossplane, and generalises the names of the two levels of default rate limiters we use. This is all in preparation of adding more configuration knobs to crossplane/crossplane and our providers (e.g. poll interval, average global RPS) 

I have:

I've built crossplane/crossplane in a branch that uses this commit and ensured it works.

### DIFF
--- a/pkg/controller/options.go
+++ b/pkg/controller/options.go
@@ -27,6 +27,18 @@ import (
 	"github.com/crossplane/crossplane-runtime/pkg/ratelimiter"
 )
 
+// DefaultOptions returns a functional set of options with conservative
+// defaults.
+func DefaultOptions() Options {
+	return Options{
+		Logger:                  logging.NewNopLogger(),
+		GlobalRateLimiter:       ratelimiter.NewGlobal(ratelimiter.DefaultGlobalRPS),
+		PollInterval:            1 * time.Minute,
+		MaxConcurrentReconciles: 1,
+		Features:                &feature.Flags{},
+	}
+}
+
 // Options frequently used by most Crossplane controllers.
 type Options struct {
 	// The Logger controllers should use.
--- a/pkg/controller/options.go
+++ b/pkg/controller/options.go
@@ -1,0 +1,46 @@
+/*
+Copyright 2021 The Crossplane Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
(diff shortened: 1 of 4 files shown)


pkg/feature/feature.go
```diff
@@ -0,0 +1,48 @@
+/*
+Copyright 2019 The Crossplane Authors.
```
hasheddan, line 2: Nitpick:
```suggestion
Copyright 2021 The Crossplane Authors.
```

pkg/feature/feature_test.go
```diff
@@ -0,0 +1,50 @@
+/*
+Copyright 2019 The Crossplane Authors.
```
hasheddan, line 2: nitpick:
```suggestion
Copyright 2021 The Crossplane Authors.
```

hasheddan (approved):
Thanks @negz! Glad to have the controller function signatures getting a little cleaner 👍🏻

negz:
Requesting another review pass because I ended up pushing a few more commits after this was reviewed. :)

pkg/controller/options.go
```diff
@@ -0,0 +1,68 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+
+	"github.com/crossplane/crossplane-runtime/pkg/feature"
+	"github.com/crossplane/crossplane-runtime/pkg/logging"
+	"github.com/crossplane/crossplane-runtime/pkg/ratelimiter"
+)
+
+// DefaultOptions returns a functional set of options with conservative
+// defaults.
+func DefaultOptions() Options {
```
hasheddan, line 32: @negz is motivation for using a non-pointer type here that we may pass the options to multiple consumers who may in turn desire to modify them? And features being a pointer type ensures that feature flags are enabled consistently throughout all consumers?

negz: > Is motivation for using a non-pointer type here that we may pass the options to multiple consumers who may in turn desire to modify them?

Close, but also kind of the opposite. I tend to prefer to pass structs by value when:

1. Consumers are expected to treat the struct's values as read-only.
2. There's no significant performance benefit to passing by reference.

I could see that passing by value might signal "you can mutate your copy", but I feel it also signals "you're not allowed to mutate my copy".

You could make a good argument that `Options` falls under the "small structs that might grow" clause of https://github.com/golang/go/wiki/CodeReviewComments#pass-values, but I'm not super worried about performance loss due to copying this struct given that it's not really in any hot paths.

> And features being a pointer type ensures that feature flags are enabled consistently throughout all consumers?

Actually I think I originally had intended for `*Flags` to be passed by value for similar reasons to `Options`, but the `go vet` `copylocks` check reminded me that I shouldn't be copying mutexes.

PR #294, opened 2021-09-24
Support true global reconcile rate limiting

negz:

This PR tweaks how ratelimiters are applied to support _actual_ global reconcile
rate limiting - that is all reconcile triggers are rate limited, not just some.

See https://github.com/crossplane/crossplane/issues/2595 for details.

I have:

I've built crossplane/crossplane against this change and tested that it works. I'd like to do a little more testing to validate that the rate limiter is actually working as expected though.

### DIFF
--- a/pkg/controller/options.go
+++ b/pkg/controller/options.go
@@ -32,7 +32,7 @@ import (
 func DefaultOptions() Options {
 	return Options{
 		Logger:                  logging.NewNopLogger(),
-		GlobalRateLimiter:       ratelimiter.NewGlobal(ratelimiter.DefaultGlobalRPS),
+		GlobalRateLimiter:       ratelimiter.NewController(),
 		PollInterval:            1 * time.Minute,
 		MaxConcurrentReconciles: 1,
 		Features:                &feature.Flags{},
@@ -63,6 +63,6 @@ type Options struct {
 func (o Options) ForControllerRuntime() controller.Options {
 	return controller.Options{
 		MaxConcurrentReconciles: o.MaxConcurrentReconciles,
-		RateLimiter:             ratelimiter.NewController(o.GlobalRateLimiter),
+		RateLimiter:             ratelimiter.NewController(),
 	}
 }
--- a/pkg/ratelimiter/reconciler.go
+++ b/pkg/ratelimiter/reconciler.go
@@ -1,0 +1,51 @@
+/*
+Copyright 2021 The Crossplane Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package ratelimiter
+
+import (
+	"context"
+	"time"
+
+	"sigs.k8s.io/controller-runtime/pkg/ratelimiter"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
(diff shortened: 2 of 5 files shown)


pkg/ratelimiter/reconciler.go
```diff
@@ -0,0 +1,51 @@
...
+
+// Reconcile the supplied request subject to rate limiting.
+func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	item := r.name + req.String()
+	if d := r.limit.When(item); d > time.Duration(0) {
+		return reconcile.Result{RequeueAfter: d}, nil
+	}
+	r.limit.Forget(item)
+	return r.inner.Reconcile(ctx, req)
+}
```
negz, line 51: There are two downsides to this approach:

1. It rate limits work queue pops, not pushes. There's an increased chance of unbounded queue growth.
2. It artificially inflates the `controller_runtime_reconcile_total` metric, since some 'reconciles' are rate limited and sent immediately back to the work queue.

pkg/controller/options.go
```diff
@@ -63,6 +63,6 @@ type Options struct {
 func (o Options) ForControllerRuntime() controller.Options {
 	return controller.Options{
 		MaxConcurrentReconciles: o.MaxConcurrentReconciles,
-		RateLimiter:             ratelimiter.NewController(o.GlobalRateLimiter),
+		RateLimiter:             ratelimiter.NewController(),
```
negz, line 66: The idea is each controller uses `NewController` to produce a per-item exponential backoff controller to rate limit reconciles that return `Requeue: true` or a non-nil `error`.

The `NewGlobal` reconciler is passed to `ratelimiter.NewReconciler` to rate limit _all_ reconciles.

pkg/ratelimiter/reconciler.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+// A Reconciler rate limits an inner, wrapped Reconciler. Requests that are rate
+// limited immediately return RequeueAfter: d without calling the wrapped
+// Reconciler, where d is imposed by the rate limiter.
+type Reconciler struct {
+	name  string
+	inner reconcile.Reconciler
+	limit ratelimiter.RateLimiter
+}
```
negz, line 34: I could see a pretty good case for this being in `pkg/reconciler` package instead. We have a practice of using subpackages there to avoid collisions so it could either be `ratelimited.Reconciler` in `pkg/reconciler/ratelimited` or we could make `pkg/reconciler` a package and have it be `reconciler.RateLimited`. No strong feelings really - here feels pretty good too.

pkg/ratelimiter/reconciler.go
```diff
@@ -0,0 +1,51 @@
...
+func NewReconciler(name string, r reconcile.Reconciler, l ratelimiter.RateLimiter) *Reconciler {
+	return &Reconciler{name: name, inner: r, limit: l}
+}
+
+// Reconcile the supplied request subject to rate limiting.
+func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	item := r.name + req.String()
+	if d := r.limit.When(item); d > time.Duration(0) {
+		return reconcile.Result{RequeueAfter: d}, nil
+	}
```
negz, line 48: Hrm, it actually looks like this doesn't work like I thought it worked.

I was thinking that `When` would return the duration after which calling `When` again would be successful (i.e. for the token bucket when it was refilled enough to handle new requests). In retrospect I believe what's happening under the hood (at least for a token bucket based rate limiter) is that it's taking a token from the bucket each time, then returning the time at which you're allowed to "cash in" the token you already took.

So in our case we shouldn't be calling `When` again in the `RequeueAfter` case and instead just proceeding, but unfortunately we don't have enough state for new reconciles to know whether they've been rate limited or not.

negz: Pushed a new commit that fixes this.

pkg/ratelimiter/reconciler.go
negz: FWIW in practice I haven't seen this being an issue during my scale testing.

PR #296, opened 2021-10-01
Add Disconnect call in Reconcile

vaspahomov:

I've add Disconnect call in end of Reconcile.
Some sdks required Close call after use. But I still want to create sdk in Connect call rather then create sdk in each Observe, Create, Update, Delete. And it's not possible to dispose sdk in the end of Reconcile.

I have:

### DIFF
--- a/pkg/reconciler/managed/reconciler.go
+++ b/pkg/reconciler/managed/reconciler.go
@@ -52,6 +52,7 @@ const (
 	errUpdateManagedAnnotations = "cannot update managed resource annotations"
 	errCreateIncomplete         = "cannot determine creation result - remove the " + meta.AnnotationKeyExternalCreatePending + " annotation if it is safe to proceed"
 	errReconcileConnect         = "connect failed"
+	errReconcileDisconnect      = "disconnect failed"
 	errReconcileObserve         = "observe failed"
 	errReconcileCreate          = "create failed"
 	errReconcileUpdate          = "update failed"
@@ -61,6 +62,7 @@ const (
 // Event reasons.
 const (
 	reasonCannotConnect       event.Reason = "CannotConnectToProvider"
+	reasonCannotDisconnect    event.Reason = "CannotDisconnectFromProvider"
 	reasonCannotInitialize    event.Reason = "CannotInitializeManagedResource"
 	reasonCannotResolveRefs   event.Reason = "CannotResolveResourceReferences"
 	reasonCannotObserve       event.Reason = "CannotObserveExternalResource"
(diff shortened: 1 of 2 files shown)


pkg/reconciler/managed/reconciler.go
```diff
@@ -673,10 +736,17 @@ func (r *Reconciler) Reconcile(_ context.Context, req reconcile.Request) (reconc
 		// condition. If not, we requeue explicitly, which will trigger
 		// backoff.
 		log.Debug("Cannot connect to provider", "error", err)
-		record.Event(managed, event.Warning(reasonCannotConnect, err))
+		record.Event(managed, event.Warning(reasonCannotDisconnect, err))
```
negz, line 739: This should stay as `CannotConnect`, right?

pkg/reconciler/managed/reconciler.go
```diff
@@ -60,7 +61,7 @@ const (
 
 // Event reasons.
 const (
-	reasonCannotConnect       event.Reason = "CannotConnectToProvider"
+	reasonCannotDisconnect    event.Reason = "CannotDisconnectToProvider"
```
negz, line 64:
```suggestion
	reasonCannotDisconnect    event.Reason = "CannotDisconnectFromProvider"
```

Also, I think `reasonCannotConnect` is still needed.

pkg/reconciler/managed/reconciler.go
```diff
@@ -269,6 +319,11 @@ func (c *NopConnecter) Connect(_ context.Context, _ resource.Managed) (ExternalC
 	return &NopClient{}, nil
 }
 
+// Disconnect do nothing. It never returns an error.
```
negz, line 322:
```suggestion
// Disconnect does nothing. It never returns an error.
```

pkg/reconciler/managed/reconciler.go
```diff
@@ -187,6 +188,46 @@ type ExternalConnecter interface {
...
+}
+
+// NewExternalConnectorClosable wraps ExternalConnecter in ExternalConnecterClosable
+func NewExternalConnectorClosable(c ExternalConnecter, d ExternalDisconnecter) ExternalConnecterClosable {
+	return &closableExternalConnector{c, d}
+}
+
+// An ExternalConnecterClosable produces a new ExternalClient given the supplied
+// Managed resource.
+type ExternalConnecterClosable interface {
```
negz, line 226:
```suggestion
type ExternalConnectDisconnecter interface {
```

What do you think about this name? I think it reads a little more like the typical Go interfaces like `io.ReadWriter` or `io.ReadWriteCloser`. I slightly prefer `Disconnecter` to `Closable` because the new method is `Disconnect` rather than `Close`.

negz: Also, technically this should probably be `Disconnector` (i.e. ending in "tor", not "ter") but we already have a spelling mistake in `ExternalConnecter`. 😦

pkg/reconciler/managed/reconciler.go
```diff
@@ -552,7 +616,7 @@ func NewReconciler(m manager.Manager, of resource.ManagedKind, o ...ReconcilerOp
 }
 
 // Reconcile a managed resource with an external resource.
-func (r *Reconciler) Reconcile(_ context.Context, req reconcile.Request) (reconcile.Result, error) { // nolint:gocyclo
+func (r *Reconciler) Reconcile(_ context.Context, req reconcile.Request) (result reconcile.Result, err error) { // nolint:gocyclo
```
negz, line 619: I'd like to avoid named return values here - this function is very complex and I think it's likely to become a footgun.

https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters

pkg/reconciler/managed/reconciler.go
```diff
@@ -677,6 +741,19 @@ func (r *Reconciler) Reconcile(_ context.Context, req reconcile.Request) (reconc
...
+			record.Event(managed, event.Warning(reasonCannotDisconnect, disconnectErr))
+			disconnectErr = errors.Wrap(disconnectErr, errReconcileDisconnect)
+			managed.SetConditions(xpv1.ReconcileError(disconnectErr))
+			if err != nil {
+				err = errors.Wrap(err, disconnectErr.Error())
+			} else {
+				err = disconnectErr
+			}
+		}
+	}()
```
negz, line 756: I think this block changed since last time I looked. Is the idea that we're trying to avoid squashing the original error in cases where the deferred function is triggered by a case in which we'd return an error? I think it's probably safe to just log and emit an event for disconnect errors without ever returning them - it's not like requeueing (which happens when we return an error) would help recover from not being able to disconnect.

negz (changes requested):

negz (approved):

pkg/reconciler/managed/reconciler.go
```diff
@@ -677,6 +741,13 @@ func (r *Reconciler) Reconcile(_ context.Context, req reconcile.Request) (reconc
 		managed.SetConditions(xpv1.ReconcileError(errors.Wrap(err, errReconcileConnect)))
 		return reconcile.Result{Requeue: true}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
 	}
+	defer func() {
+		if disconnectErr := r.external.Disconnect(ctx); disconnectErr != nil {
+			log.Debug("Cannot disconnect from provider", "error", disconnectErr)
+			record.Event(managed, event.Warning(reasonCannotDisconnect, disconnectErr))
+			managed.SetConditions(xpv1.ReconcileError(errors.Wrap(disconnectErr, errReconcileDisconnect)))
```
negz, line 748: I think this `SetCondition` call might cause flapping/overwritten conditions in cases where the deferred function is called because we're returning from another error case where we would also update the same condition. I'll fix this up in a future PR though.

PR #297, opened 2021-10-04
Add wildcard expand method to fieldpath.Paved

turkenh:

During terrajet work, we need a way to define [sensitive fields](https://github.com/crossplane-contrib/terrajet/pull/77) as fieldpaths. However, since these paths are defined during code generation, where actual data is not available yet (e.g. managed resource spec), we needed a way to define them as wildcards for fields with arrays and/or maps. 

This PR adds `ExpandWildcards` method on `fieldpath.Paved` object. This function returns available fieldpaths by expanding wildcard (`*`) characters in the given input.

Example: 

For a Paved object with the following data: 
```
[]byte(`{"spec":{"containers":[{"name":"cool", "image": "latest", "args": ["start", "now", "debug"]}]}}`),
```

`ExpandWildcards("spec.containers[*].args[*]")` returns:

`[]string{"spec.containers[0].args[0]", "spec.containers[0].args[1]", "spec.containers[0].args[2]"}`

I have:

With unit tests and [the terrajet PR](https://github.com/crossplane-contrib/terrajet/pull/77) relying on this functionality.

### DIFF
--- a/pkg/fieldpath/paved.go
+++ b/pkg/fieldpath/paved.go
@@ -17,6 +17,8 @@ limitations under the License.
 package fieldpath
 
 import (
+	"strconv"
+
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/json"
 
@@ -83,7 +85,10 @@ func (p *Paved) SetUnstructuredContent(content map[string]interface{}) {
 }
 
 func (p *Paved) getValue(s Segments) (interface{}, error) {
-	var it interface{} = p.object
+	return getValueFromInterface(p.object, s)
+}
+
+func getValueFromInterface(it interface{}, s Segments) (interface{}, error) {
 	for i, current := range s {
 		final := i == len(s)-1
 		switch current.Type {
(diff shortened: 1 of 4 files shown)


pkg/fieldpath/paved.go
```diff
@@ -120,6 +124,80 @@ func (p *Paved) getValue(s Segments) (interface{}, error) {
...
+//
+// For a Paved object with the following data: []byte(`{"spec":{"containers":[{"name":"cool", "image": "latest", "args": ["start", "now", "debug"]}]}}`),
+// ExpandWildcards("spec.containers[*].args[*]") returns:
+// []string{"spec.containers[0].args[0]", "spec.containers[0].args[1]", "spec.containers[0].args[2]"},
+func (p *Paved) ExpandWildcards(path string) ([]string, error) {
+	segments, err := Parse(path)
+	if err != nil {
+		return nil, errors.Wrapf(err, "cannot parse path %q", path)
+	}
+	segmentsArray, err := expandWildcards(p.object, segments)
```
muvaf, line 141: Would it make sense to create `GetValues(path string) ([]interface{}, error)` (and other variants of it working with array) and call `expandWildcards` and `GetValue` functions?

turkenh: I am a bit hesitant to implement each variant at this moment, given it is already possible to first expand values and then iterate over returned fieldpaths and use existing Get* methods. I would prefer to consider it later if that is needed.

muvaf (approved):
LGTM! Thanks @turkenh !

PR #300, opened 2021-10-29
Only attempt object scheme parsing if object is not registered in meta

### LINKED ISSUE #299: Malformed `crossplane.yaml` leads to misleading errors
### What happened?

I was just trying to install a provider I had built, but found its revision was failing:

```
Status:
  Conditions:
    Last Transition Time:  2021-09-25T06:18:09Z
    Reason:                UnhealthyPackageRevision
    Status:                False
    Type:                  Healthy
  Controller Ref:
    Name:  
Events:
  Type     Reason             Age                From                                         Message
  ----     ------             ----               ----                                         -------
  Normal   BindClusterRole    29s (x4 over 35s)  rbac/providerrevision.pkg.crossplane.io      Bound system ClusterRole to provider ServiceAccount(s)
  Normal   ApplyClusterRoles  29s (x4 over 35s)  rbac/providerrevision.pkg.crossplane.io      Applied RBAC ClusterRoles
  Warning  ParsePackage       25s (x2 over 29s)  packages/providerrevision.pkg.crossplane.io  cannot parse package contents: no kind "Provider" is registered for version "meta.pkg.crossplane.io/v1alpha1" in scheme "pkg/runtime/scheme.go:100"
```

In my case I was actually running a branch with some refactoring to the revision reconciler, so I spent a good amount of time trying to figure out how I broke it. Turns out this is actually runtime scheme for "your YAML isn't valid". 😞 The real problem was that I was missing a `:` after the `spec.controller.image` field of my provider.

### How can we reproduce it?

Take a `crossplane.yaml` file (in my case [this one](https://github.com/crossplane/provider-gcp/blob/f20119dac2a1fbffadc5f00dbe96be4240d4e1a5/package/crossplane.yaml), remove the colon from the final line, and set a valid controller image.

Crossplane version: Roughly 2b5b90dc7e2763e73eeb356f0b2c66b565a87386.

hasheddan:

Updates the package parsing logic to only attempt decoding with the
object scheme in the case that the error from decoding in the meta
scheme is due to the GVK not being registered. This does not change the
definition of a valid package, but does result in more informative
errors being returned when a package is invalid due to a malformed meta
type.

Fixes #299 

I have:

Tested with example in #299 and returned the following error (this is using Crossplane CLI, but same parsing is used in package manager, minus the annotated file position):
```
kubectl crossplane: error: failed to build package: failed to parse package: {path:/home/dan/code/github.com/crossplane/crossplane/pack/crossplane.yaml position:0}: v1alpha1.Provider.Spec: v1alpha1.ProviderSpec.Controller: readObjectStart: expect { or n, but found ", error found in #10 byte of ...|troller":"image cros|..., bigger context ...|cp"},"name":"provider-gcp"},"spec":{"controller":"image crossplane/provider-gcp-controller:VERSION"}|...
```

### DIFF
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -19,7 +19,6 @@ package parser
 import (
 	"bufio"
 	"context"
-	"fmt"
 	"io"
 	"io/ioutil"
 	"strings"
@@ -150,7 +149,7 @@ func isWhiteSpace(bytes []byte) bool {
 // annotateErr annotates an error if the reader is an AnnotatedReadCloser.
 func annotateErr(err error, reader io.ReadCloser) error {
 	if anno, ok := reader.(AnnotatedReadCloser); ok {
-		return errors.Wrap(err, fmt.Sprintf("%+v", anno.Annotate()))
+		return errors.Wrapf(err, "%+v", anno.Annotate())
 	}
 	return err
 }
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -112,26 +112,19 @@ func (p *PackageParser) Parse(ctx context.Context, reader io.ReadCloser) (*Packa
 		if len(bytes) == 0 {
 			continue
 		}
+		if isWhiteSpace(bytes) {
+			continue
+		}
 		m, _, err := dm.Decode(bytes, nil, nil)
 		if err != nil {
+			// NOTE(hasheddan): we only try to decode with object scheme if the
+			// error is due the object not being registered in the meta scheme.
+			if !runtime.IsNotRegisteredError(err) {
+				return pkg, annotateErr(err, reader)
+			}
 			o, _, err := do.Decode(bytes, nil, nil)
 			if err != nil {
-				empty := true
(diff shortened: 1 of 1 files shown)


pkg/parser/parser.go
```diff
@@ -141,6 +134,27 @@ func (p *PackageParser) Parse(ctx context.Context, reader io.ReadCloser) (*Packa
...
+			break
+		}
+	}
+	return empty
+}
+
+// annotateErr annotates an error if the reader is an AnnotatedReadCloser.
+func annotateErr(err error, reader io.ReadCloser) error {
+	if anno, ok := reader.(AnnotatedReadCloser); ok {
+		return errors.Wrap(err, fmt.Sprintf("%+v", anno.Annotate()))
```
negz, line 153: Could you use `errors.Wrapf` here?

negz (approved):

pkg/parser/parser.go
hasheddan: Ahh yes good call 👍🏻

PR #306, opened 2021-12-02
Tweak ExternalDisconnecter implementation

negz:

The primary functional change here is to avoid setting a status condition when a deferred disconnect fails. We don't want to overwrite the original status condition that may have been written if we're returning from Reconcile because we hit an error. Emitting an event and a debug log should be sufficient.

This commit also tweaks a bit of grammar and updates the NopConnectDisconnecter implementation to more closely match its docstring description.

I have:

I don't believe there are any consumers of this code yet, so I'm relying on unit tests.

### DIFF
--- a/pkg/reconciler/managed/reconciler.go
+++ b/pkg/reconciler/managed/reconciler.go
@@ -52,7 +52,6 @@ const (
 	errUpdateManagedAnnotations = "cannot update managed resource annotations"
 	errCreateIncomplete         = "cannot determine creation result - remove the " + meta.AnnotationKeyExternalCreatePending + " annotation if it is safe to proceed"
 	errReconcileConnect         = "connect failed"
-	errReconcileDisconnect      = "disconnect failed"
 	errReconcileObserve         = "observe failed"
 	errReconcileCreate          = "create failed"
 	errReconcileUpdate          = "update failed"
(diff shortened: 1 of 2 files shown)


pkg/reconciler/managed/reconciler.go
```diff
@@ -189,37 +188,32 @@ type ExternalConnecter interface {
...
-func NopDisconnect(_ context.Context) error {
-	return nil
-}
-
-// externalConnectDisconnecter is wrap ExternalConnecter in externalConnectDisconnecter
-// with nop Disconnect.
-type externalConnectDisconnecter struct {
+// A NopDisconnecter converts an ExternalConnecter into an
+// ExternalConnectDisconnecter with a no-op Disconnect method.
+type NopDisconnecter struct {
```
negz, line 199: FWIW I took a pass at a commit that fixes the various 'connecter' spelling mistakes, changing them to 'connector' mostly using type aliases. I decided to abandon it since the mistake has been there forever, and marking the incorrect spelling deprecated would cause quite a lot of downstream churn.

hasheddan: I actually didn't even realize that was a mistake -- I believe they are both words right?

hasheddan (approved):
LGTM -- thanks @negz!

pkg/reconciler/managed/reconciler.go
negz: https://www.askdifference.com/connector-vs-connecter/ - Huh I guess so. In that case we're inconsistent but at least nothing is incorrect haha.

PR #314, opened 2022-02-02
Allow manage objects with scope Namespaced

### LINKED ISSUE #251: Support Namespaced Managed Resources

Right now, crossplane-runtime does not formally support namespaced managed resources. It kind of works and only breaks if certain features are used. Because we (my team and I) were unaware that this was unintentional, we simply started using it.

It would be great if crossplane-runtime formally supported namespaced managed resources. I initially thought this [was a bug](https://github.com/crossplane/crossplane-runtime/pull/246), but it turns out this was simply not intended to work. But most of crossplane-runtime already works when using a namespaced managed resource, just by instructing kubebuilder: `+kubebuilder:resource:scope=Namespaced`. While we are aware of all the other Crossplane constructs that allow to have namespaced resources (compositions, publications etc.), it introduces a lot of configuration overhead, where we just could use the managed resource directly for some resources.

After a few glances at the code, it seems only a few changes here and there would be required to make references, externalName etc. work properly with namespaced resources. If you are not totally opposed to this, I can offer to contribute the required changes.

vaspahomov:
Signed-off-by: vaspahomov <vas2142553@gmail.com>

Added support of Namespaced Managed Resources. In my mind this patch will not affect on current behaviour with Cluster scoped resources. But with this patch you can use Namespaced objects as well as Cluster scoped.

Fixes #251

I have:

Manually tested with provider-aws, provider-azure with replace in go.mod

### DIFF
--- a/pkg/reconciler/managed/api.go
+++ b/pkg/reconciler/managed/api.go
@@ -176,7 +176,7 @@ func NewRetryingCriticalAnnotationUpdater(c client.Client) *RetryingCriticalAnno
 func (u *RetryingCriticalAnnotationUpdater) UpdateCriticalAnnotations(ctx context.Context, o client.Object) error {
 	a := o.GetAnnotations()
 	err := retry.OnError(retry.DefaultRetry, resource.IsAPIError, func() error {
-		nn := types.NamespacedName{Name: o.GetName()}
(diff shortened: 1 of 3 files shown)


negz:
I don't think we're likely to support this without a major version bump per my comment at https://github.com/crossplane/crossplane/issues/2255.

PR #323, opened 2022-03-04
External Secret Support in Composition Types

turkenh:

This PR makes the required changes to use External Secret Stores with Composition types:

- Refactors existing interfaces (e.g. `ConnectionPublisher.PublishConnection`) to use the same interface for both managed and Composition resources.
- Adds `publishConnectionDetailsTo` field to managed resource spec.
- Modifies `SecretStore` interface to support publishing/propagating for Composition types.
- Adds metadata support for Vault kv v1
- Tracks ownership of connection secrets with owner uid.
- Handles owner management for connection secrets, e.g.: 
  - cannot establish control of existing connection secret
  - existing secret is not owned by UID ...
  - refusing to modify unowned secret of type ...

Partially Fixes https://github.com/crossplane/crossplane/issues/2881

I have:

See https://github.com/crossplane/crossplane/pull/2940


apis/common/v1/resource.go
```diff
@@ -136,6 +136,14 @@ type ResourceSpec struct {
 	// +optional
 	WriteConnectionSecretToReference *SecretReference `json:"writeConnectionSecretToRef,omitempty"`
 
+	// PublishConnectionDetailsTo specifies the connection secret config which
+	// contains a name, metadata and a reference to secret store config to
+	// which any connection details for this managed resource should be written.
+	// Connection details frequently include the endpoint, username,
+	// and password required to connect to the managed resource.
+	// +optional
+	PublishConnectionDetailsTo *PublishConnectionDetailsTo `json:"publishConnectionDetailsTo,omitempty"`
```
negz, line 145: Cross reference to https://github.com/crossplane/crossplane/issues/2929 - might be worth considering before we introduce this new field.

turkenh: Responded there: https://github.com/crossplane/crossplane/issues/2929#issuecomment-1060493213

negz: Either way, it's probably worth documenting here and/or on WriteConnectionSecretToRef what happens when both are set / what their relationship is.

pkg/connection/fake/mocks.go
```diff
@@ -28,61 +30,26 @@ import (
...
-}
-
 // SecretStore is a fake SecretStore
 type SecretStore struct {
-	ReadKeyValuesFn   func(ctx context.Context, i store.Secret) (store.KeyValues, error)
-	WriteKeyValuesFn  func(ctx context.Context, i store.Secret, kv store.KeyValues) error
-	DeleteKeyValuesFn func(ctx context.Context, i store.Secret, kv store.KeyValues) error
+	ReadKeyValuesFn   func(ctx context.Context, i store.Secret) (managed.ConnectionDetails, error)
+	WriteKeyValuesFn  func(ctx context.Context, i store.Secret, conn managed.ConnectionDetails) error
+	DeleteKeyValuesFn func(ctx context.Context, i store.Secret, conn managed.ConnectionDetails) error
```
negz, line 37:
```suggestion
	DeleteConnectionDetailsFn func(ctx context.Context, i store.Secret, conn managed.ConnectionDetails) error
```

Is it worth changing the names of these methods to match the new args/return values?

pkg/connection/manager.go
```diff
@@ -156,5 +136,81 @@ func (m *DetailsManager) unpublishConnection(ctx context.Context, so SecretOwner
...
+		Name:     from.GetPublishConnectionDetailsTo().Name,
+		Scope:    from.GetNamespace(),
+		Metadata: from.GetPublishConnectionDetailsTo().Metadata,
+	})
+	if err != nil {
+		return false, errors.Wrap(err, "errGetSecret")
+	}
+
+	// TODO(turkenh): Implement an equivalent functionality to
+	//  "resource.ConnectionSecretMustBeControllableBy"
```
negz, line 188: I'm hazy on how this used to work, but I want to make sure that we won't face a security regression if we don't implement this. I think we might have been partly using `ConnectionSecretMustBeControllableBy` to try to avoid malicious folks abusing Crossplane by asking it to overwrite a secret that it shouldn't - for example to propagate `EvilXR`'s connection details to `VictimXRC`'s connection secret.

pkg/connection/manager_test.go
```diff
@@ -20,6 +20,10 @@ import (
 	"context"
 	"testing"
 
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
+
+	"github.com/crossplane/crossplane-runtime/pkg/reconciler/managed"
```
negz, line 25: Nit - extra whitespace between these imports.

pkg/reconciler/managed/api.go
```diff
@@ -100,21 +100,21 @@ func NewAPISecretPublisher(c client.Client, ot runtime.ObjectTyper) *APISecretPu
 // PublishConnection publishes the supplied ConnectionDetails to a Secret in the
 // same namespace as the supplied Managed resource. It is a no-op if the secret
 // already exists with the supplied ConnectionDetails.
-func (a *APISecretPublisher) PublishConnection(ctx context.Context, mg resource.Managed, c ConnectionDetails) error {
+func (a *APISecretPublisher) PublishConnection(ctx context.Context, o resource.ConnectionSecretOwner, c ConnectionDetails) error {
```
negz, line 103: Nice, I like making these take a narrower interface.

pkg/resource/interfaces.go
```diff
@@ -233,6 +243,8 @@ type CompositeClaim interface {
 	CompositionRevisionReferencer
 	CompositeResourceReferencer
 	LocalConnectionSecretWriterTo
+	// TODO(turkenh): LocalConnectionDetailsPublisherTo?
```
negz, line 246: I don't think that would be needed under the new design. I'm pretty sure the difference between `Connection` and `LocalConnection` was that one had a namespaced ref (e.g. with name and namespace) while the other was just a name (implying writing to the claim's namespace).

negz (commented):
Not approving just yet because this needs a rebase now that #322 is merged, and it seems like there's a few TODOs left to address but overall this looks good.

pkg/connection/fake/mocks.go
turkenh: Hmm, checking this again, I didn't want to leak `managed.ConnectionDetails` in `connection.store` package because this is the interface that SecretStore implementations need to satisfy and these are just KeyValues for them, i.e. connection details does not mean something for a secret store.
Refactored accordingly.

pkg/resource/interfaces.go
turkenh: Yeah, I can confirm this with my testing so far.

turkenh:
@negz this is ready for another pass.

I've resolved previous todo's and had to make some further changes especially to handle ownership and related checks. Tested end to end with https://github.com/crossplane/crossplane/pull/2940 and that PR has the latest changes.

apis/common/v1/connection_details.go
```diff
@@ -54,6 +60,22 @@ type ConnectionSecretMetadata struct {
 	Type *corev1.SecretType `json:"type,omitempty"`
 }
 
+// SetOwnerUID sets owner object uid label.
+func (in *ConnectionSecretMetadata) SetOwnerUID(so metav1.Object) {
+	if in.Labels == nil {
+		in.Labels = map[string]string{}
+	}
+	in.Labels[LabelKeyOwnerUID] = string(so.GetUID())
+}
```
negz, line 69: Presuming this is used to track ownership, should we add a note on the `ConnectionSecretMetadata` type that Secret Store implementations must implement label support in order to track secret ownership?

apis/common/v1/connection_details.go
```diff
@@ -18,6 +18,12 @@ package v1
 
 import (
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	// LabelKeyOwnerUID is the UID of the owner resource of a connection secret.
+	LabelKeyOwnerUID = "secret.crossplane.io/owner-uid"
```
negz, line 26: This seems like it would duplicate Kubernetes Secret controller references in some cases. I suppose we need to go this path if we want to be consistent though, given that some Kubernetes Secrets might live in different clusters, and obviously controller references don't exist in Vault. Perhaps a comment noting that would help future readers understand why we use labels for this?

pkg/connection/manager.go
```diff
@@ -133,28 +135,110 @@ func (m *DetailsManager) publishConnection(ctx context.Context, so SecretOwner,
...
-		Scope:    so.GetNamespace(),
+	err = ss.DeleteKeyValues(ctx, &store.Secret{
+		ScopedName: store.ScopedName{
+			Name:  p.Name,
+			Scope: so.GetNamespace(),
+		},
 		Metadata: p.Metadata,
-	}, kv), errWriteStore)
+		Data:     store.KeyValues(conn),
+	})
```
negz, line 145: Would using `SecretMustBeOwnedBy` make sense here too, to make sure we don't delete a secret we don't own?

pkg/connection/manager_test.go
```diff
@@ -205,8 +211,8 @@ func TestManagerPublishConnection(t *testing.T) {
 				err: errors.Wrap(errors.Wrapf(kerrors.NewNotFound(schema.GroupResource{}, "non-existing"), errGetStoreConfig), errConnectStore),
 			},
 		},
-		"SuccessfulPublish": {
-			reason: "We should return no error when published successfully.",
+		"CannotToPublish": {
```
negz, line 214:
```suggestion
		"CannotPublishTo": {
```

Nit

pkg/reconciler/managed/api.go
```diff
@@ -100,21 +102,37 @@ func NewAPISecretPublisher(c client.Client, ot runtime.ObjectTyper) *APISecretPu
...
 	s.Data = c
-	return errors.Wrap(a.secret.Apply(ctx, s, resource.ConnectionSecretMustBeControllableBy(mg.GetUID())), errCreateOrUpdateSecret)
+	err := a.secret.Apply(ctx, s,
+		resource.ConnectionSecretMustBeControllableBy(o.GetUID()),
+		resource.AllowUpdateIf(func(current, desired runtime.Object) bool {
+			// We consider the update to be a no-op and don't allow it if the
+			// current and existing secret data are identical.
+			return !cmp.Equal(current.(*corev1.Secret).Data, desired.(*corev1.Secret).Data, cmpopts.EquateEmpty())
+		}),
+	)
```
negz, line 120: Huh, weird, I guess the comment for this function was previously inaccurate in that it said it would be a no-op if the secret already existed.

pkg/reconciler/managed/publisher_test.go
```diff
@@ -22,6 +22,7 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 
+	v1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
```
negz, line 25: Nit: I think we usually import this as `xpv1`.

negz (approved):
This looks good to me - thank you! I've added a few comments but none of them are blocking. I won't merge right now just in case you want to follow up on any of the comment, but feel free to have @muvaf or someone else with access merge it if you don't want to wait for me to come online.

pkg/connection/manager.go
turkenh: Yes, good idea, done!

pkg/connection/manager.go
```diff
@@ -221,24 +221,25 @@ func (m *DetailsManager) connectStore(ctx context.Context, p *v1.PublishConnecti
...
-	return func(_ context.Context, current, desired *store.Secret) error {
-		o := ""
-		if current.Metadata != nil {
-			o = current.Metadata.GetOwnerUID()
-		}
-
-		switch {
-		case o == "" && current.Metadata != nil && current.Metadata.Type != nil && desired.Metadata != nil && current.Metadata.Type != desired.Metadata.Type:
-			return errors.Errorf(errFmtRefusingUnowned, *current.Metadata.Type)
-		case o == "":
```
turkenh: Removed taking ownership of an unowned resource assuming this was for some sort of backward compatibility thing. This simplified the logic here a lot.

turkenh:
> This looks good to me - thank you! I've added a few comments but none of them are blocking. I won't merge right now just in case you want to follow up on any of the comment, but feel free to have @muvaf or someone else with access merge it if you don't want to wait for me to come online.

Thanks a lot @negz for the review and approval. Resolved your last set of comments. 

@muvaf would you mind hitting the merge button? it would also be great if you could review the last two commits after @negz 's approval.

muvaf:
@turkenh I don't have all the context so just hit the merge button since it's already got the approval. Thanks!

PR #324, opened 2022-03-07
package.parser: make Or linter work with arbitrary number of linters

muvaf:

We need this in provider linter where we need to allow three distinct types, `CustomResourceDefinition`, `ValidatingWebhookConfiguration` and `MutatingWebhookConfiguration`.

I have:

Unit tests. No breaking change in the interface only the format of the printed error is different now.

### DIFF
--- a/pkg/parser/linter.go
+++ b/pkg/parser/linter.go
@@ -99,7 +99,7 @@ func (l *PackageLinter) Lint(pkg *Package) error {
 // error.
 func Or(linters ...ObjectLinterFn) ObjectLinterFn {
 	return func(o runtime.Object) error {
-		var errSet string
+		var errs []string
 		for _, l := range linters {
 			if l == nil {
 				return errors.New(errNilLinterFn)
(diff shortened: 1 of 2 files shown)


pkg/parser/linter.go
```diff
@@ -95,16 +97,19 @@ func (l *PackageLinter) Lint(pkg *Package) error {
...
-		bErr := b(o)
-		if aErr == nil || bErr == nil {
-			return nil
+		var errSet string
+		for _, l := range linters {
+			if l == nil {
+				return errors.New(errNilLinterFn)
+			}
+			err := l(o)
+			if err == nil {
```
muvaf, line 108: This is a behavioral difference that if any of the linters pass, we don't execute the next ones at all, which is I believe more in-line with how `OR` behavior is implemented in general.

pkg/parser/linter.go
```diff
@@ -95,16 +97,19 @@ func (l *PackageLinter) Lint(pkg *Package) error {
...
+				return nil
+			}
+			errSet += err.Error() + ", "
 		}
-		return errors.Errorf(errOrFmt, aErr, bErr)
+		return errors.Errorf(errOrFmt, strings.TrimSuffix(errSet, ", "))
```
negz, line 113: Nit: Could you just build a slice of strings then join them, to avoid needing to trim?

negz (approved):

pkg/parser/linter.go
muvaf: Ah I knew there was a better way to do this, thanks!

PR #328, opened 2022-04-14
Support for having circular dependencies while using referencers

### LINKED ISSUE #250: Referencers are resolved only once

I created a `DBSubnetGroup` with a single element `subnetIdRefs` array. It resolved the reference, reconciled but apparently a single subnet wasn't enough to create a DB Subnet Group in AWS: `DBSubnetGroupDoesNotCoverEnoughAZs:
      DB Subnet Group doesn''t meet availability zone coverage requirement. Please
      add subnets to cover at least 2 availability zones. Current coverage: 1`

Then I created another `Subnet` and added it to the `subnetIdRefs`. But there is no change; the list of actual subnetIds `subnetIds` still had only one element. The reference resolver didn't try to resolve my new referenced object.

I'd expect the `subnetIds` to include the new element. Delete & recreate solved this case for me but in case of update-able fields that are referenced, I think we should resolve them in every reconcile and return no-op in case there is no change in the resulting list.

### LINKED ISSUE #311: Resource creation fails when having circular dependencies using referencers
### What happened?
It does not seem possible to create objects with circular dependencies. A practical use case for this is an AWS security group which allows inbound traffic from other members of the same security group:

```yaml
kind: SecurityGroup
metadata:
  name: self-referential
spec:
  forProvider:
    region: eu-central-1
    vpcIdRef:
      name: sample-vpc
    groupName: self-referential
    description: self-referential security group
    ingress:
      - fromPort: 0
        toPort: 65535
        ipProtocol: tcp
        userIdGroupPairs:
          - groupIdRef: self-referential
  providerConfigRef:
    name: default
```

### How can we reproduce it?
Create object above

```
Name:         self-referential
Namespace:    
Labels:       <none>
Annotations:  <none>
API Version:  ec2.aws.crossplane.io/v1beta1
Kind:         SecurityGroup
...
Status:
  At Provider:
    Owner Id:           
    Security Group ID:  
  Conditions:
    Last Transition Time:  2022-01-21T20:36:38Z
    Message:               cannot resolve references: spec.forProvider.ingress[0].userIdGroupPairs[0].groupId: referenced field was empty (referenced resource may not yet be ready)
    Reason:                ReconcileError
    Status:                False
    Type:                  Synced
Events:
  Type     Reason                           Age                From                                         Message
  ----     ------                           ----               ----                                         -------
  Warning  CannotResolveResourceReferences  14s (x7 over 75s)  managed/securitygroup.ec2.aws.crossplane.io  cannot resolve references: spec.forProvider.ingress[0].userIdGroupPairs[0].groupId: referenced field was empty (referenced resource may not yet be ready)
```

Crossplane version: 1.6.1, provider-aws 0.23, EKS

sergenyalcin:

Fixes #311 
Fixes #250 

This PR adds a new API about the reference resolution/resolve policies. The new Policy struct has two Policy types:

- ResolvePolicy -> Always
- ResolutionPolicy -> Required, Optional

```
vpcIdRef:
  name: my-vpc
  policy:
    # Resolve specifies when this reference should be resolved. The default
    # is 'IfNotPresent', which will attempt to resolve the reference only when
    # the corresponding field is not present. Use 'Always' to resolve the
    # reference on every reconcile.
    resolve: Always
    # Resolution specifies whether resolution of this reference is required.
    # The default is 'Required', which means the reconcile will fail if the
    # reference cannot be resolved. 'Optional' means this reference will be
    # a no-op if it cannot be resolved.
    resolution: Optional
```

This Policy API was added for both Reference and Selector structs. Also, to provide the Always resolution, the order of the resolving changed. Before first Reference then Selector, now first Selector then Reference.

For details please check the following comments:

https://github.com/crossplane/crossplane-runtime/pull/328#discussion_r857786890
https://github.com/crossplane/crossplane-runtime/pull/328#issuecomment-1114225326
https://github.com/crossplane/crossplane-runtime/pull/328#issuecomment-1123308958
https://github.com/crossplane/crossplane-runtime/pull/328#discussion_r869985829

I have:

For circular dependency fix, this PR was tested by provisioning the SecurityGroup resource in provider-aws (with the following manifest). This resource includes a circular dependency:

```yaml
kind: SecurityGroup
metadata:
  name: sample-cluster-sg
spec:
  forProvider:
    region: us-east-1
    vpcIdRef:
      name: sample-vpc
    groupName: sample-ekscluster-sg
    description: Cluster communication with worker nodes
    ingress:
      - fromPort: 80
        toPort: 80
        ipProtocol: tcp
        userIdGroupPairs:
          - groupIdRef:
              name: sample-cluster-sg
              policy: 
                 resolution: Optional
        ipRanges:
          - cidrIp: 10.0.0.0/8
  providerConfigRef:
    name: example
```

For resolved only once issue, this PR was tested by provisioning the DBSubnetGroup resource with a problematic subnet reference list (with the following manifest):

```yaml
apiVersion: database.aws.crossplane.io/v1beta1
kind: DBSubnetGroup
metadata:
  name: sergen-subnet-group
spec:
  forProvider:
    region: us-east-1
    description: "sample group"
    subnetIdRefs:
      - name: sergen-subnet
  providerConfigRef:
    name: example
```

Then this resource was updated with the following subnet reference list:

```yaml
    subnetIdRefs:
      - name: sergen-subnet
         policy:
          resolve: Always
      - name: sergen-subnet2
        policy:
          resolve: Always
```

After adding the Always policy, I observed that the references were resolved for every reconcile loop.


apis/common/v1/resource.go
```diff
@@ -77,10 +77,47 @@ type SecretKeySelector struct {
...
+	// reference will be tried to resolve for every reconcile loop.
+	ReferencePolicyAlways ReferenceResolutionPolicy = "Always"
+)
+
 // A Reference to a named object.
 type Reference struct {
 	// Name of the referenced object.
 	Name string `json:"name"`
+	// Policies of the referenced object.
+	Policies []ReferenceResolutionPolicy `json:"policies,omitempty"`
```
muvaf, line 100: I'm not sure if the array is the best way to represent that configuration since it can take arbitrary number of policies whereas we only have two knobs to configure. Two distinct fields with policy string values seem like the way to go (see last point [here](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types)) but I'm struggling with coming up with names for those fields. What do you think about something like the following?
```yaml
vpcIdRef:
  name: my-vpc
  policy:
    resolve: Always # "IfNotPresent" by default.
    require: Never # "IfNotPresent" by default. "Always" could be the third option in the future.
```

apis/common/v1/resource.go
```diff
@@ -77,10 +77,47 @@ type SecretKeySelector struct {
...
+	ReferencePolicyOptional ReferenceResolutionPolicy = "Optional"
+
+	// ReferencePolicyAlways is a resolution option.
+	// When the ReferenceResolutionPolicy is set to ReferencePolicyAlways the
+	// reference will be tried to resolve for every reconcile loop.
+	ReferencePolicyAlways ReferenceResolutionPolicy = "Always"
+)
+
 // A Reference to a named object.
 type Reference struct {
```
muvaf, line 96: The policies here affect `ref -> value` transition but there is also `selector -> ref` where similar policies could apply. But as far as I can tell, both issues are about `ref -> value`, so having the policies only here could make sense for now until users see a need to customize `selector -> ref` as well.

apis/common/v1/resource.go
negz: This is a tricky one - I'm also having a tough time with naming. Writing it all out, I think we're trying to configure:

1. When the reference is resolved. Either only when the corresponding field is unset ("not present"), or always.
2. Whether resolution of this reference is optional or required.

We have a similar case to 2 already in our [`PatchPolicy` type](https://github.com/negz/crossplane/blob/2ad8a10/apis/apiextensions/v1/composition_patches.go#L55), where we have (e.g.) `fromFieldPath: Optional`. With that in mind I'd make a small tweak:

```yaml
vpcIdRef:
  name: my-vpc
  policy:
    # Resolve specifies when this reference should be resolved. The default
    # is 'IfNotPresent', which will attempt to resolve the reference only when
    # the corresponding field is not present. Use 'Always' to resolve the
    # reference on every reconcile.
    resolve: Always
    # Resolution specifies whether resolution of this reference is required.
    # The default is 'Required', which means the reconcile will fail if the
    # reference cannot be resolved. 'Optional' means this reference will be
    # a no-op if it cannot be resolved.
    resolution: Optional
```

As a side-note, I'm still wary of putting the policy _inside_ the reference itself, but I can't think of a better approach (apart from fixing this in generic references instead). This used to be explicitly disallowed by API conventions but that restriction appears to have been removed.

apis/common/v1/resource.go
negz: Doesn't for example https://github.com/crossplane/crossplane-runtime/issues/250#issuecomment-912800805 pertain to selectors? I could imagine the race @chlunde describes being a problem here, i.e.:

1. Foo A is created.
2. fooSelector selects Foo A and creates a reference.
3. Foo B is created
4. fooSelector would have selected Foo B, but doesn't because it already successfully selected its references.

sergenyalcin: Thank you for bringing up this, it was a point that I thought while working on these issues. Implemented supports/API for the `Reference` field can also be provided for `Selector` field, as you said. However, at this point, I think a new discussion can start.

Let me share my thoughts via an example:

```yaml
..
forProvider:
  vpcIdSelector:
    matchLabels:
      key: value
..
```

As you know, while we using the Selector field for resolution, we are filling the Reference field. So after we selected and resolved the reference by using the Selector, new state of the resource will be like that:

```yaml
..
forProvider:
  vpcId: vpc-1111111111
  vpcIdRef:
    name: sample-vpc
  vpcIdSelector:
    matchLabels:
      key: value
..
```

Now, when we review the [function] (this is Resolve function), we are observing that, firstly the Reference field is checked and then if it is not empty, the resolution process is done by using the Reference field and we return early. In other words, if the Reference field is set, we do not process the Selector field. Therefore, after this point, even if I update the Selector field, it will have no effect. Because the Reference field is full and only that field will be processed in the next loops.

I think this is an independent point from the resolution/resolve policies. Because, at this point, issues such as which resolution method is more dominant and which one to choose for situations where both are set will come to the fore. Let me share an example to clarify my point. Imagine that a user tries to provision an external resource by using the following manifest:

```yaml
..
forProvider:
  vpcIdRef:
    name: sample-vpc
  vpcIdSelector:
    matchLabels:
      key: value
..
```

For a resource that is being create by this manifest, we can say that the reference in the Reference field is dominant. So, the Selector field will not be used for resolution process. In fact, this is the same situation we encountered after selecting references and filling Reference field via Selector. And from this point on, the Selector field will not be considered again. Therefore, changes on this field, such as updating list or changing the policy, will not be affective.

As a result, I think providing support of new Policy API for Selector field makes sense. However, if I do not miss any point, it seems that we will have to do this with a solution of the point I tried to explain above.

[function]: https://github.com/crossplane/crossplane-runtime/blob/0e8935d6d9a942825a29ab3ab34f057cfb1dc050/pkg/reference/reference.go#L210

apis/common/v1/resource.go
```diff
@@ -77,10 +77,75 @@ type SecretKeySelector struct {
...
+)
+
+// IsResolutionPolicyOptional checks whether the resolution policy of relevant reference is Optional.
+func (p *Policy) IsResolutionPolicyOptional() bool {
+	if p != nil {
+		if p.Resolution != nil {
+			return *p.Resolution == ResolutionPolicyOptional
+		}
+	}
+	return false
```
negz, line 128:
```suggestion
	if p == nil || p.Resolution == nil {
		return false
	}
	return *p.Resolution == ResolutionPolicyOptional
```

A small nitpick - this is a good example of where returning early can reduce nested code blocks.

https://github.com/crossplane/crossplane/blob/master/CONTRIBUTING.md#return-early

sergenyalcin:
About my [comment], I want to propose a few solutions. Firstly, let me summarize my point again. When we set a Selector, we are selecting reference resources and putting them to the Reference field. In our Resolve functions, the Reference field is checking firstly, and after resolving references by using this Reference field we return early. This means that, Reference field is used as cache of Selector. So this blocks 'resolving always scenario' for Selector field (editing Selector and expecting some changes). Some proposed solutions:

- Changing the order of Reference and Selector fields. By this way, the Selector field will be dominant and for every loop, if there is a change in Selector field, Reference field will be updated.
- Now, we are writing the selected reference from Selector to Reference field. The another option can be that, differentiate the two APIS (Reference and Selector). I mean that, selected reference resource name from Selector cannot be written to the Reference field. By this way, we can protect the current logic in Resolve function.
- We may consider that, do not return early after resolving the Reference field. We can let that resolving process can continue for the Selector field also. With this approach, both of the two fields will be tried to resolve for every loop. As a behavioral change of this approach, we will support the using two APIs together.

To be honest, I don't have a strong feeling about which of these would be the better approach. What are your thoughts @negz about this point? Do you have any other approach for addressing this?

[comment]: https://github.com/crossplane/crossplane-runtime/pull/328#discussion_r857786890

muvaf:
> Changing the order of Reference and Selector fields. By this way, the Selector field will be dominant and for every loop, if there is a change in Selector field, Reference field will be updated.

I think this option is more intuitive from the perspective of users since the main job of selector is to select a reference and it makes it straight-forward to understand. If we add policy field to the selector as well, similar to reference, I think users could configure it while being aware of the risk that `if a new resource with the same labels is added, then the list will change`. I'd personally not opt for that risk but it's at least clear. In practice, I think it'll likely to be used temporarily during operations when a resource is stuck or wrong references are resolved and once it's done, the user would revert it back.

apis/common/v1/resource.go
muvaf: @negz Yes, you're right. We need to address the selector as well.

> As a result, I think providing support of new Policy API for Selector field makes sense. However, if I do not miss any point, it seems that we will have to do this with a solution of the point I tried to explain above.

@sergenyalcin If I understood you correctly, you're saying that the fact that we don't take any action after the ref is found will be changed. I think that's fine since it's configured by the user.

apis/common/v1/resource.go
sergenyalcin: Done!

pkg/reference/reference.go
```diff
@@ -213,32 +239,33 @@ func (r *APIResolver) Resolve(ctx context.Context, req ResolutionRequest) (Resol
 		return ResolutionResponse{ResolvedValue: req.CurrentValue, ResolvedReference: req.Reference}, nil
 	}
 
-	// The reference is already set - resolve it.
-	if req.Reference != nil {
-		if err := r.client.Get(ctx, types.NamespacedName{Name: req.Reference.Name}, req.To.Managed); err != nil {
-			return ResolutionResponse{}, errors.Wrap(err, errGetManaged)
+	// The selector is set - resolve it.
+	if req.Selector != nil {
```
muvaf, line 243: This results in behavior change for cases where both ref and selector are given and no policy is set. Chatted offline about possible solutions.

sergenyalcin: As we discussed, I preserved the order (first Reference field then Selector field). 

For new Always policy scenario, I implemented a possible solution (we discussed offline) that sets the Reference field to nil when Selector is not empty and resolve policy is always.

pkg/reference/reference.go
```diff
@@ -19,6 +19,8 @@ package reference
 import (
 	"context"
 
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+
 	"k8s.io/apimachinery/pkg/types"
```
muvaf, line 24:
```suggestion
	kerrors "k8s.io/apimachinery/pkg/api/errors"
	"k8s.io/apimachinery/pkg/types"
```
goimports

pkg/reference/reference.go
```diff
@@ -112,11 +114,24 @@ type ResolutionRequest struct {
...
-func (rr ResolutionRequest) IsNoOp() bool {
-	// We don't resolve values that are already set; we effectively cache
-	// resolved values. The CR author can invalidate the cache and trigger a new
-	// resolution by explicitly clearing the resolved value.
-	if rr.CurrentValue != "" {
+func (rr *ResolutionRequest) IsNoOp() bool {
+	isAlways := false
+	if rr.Selector != nil {
+		if rr.Selector.Policy.IsResolvePolicyAlways() {
+			rr.Reference = nil
```
muvaf, line 121: Could you leave a comment here about why we set it to `nil`?

muvaf (approved):
LGTM! Thank you @sergenyalcin for driving such a core change! I have some nits but it's good to go, let me know when it's ready to merge.

elohmrow:
💯

muvaf:
> As you can see, some inconsistent situations may occur when the new API is used. For now, when we see an Always policy for an element of list, we are resolving all references for every reconcile loop. In other words, it is sufficient for only one of the elements in the list have an Always policy for resolution in each loop.

@sergenyalcin Could you open an issue regarding this? I don't love that each element has its own policy but there isn't really a good way of adding a policy to the whole array without a breaking change. Maybe an annotation to point to the array field and say `policy of this field is Always`?

> For the first state of the resource, the Reference field was empty so filling this via controller is the late-initialization concept. However, after changed this value manually, this means that the user gives a new desired state. So I am not sure that we must override this.

I think this is pretty much the core use case users want with `Always` policy. In an ideal case, because the reference is changed by a controller, it would appear under `status` but then that reference field itself is actually a parameter. So, it's really hard to get both to work. Though personally, I feel better that it's configurable and default is consistent and IMO both policies will be used seldom.

apis/common/v1/resource.go
```diff
@@ -77,10 +77,48 @@ type SecretKeySelector struct {
...
+	// reference on every reconcile.
+	// +optional
+	Resolve *ResolvePolicy `json:"resolve,omitempty"`
+
+	// Resolution specifies whether resolution of this reference is required.
+	// The default is 'Required', which means the reconcile will fail if the
+	// reference cannot be resolved. 'Optional' means this reference will be
+	// a no-op if it cannot be resolved.
+	// +optional
+	// +kubebuilder:default=Required
```
muvaf, line 94: We can make this enum.

apis/common/v1/resource.go
```diff
@@ -77,10 +77,48 @@ type SecretKeySelector struct {
...
 }
 
+// Policy represents the Resolve and Resolution policies of Reference instance.
+type Policy struct {
+	// Resolve specifies when this reference should be resolved. The default
+	// is 'IfNotPresent', which will attempt to resolve the reference only when
+	// the corresponding field is not present. Use 'Always' to resolve the
+	// reference on every reconcile.
+	// +optional
+	Resolve *ResolvePolicy `json:"resolve,omitempty"`
```
muvaf, line 87: We can make this enum.

negz: Would it make sense for it to be possible to explicitly specify `IfNotPresent`? Currently that's not possible because the enum validation only allows `Always`.

muvaf: Yes, I think enum should include all possibilities. @sergenyalcin would you mind opening a PR to add this?

PR #337, opened 2022-06-22
 Added Injected Identity Source

### LINKED ISSUE #335: Support Injected Identity as a Common Credential Extractor
Would like to have the ability to use InjectedIdentity source type for an identity credential source. This would allow injected identities to be configured without needing to pass around credentials.

Add `xpv1.CredentialsSourceInjectedIdentity` as an option for `resource.CommonCredentialExtractor`. I was considering treating it the same as `xpv1.CredentialsSourceNone` and return `nil` and letting the provider determine what to do with `nil` creds.

bradkwadsworth-mw:
Signed-off-by: Brad Wadsworth <brad.wadsworth@mavenwave.com>

Add the ability to use InjectedIdentity as a CommonCredentialExtractor option. This will just return a nil value and will rely on the specific provider to take the appropriate action in order to fetch the injected credentials.

Fixes #335 

I have:

Tested with workload identity using a forked helm provider in a google GKE cluster.


djalanmwp:
Related code section:
./apis/common/v1/resource.go:236:       
`CredentialsSourceInjectedIdentity CredentialsSource = "InjectedIdentity"`

pkg/resource/providerconfig.go
```diff
@@ -95,6 +95,8 @@ func CommonCredentialExtractor(ctx context.Context, source xpv1.CredentialsSourc
 		return ExtractFs(ctx, afero.NewOsFs(), selector)
 	case xpv1.CredentialsSourceSecret:
 		return ExtractSecret(ctx, client, selector)
+	case xpv1.CredentialsSourceInjectedIdentity:
```
turkenh, line 98: Given we don't have to extract any credentials, in this case, we can check whether the source equals `xpv1.CredentialsSourceInjectedIdentity` before calling `CommonCredentialExtractor`. 
That approach makes it more explicit regarding the intent.

See [this](https://github.com/crossplane-contrib/provider-helm/blob/a51a31a6a4f8c9ce31503a0c48513b089cb2b929/pkg/controller/release/release.go#L171) as an example.

bradkwadsworth-mw:
Decided to only modify provider for InjectedIdentity.

PR #342, opened 2022-07-22
Upgrade to Go 1.18

epk:
Signed-off-by: Aditya Sharma <git@adi.run>

- Upgrades Go to 1.18
- Fixes some minor lint errors

I have:


pkg/fieldpath/paved.go
```diff
@@ -38,15 +38,15 @@ func (e errNotFound) IsNotFound() bool {
 // index was out of bounds in an array.
 func IsNotFound(err error) bool {
 	cause := errors.Cause(err)
-	_, ok := cause.(interface {
+	_, ok := cause.(interface { //nolint: errorlint // Skip errorlint for interface type
```
muvaf, line 41: Is there no way to use `any` here as well?

pkg/parser/parser.go
```diff
@@ -102,10 +102,10 @@ func (p *PackageParser) Parse(ctx context.Context, reader io.ReadCloser) (*Packa
 	do := json.NewSerializerWithOptions(json.DefaultMetaFactory, p.objScheme, p.objScheme, json.SerializerOptions{Yaml: true})
 	for {
 		bytes, err := yr.Read()
-		if err != nil && err != io.EOF {
+		if err != nil && !errors.Is(err, io.EOF) {
 			return pkg, err
 		}
-		if err == io.EOF {
+		if errors.Is(err, io.EOF) {
```
muvaf, line 108: Just checked that `errors.Is` returns the same answer for `== io.EOF` by running the tests with coverage and seeing it go into this statement.

muvaf (approved):
Thanks @epk !

pkg/fieldpath/paved.go
epk: `any` is only explicitly interoperable with `interface{}`, here it is interface with function definitions

muvaf: Do you know if there is a plan to make it available for use in these cases as well, or are we stuck with `interface{}` for the foreseeable future?

Issue #343, opened 2022-08-05
Deletion is missing in fieldpath library


muvaf:
### What problem are you facing?

There is no way to delete a field completely. You can only set its value to `nil`.

### How could Crossplane help solve your problem?

We can implement `DeleteField` that takes a path and deletes the object on that path.

PR #344, opened 2022-08-05
fieldpath: add `DeleteField` function to delete elements from Paved

muvaf:

Introduces `DeleteField` to delete a field from the given object.

Fixes https://github.com/crossplane/crossplane-runtime/issues/343

I have:

Unit tests with 100% coverage.

### DIFF
--- a/pkg/fieldpath/paved.go
+++ b/pkg/fieldpath/paved.go
@@ -471,3 +471,93 @@ func (p *Paved) SetBool(path string, value bool) error {
 func (p *Paved) SetNumber(path string, value float64) error {
 	return p.SetValue(path, value)
 }
+
+// DeleteField deletes the field from the object.
+// If the path points to an entry in an array, the element
+// on that index is removed and the next ones are pulled
+// back. If it is a field on a map, the field is
+// removed from the map.
+func (p *Paved) DeleteField(path string) error {
+	segments, err := Parse(path)
+	if err != nil {
+		return errors.Wrapf(err, "cannot parse path %q", path)
+	}
+	return p.delete(segments)
+}
+
+func (p *Paved) delete(segments Segments) error {
+	var in any = p.object
+	for i, current := range segments {
+		// final is true for the element before the last one because
+		// slices cannot be changed in place and Go does not allow
+		// taking address of map elements which prevents us from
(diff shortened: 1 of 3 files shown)


turkenh (approved):

muvaf:
I've noticed that two cases weren't handled and added them:
* Top-level field could not be removed since iteration assumed there is at least two segments, i.e. accessing `i+1`. Now we handle this as special case.
* In cases where the given path does not exist even before the last node, i.e. `metadata.some.another.other` where there is no `metadata.some`, it returned error. Now it's no-op since the target path is already gone.

Added unit tests for both but couldn't reduce cyclomatic complexity without breaking `p.delete` into pieces that don't make much sense.

@turkenh could you take another look?

pkg/fieldpath/paved.go
```diff
@@ -471,3 +471,113 @@ func (p *Paved) SetBool(path string, value bool) error {
 func (p *Paved) SetNumber(path string, value float64) error {
 	return p.SetValue(path, value)
 }
+
+// DeleteField deletes the field from the object.
+// If the path points to an entry in an array, the element
+// on that index is removed and the next ones are pulled
+// back. If it is a field on a map, the field is
+// removed from the map.
+func (p *Paved) DeleteField(path string) error {
```
muvaf, line 480: BTW, I could be convinced to rename this to `DeleteKey`.

turkenh (approved):

pkg/fieldpath/paved.go
turkenh: `Key` wouldn't be a good fit if array. What about `DeleteItem` ?

negz: Would simply `Delete` work? As in `paved.Delete` to delete whatever is at the supplied field path?

muvaf: @turkenh I feel like `Field` is more familiar than `Item` in this context for the users of fieldpath.

@negz I think it'd read nice if we had `Get`/`Set`/`Delete` but given that there is `GetValue`/`SetValue`, `paved.Delete` feels a bit off without a noun that says what you're deleting.

Thanks for the suggestions folks!

PR #347, opened 2022-08-25
UpdateCriticalAnnotations now supports namespace scope objects

### LINKED ISSUE #346: UpdateCriticalAnnotations does not support Namespace scope mange resources
### What happened?

When creating a managed resource in namespace scope instead of the default cluster scope, I'm facing lots of issues regarding cross-plane annotations. After some investigation, I noticed that the function [UpdateCriticalAnnotations](https://github.com/crossplane/crossplane-runtime/blob/d9c25ae277042c2ca2885fac24f74aec97349bc5/pkg/reconciler/managed/api.go#L194) does not support working with a managed resource who lives inside a namespace, this can be fixed easily by adding 
`namespace: o.GetNamespace()` to the NamespacedName struct. This would not break cluster-wide resources since it will use `""` for them

raz-bn:

Fixes #346

I have:


negz (changes requested):
I'm not sure about this. On one hand it wouldn't break anything to enable this, but on the other hand namespaced managed resources are explicitly not supported by Crossplane.

raz-bn:
> I'm not sure about this. On one hand it wouldn't break anything to enable this, but on the other hand namespaced managed resources are explicitly not supported by Crossplane.
@negz 
I wonder why would crossplane limit the use of namespaced managed resources. K8s Namespace give the ability to easily divide managed resources into separate projects while also will help promote multi tenant crossplane cluster. 
While I do not claim I completely understand the entire project code base, I don't see how supporting namespaced managed resources could break anything. 
For my use case namespaced managed resources are necessary, and I would be more then happy to discuss it farther more and also collaborate to make it happen 😊😊

raz-bn:
After reading more of the cross-plane doc, I saw I could achieve my goal using XRs.
However, it feels a bit wired. I need to create an XR for each managed resource I would like to exist in namespace scope.
At first glance, XR appeared to me as a “tool” to create resources that are based on two or more managed resources, and it makes a lot of sense. 
Using them as a “tool” to enable central cluster multi-tenancy feels redundant.
@negz 
@

negz:
You can read a little more about this in https://blog.crossplane.io/crossplane-vs-cloud-infrastructure-addons/.

Regardless of how folks might feel about the decision, Crossplane is not designed to support namespaced managed resources. Beginning to do so now would break a lot of assumptions (e.g. the one you found here) and increase the complexity of the project. I don't think it's something we're likely to support until v2, if at all.

PR #348, opened 2022-08-26
Ensure the stale workflow uses fewer operations

epk:
We only get 1,000 operations per hour for every workflow that uses
GITHUB_TOKEN. We're now seeing the stale workflow use relatively few
operations with few to no stale issues open.

In addition to lowering the workflow's operation limit this change has
the workflow run once a day. It adds a new `/fresh` comment command that
can be used to instantly mark an issue or PR as not stale, rather than
waiting up to a day for the stale workflow to run.

Signed-off-by: Aditya Sharma <git@adi.run>

Similar to https://github.com/crossplane/crossplane/pull/3258

I have:


negz:
Thanks @epk! Would you mind updating the PR description? I'm guessing this is what the commits you copied over from c/c say, but in this case we're adding net new actions and I'd like the PR description to reflect that.

negz:
FYI it looks like `/fresh` is broken per https://github.com/crossplane/crossplane/pull/3276

PR #353, opened 2022-09-06
feat(composite):  Add support for EnvironmentConfigs

MisterMX:

Adds interface and mock functions to allow storing references to EnvironmentConfigs in XRs (`spec.environmentConfigRefs`).

Ref. design document https://github.com/crossplane/crossplane/pull/3008.

I have:

n.A.

### DIFF
--- a/pkg/resource/unstructured/composite/composite.go
+++ b/pkg/resource/unstructured/composite/composite.go
@@ -218,3 +218,25 @@ func (c *Unstructured) GetConnectionDetailsLastPublishedTime() *metav1.Time {
 func (c *Unstructured) SetConnectionDetailsLastPublishedTime(t *metav1.Time) {
 	_ = fieldpath.Pave(c.Object).SetValue("status.connectionDetails.lastPublishedTime", t)
 }
+
+// GetEnvironmentConfigReferences of this Composite resource.
+func (c *Unstructured) GetEnvironmentConfigReferences() []corev1.ObjectReference {
+	out := &[]corev1.ObjectReference{}
+	_ = fieldpath.Pave(c.Object).GetValueInto("spec.environmentConfigRefs", out)
+	return *out
+}
+
+// SetEnvironmentConfigReferences of this Composite resource.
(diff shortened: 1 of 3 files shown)


pkg/resource/unstructured/composite/composite.go
```diff
@@ -218,3 +218,25 @@ func (c *Unstructured) GetConnectionDetailsLastPublishedTime() *metav1.Time {
...
+	return *out
+}
+
+// SetEnvironmentConfigReferences of this Composite resource.
+func (c *Unstructured) SetEnvironmentConfigReferences(refs []corev1.ObjectReference) {
+	empty := corev1.ObjectReference{}
+	filtered := make([]corev1.ObjectReference, 0, len(refs))
+	for _, ref := range refs {
+		// TODO(negz): Ask muvaf to explain what this is working around. :)
+		// TODO(muvaf): temporary workaround.
```
negz, line 235: I think @muvaf is on vacation, but I'd still love to know what this is doing. 😄

negz (approved):

pkg/resource/unstructured/composite/composite.go
muvaf: I think it was about having empty elements in the array causing problem since the element would be omitted but I'm not 100% sure.

pkg/resource/unstructured/composite/composite.go
```diff
@@ -218,3 +218,25 @@ func (c *Unstructured) GetConnectionDetailsLastPublishedTime() *metav1.Time {
...
+		if ref.String() == empty.String() {
+			continue
+		}
+		filtered = append(filtered, ref)
+	}
+	_ = fieldpath.Pave(c.Object).SetValue("spec.resourceRefs", filtered)
```
ytsarev, line 241: Shouldn't it set `spec.environmentConfigRefs` here instead?

MisterMX: Yes --> https://github.com/crossplane/crossplane-runtime/pull/365

PR #355, opened 2022-09-12
Support foreground cascading deletion

### LINKED ISSUE #354: Support for cascading foregroundDeletion
This is the crossplane-runtime tracking issue for the changes related to https://github.com/crossplane/crossplane/issues/3225

These changes will add support for cascading policy foregrounDeletion by adding a new Claim attribute "compositeDeletePolicy" which will default to Background but can be set to Foreground to cause the Claim reconciler to delete the associated Composite resource with the Foreground propagation policy attribute to Delete().

Note that due to the fact that Claims are Namespaced and Composites are not, the Composite resource cannot have an ownerReference that points at the Claim.  That means that it is not possible to trigger Foreground deletion via the kubectl --cascade=foreground command on the claim.  The workaround for this is to specify the compositeDeletePolicy: Foreground in the Claim to invoke the foreground cascading deletion on the associated Composite object and all of it's child objects.

These changes also do the following:
- set the blockOwnerDeletion flag to true on all controller ownerReferences
- update the managed Reconciler to not call the external Delete() function until all "external" finalizers have been removed from the managed resource

bobh66:
These changes are in support of https://github.com/crossplane/crossplane/issues/3225

These changes add support for Foreground Cascading Deletion by setting the blockOwnerDeletion flag to true on all controller ownerReferences and adding a new Claim attribute "compositeDeletePolicy".  This attribute defaults to Background but can be set to Foreground to cause the Claim reconciler to delete the associated Composite resource with the Foreground propagation policy attribute to Delete().

Note that due to the fact that Claims are Namespaced and Composites are not, the Composite resource cannot have an ownerReference that points at the Claim. That means that it is not possible to trigger Foreground deletion via the kubectl --cascade=foreground command on the claim. The workaround for this is to specify the compositeDeletePolicy: Foreground in the Claim to invoke the foreground cascading deletion on the associated Composite object and all of it's child objects.

Fixes #354 

I have:

- Created a Claim with the compositeDeletePolicy set to Foreground (using related crossplane changes)
- Deployed the claim/composite which deployed multiple managed resources
- Manually added a finalizer to one of the Managed resources
- Deleted the Claim and verified that the Composite and the Managed Resource with the manual finalizer added were not deleted as long as the manual finalizer is in place
- Removed the manual finalizer on the Managed Resource and verified that the MR was deleted first, followed by the Composite


pkg/resource/fake/mocks.go
```diff
@@ -204,6 +204,19 @@ func (m *CompositionUpdater) GetCompositionUpdatePolicy() *xpv1.UpdatePolicy {
 	return m.Policy
 }
 
+// CompositeDeleter is a mock that implements CompositeDeleter interface.
+type CompositeDeleter struct{ Policy *xpv1.CompositeDeletePolicy }
```
negz, line 208: Just thinking out loud here, but I'm wary of the amount of "machinery fields" we're injecting into claims and XRs - fields that pertain to Crossplane rather than to the API abstractions folks are trying to build. A little part of me wishes we'd either homed them all under a parent field (e.g. `spec.crossplane`) or leaned more into annotations.

Unfortunately I think that ship has sailed, and thus your approach of adding this new policy as a top-level spec field is the most idiomatic and consistent one I can think of.

negz (changes requested):
This mostly looks good - just needs a few small fixes.

pkg/meta/meta.go
```diff
@@ -114,6 +114,7 @@ func AsController(r *xpv1.TypedReference) metav1.OwnerReference {
 	c := true
 	ref := AsOwner(r)
 	ref.Controller = &c
+	ref.BlockOwnerDeletion = &c
```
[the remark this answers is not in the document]
bobh66: Yes - the only references to it's value are in the kubernetes garbage collection graph code, along with this:  https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/controller_utils.go#L510 which implies that it is required to be True for all controllers.

apis/common/v1/policies.go
```diff
@@ -31,6 +31,21 @@ const (
...
+const (
+	// CompositeDeleteBackground means the composite resource will be deleted using
+	// the Background Propagation Policy when the claim is deleted.
+	CompositeDeleteBackground CompositeDeletePolicy = "Background"
+
+	// CompositeDeleteForeground means the composite resource will be deleted using
+	// the Foreground Propagation Policy when the claim is deleted.
+	CompositeDeleteForeground CompositeDeletePolicy = "Foreground"
+)
+
```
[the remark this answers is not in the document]
bobh66: Apparently - not sure how that happened - I'll squash them.

pkg/meta/meta_test.go
```diff
@@ -176,11 +176,12 @@ func TestAsController(t *testing.T) {
...
-				Kind:       kind,
-				Name:       name,
-				UID:        uid,
-				Controller: &controller,
+				APIVersion:         groupVersion,
+				Kind:               kind,
+				Name:               name,
+				UID:                uid,
+				Controller:         &controller,
+				BlockOwnerDeletion: &controller,
```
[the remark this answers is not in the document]
bobh66: Updated variable name to "flag"

pkg/reconciler/managed/reconciler.go
```diff
@@ -662,12 +662,15 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
 
-	// If managed resource has a deletion timestamp and and a deletion policy of
+	// If managed resource has a deletion timestamp and a deletion policy of
 	// Orphan, we do not need to observe the external resource before attempting
 	// to unpublish connection details and remove finalizer.
 	if meta.WasDeleted(managed) && managed.GetDeletionPolicy() == xpv1.DeletionOrphan {
 		log = log.WithValues("deletion-timestamp", managed.GetDeletionTimestamp())
-
+		if len(managed.GetFinalizers()) > 1 {
+			log.Debug("Waiting for additional finalizers to be removed")
```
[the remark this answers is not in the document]
bobh66: Great idea - thanks.   I'll make the same changes for the equivalent code in the crossplane reconcilers

pkg/reconciler/managed/reconciler.go
```diff
@@ -792,6 +795,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	if meta.WasDeleted(managed) {
 		log = log.WithValues("deletion-timestamp", managed.GetDeletionTimestamp())
 
+		if len(managed.GetFinalizers()) > 1 {
+			log.Debug("Waiting for additional finalizers to be removed")
```
negz, line 799: Similar to my previous comment above, logging the number of remaining finalizers would be useful here.

pkg/reconciler/managed/reconciler.go
```diff
@@ -662,12 +662,15 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+	// If managed resource has a deletion timestamp and a deletion policy of
 	// Orphan, we do not need to observe the external resource before attempting
 	// to unpublish connection details and remove finalizer.
 	if meta.WasDeleted(managed) && managed.GetDeletionPolicy() == xpv1.DeletionOrphan {
 		log = log.WithValues("deletion-timestamp", managed.GetDeletionTimestamp())
-
+		if len(managed.GetFinalizers()) > 1 {
+			log.Debug("Waiting for additional finalizers to be removed", "finalizer_count", len(managed.GetFinalizers()))
+			return reconcile.Result{Requeue: true}, nil
+		}
```
negz, line 673: I wonder if there's a chance of a deadlock here. Probably quite unlikely?

I'm imagining some scenario in which an external system has added a finalizer that it won't remove until the underlying external resource is gone, the connection secret is gone, etc.

bobh66: Cyclic dependencies are always a possibility but I agree they should be quite unlikely.

pkg/reconciler/managed/reconciler.go
bobh66: This code was removed as agreed.

negz (approved):
Thanks @bobh66 - this looks great. Would you mind squashing your commits and updating the PR description to remove reference to the functionality we opted not to include:

> update the managed Reconciler to not call the external Delete() function until all "external" finalizers have been removed from the managed resource

bobh66:
> Thanks @bobh66 - this looks great. Would you mind squashing your commits and updating the PR description to remove reference to the functionality we opted not to include:
> 
> > update the managed Reconciler to not call the external Delete() function until all "external" finalizers have been removed from the managed resource

Done - thanks!

PR #357, opened 2022-09-28
Add support for a pause annotation which pauses reconciliations on managed resources

### LINKED ISSUE #351: A `crossplane.io/pause` annotation to stop reconciliation

There are cases where users want Crossplane or provider controllers to stop reconciling a certain resource such as migration from a CRD to another (like [here](https://github.com/crossplane-contrib/provider-aws/blob/master/cluster/UPGRADE.md)) or an in-cluster YAML surgery where you don't want to deal with racing conditions. Currently, such operations require creating a `ControllerConfig` for providers to scale down and then shutting down Crossplane by scaling it down. This method stops the world for every resource in the cluster. 

I think this was discussed somewhere but I couldn't find it.

We can have an controllar-runtime event filter to add to all reconcilers in runtime and core Crossplane to filter out all instances with the annotation `crossplane.io/pause: "true"`. It should be part of the Crossplane Resource Model (XRM) and documented [here](https://github.com/crossplane/crossplane/blob/master/design/one-pager-managed-resource-api-design.md).

ulucinar:

Fixes #351 

With this PR, if a managed resource has the `crossplane.io/paused` annotation with its value set to `true`, then the managed reconciler emits an event indicating that further reconciliations on that resource are paused and returns early after setting a `Synced` status condition with `False` status and `ReconcilePaused` reason. If the annotation is removed or its value is changed, reconciliation resumes.

I have:

Manually tested with the following resource manifest:
```yaml
apiVersion: ec2.aws.upbound.io/v1beta1
kind: VPC
metadata:
  name: paused
  annotations:
    crossplane.io/paused: "true"
spec:
  forProvider:
    cidrBlock: 192.168.0.0/16
    region: us-west-2
    tags:
      Name: alper-xp-351
  providerConfigRef:
    name: default
```

With the `crossplane.io/paused: "true"` annotation, the managed resource acquires the following status condition:
```yaml
status:
  atProvider: {}
  conditions:
  - lastTransitionTime: "2022-09-28T12:30:45Z"
    reason: ReconcilePaused
    status: "False"
    type: Synced
```
And the following associated events are emitted:
```
Events:
  Type    Reason                Age                  From                                          Message
  ----    ------                ----                 ----                                          -------
  Normal  ReconciliationPaused  2m3s (x2 over 2m3s)  managed/ec2.aws.upbound.io/v1beta1, kind=vpc  Reconciliation is paused via the pause annotation
```

If the value of the annotation is changed to `false`, reconciliation resumes and the status of the `Synced` status condition is updated:
```yaml
status:
  atProvider:
     ...
  conditions:
  - lastTransitionTime: "2022-09-28T12:34:30Z"
    reason: ReconcileSuccess
    status: "True"
    type: Synced
```

### DIFF
--- a/pkg/reconciler/managed/reconciler.go
+++ b/pkg/reconciler/managed/reconciler.go
@@ -664,9 +664,9 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		"external-name", meta.GetExternalName(managed),
 	)
 
-	// check the pause annotation and return if it has the value "true"
+	// Check the pause annotation and return if it has the value "true"
 	// after logging, publishing an event and updating the SYNC status condition
-	if managed.GetAnnotations()[meta.AnnotationKeyReconciliationPaused] == "true" {
+	if meta.IsPaused(managed) {
 		log.Debug("Reconciliation is paused via the pause annotation", "annotation", meta.AnnotationKeyReconciliationPaused, "value", "true")
 		record.Event(managed, event.Normal(reasonReconciliationPaused, "Reconciliation is paused via the pause annotation"))
 		managed.SetConditions(xpv1.ReconcilePaused())
--- a/pkg/reconciler/managed/reconciler.go
+++ b/pkg/reconciler/managed/reconciler.go
@@ -76,6 +76,8 @@ const (
 	reasonCreated event.Reason = "CreatedExternalResource"
 	reasonUpdated event.Reason = "UpdatedExternalResource"
 	reasonPending event.Reason = "PendingExternalResource"
+
+	reasonReconciliationPaused event.Reason = "ReconciliationPaused"
 )
 
 // ControllerName returns the recommended name for controllers that use this
@@ -661,6 +663,17 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		"version", managed.GetResourceVersion(),
 		"external-name", meta.GetExternalName(managed),
 	)
+
+	// check the pause annotation and return if it has the value "true"
+	// after logging, publishing an event and updating the SYNC status condition
+	if managed.GetAnnotations()[meta.AnnotationKeyReconciliationPaused] == "true" {
+		log.Debug("Reconciliation is paused via the pause annotation", "annotation", meta.AnnotationKeyReconciliationPaused, "value", "true")
+		record.Event(managed, event.Normal(reasonReconciliationPaused, "Reconciliation is paused via the pause annotation"))
+		managed.SetConditions(xpv1.ReconcilePaused())
+		// if the pause annotation is removed, we will have a chance to reconcile again and resume
+		// and if status update fails, we will reconcile again to retry to update the status
+		return reconcile.Result{}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
+	}
 
 	// If managed resource has a deletion timestamp and and a deletion policy of
 	// Orphan, we do not need to observe the external resource before attempting
--- a/apis/common/v1/condition.go
+++ b/apis/common/v1/condition.go
@@ -51,6 +51,7 @@ const (
 const (
 	ReasonReconcileSuccess ConditionReason = "ReconcileSuccess"
 	ReasonReconcileError   ConditionReason = "ReconcileError"
+	ReasonReconcilePaused  ConditionReason = "ReconcilePaused"
 )
 
 // A Condition that may apply to a resource.
@@ -248,5 +249,16 @@ func ReconcileError(err error) Condition {
 		LastTransitionTime: metav1.Now(),
 		Reason:             ReasonReconcileError,
 		Message:            err.Error(),
+	}
+}
+
+// ReconcilePaused returns a condition that indicates reconciliation on
+// the managed resource is paused via the pause annotation.
+func ReconcilePaused() Condition {
+	return Condition{
+		Type:               TypeSynced,
+		Status:             corev1.ConditionFalse,
+		LastTransitionTime: metav1.Now(),
+		Reason:             ReasonReconcilePaused,
 	}
 }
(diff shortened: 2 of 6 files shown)


pkg/reconciler/managed/reconciler.go
```diff
@@ -637,7 +640,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	// Be wary of adding additional complexity.
 
 	log := r.log.WithValues("request", req)
-	log.Debug("Reconciling")
+	log.Debug("Reconciling if not paused")
```
negz, line 643: I think for debug logging purposes it's safe to just leave this as it was. It will be quickly followed by a notice that reconcile is paused if that is the case.

pkg/reconciler/managed/reconciler.go
```diff
@@ -662,6 +665,21 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		"external-name", meta.GetExternalName(managed),
 	)
 
+	// check the pause annotation and return if it has the value "true"
+	// after logging, publishing an event and updating the SYNC status condition
+	if val, ok := managed.GetAnnotations()[meta.AnnotationKeyReconciliationPaused]; ok && val == "true" {
```
negz, line 670: Nit: I don't think there's much value in checking `ok` here. Annotations is a `map[string]string` so the default value will be `""` if the annotation doesn't exist, and we only proceed if the value is `"true"`.

pkg/reconciler/managed/reconciler.go
```diff
@@ -662,6 +665,21 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+		record.Event(managed, event.Normal(reasonReconciliationPaused, "Reconciliation is paused via the pause annotation"))
+		managed.SetConditions(xpv1.ReconcilePaused())
+		// if the pause annotation is removed, we will have a chance to reconcile again and resume
+		// and if status update fails, we will reconcile again to retry to update the status
+		return reconcile.Result{}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
+	}
+	if managed.GetCondition(xpv1.TypeSynced).Reason == xpv1.ReasonReconcilePaused {
+		log.Debug("Reconciliation resumes")
+		record.Event(managed, event.Normal(reasonReconciliationResumes, "Reconciliation resumes"))
+	}
```
negz, line 681: It's possible these could be emitted multiple times, e.g. if the following logic encountered an error before updating the synced condition. My preference would be to just omit this block - I think it would be pretty easy to infer that the resource was unpaused when it started emitting other events/statuses (create, update, etc).

pkg/reconciler/managed/reconciler.go
ulucinar: Done.

ulucinar:
Thanks @negz. I think we will also need a PR to update the Crossplane API conventions document if this gets merged.

pkg/reconciler/managed/reconciler.go
```diff
@@ -662,6 +664,17 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		"external-name", meta.GetExternalName(managed),
 	)
 
+	// check the pause annotation and return if it has the value "true"
+	// after logging, publishing an event and updating the SYNC status condition
+	if managed.GetAnnotations()[meta.AnnotationKeyReconciliationPaused] == "true" {
```
muvaf, line 669: It might be easier to read if we had a separate `IsPaused(cr) bool` function we call here.

pkg/reconciler/managed/reconciler.go
```diff
@@ -662,6 +664,17 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		"external-name", meta.GetExternalName(managed),
 	)
 
+	// check the pause annotation and return if it has the value "true"
```
muvaf, line 667: nitpick: The comments in the rest of the file starts with capital letter.

pkg/reconciler/managed/reconciler.go
```diff
@@ -662,6 +664,17 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+	// after logging, publishing an event and updating the SYNC status condition
+	if managed.GetAnnotations()[meta.AnnotationKeyReconciliationPaused] == "true" {
+		log.Debug("Reconciliation is paused via the pause annotation", "annotation", meta.AnnotationKeyReconciliationPaused, "value", "true")
+		record.Event(managed, event.Normal(reasonReconciliationPaused, "Reconciliation is paused via the pause annotation"))
+		managed.SetConditions(xpv1.ReconcilePaused())
```
muvaf, line 672: We should either remove the condition or set it to another value once the annotation is removed, which could be right after this condition block.

pkg/reconciler/managed/reconciler.go
ulucinar: Done.

pkg/reconciler/managed/reconciler.go
ulucinar: Done.

pkg/reconciler/managed/reconciler.go
ulucinar: Looks like if the reconciliation is not paused, `Synced` status condition either stays at the `False` state but this time with the reason `ReconcileError` or transitions to the state `True` with reason `ReconcileSuccess`. So the idea is if the reconciliation is paused via the annotation, we have `Synced` condition set to `False` with reason `ReconcilePaused`. We stay at this state (condition) as long as the annotation is set to `true`. Then if the annotation's value changes, then the managed resource acquires either the `Synced` condition set to `True` with the reason `ReconcileSuccess` or again the `Synced` condition set to `False` but this time with the reason `ReconcileError`. Do you think there is a possibility this transition would not happen (apart from a failing status update)?

muvaf: Oh OK, sorry, I thought this was a separate condition. All good, thank you for explanation!

muvaf (approved):
Thank you @ulucinar !

PR #358, opened 2022-09-28
Add a Server-Side Applicator

muvaf:

Adds the ability to utilize server-side apply mechanism by providers and core Crossplane. Unfortunately, `ApplyOption` argument is not relevant for this method since there is no fetching of an existing object. However, there are some patching options that users may want to set per-call, especially the owner. Initially, it was part of `NewAPIServerSideApplicator` function but that'd force you to use single owner for many calls, i.e. we'd have to use `composite` as owner for all managed resources. Now, we can actually use the composite name, which would be more accurate and helpful for debugging purposes.

I have:

Unit tests and basic usage test. I'll expand this section with more usage in composition.

### DIFF
--- a/pkg/resource/api.go
+++ b/pkg/resource/api.go
@@ -161,33 +161,38 @@ func (p *patch) Data(_ client.Object) ([]byte, error) { return json.Marshal(p.fr
 // API server to apply changes to given resource.
 type APIServerSideApplicator struct {
 	client client.Client
-	owner  string
 }
 
 // NewAPIServerSideApplicator returns a new APIServerSideApplicator.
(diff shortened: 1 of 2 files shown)


blakebarnett:
It looks like this is mostly ready to use, what else would be needed to start having providers take advantage of it? Maybe a `--enable-server-side-apply` option?

negz:
I'm going to close this. We are still investigating this avenue per https://github.com/crossplane/crossplane/issues/4047, but @muvaf hasn't touched this PR in a while and is no longer working on Crossplane full-time so I don't think we're likely to get this implementation merged.

PR #367, opened 2022-12-08
add support for float pointer references

### LINKED ISSUE #366: resource references only support string fields
### What happened?
Some terraform providers, such as the [Linode](https://registry.terraform.io/providers/linode/linode/latest/docs/resources/instance) provider use resource IDs that are not strings and are instead of type [int](https://github.com/linode/terraform-provider-linode/blob/dev/linode/instance/schema_resource.go). This causes an issue when trying to create references between resources because the [FromPtrValue](https://github.com/crossplane/crossplane-runtime/blob/master/pkg/reference/reference.go#L46-L51) and [ToPtrValue](https://github.com/crossplane/crossplane-runtime/blob/master/pkg/reference/reference.go#L54-L59) only support strings for references. 

### How can we reproduce it?
I have reproduced this issue in this repository where you can see the issue [here](https://github.com/eljohnson92/provider-linode/blob/incorrect_machine_id/apis/instance/v1alpha1/zz_generated.resolvers.go#L41). it is possible to overwrite the terraform schema for parts of the codebase to be a string instead of an int, but it does cause some problems when reconciling external resources so if it were possible for these functions to just support int/float references as well as strings. 

I am happy to try and contribute a fix to have ToPtrValue and ForPtrValue support more types if that is what is needed but please let me know if I might be missing something or if there is a more appropriate place to fix this issue.

crossplane version: [v0.19.0-rc.0](https://github.com/crossplane/crossplane-runtime/releases/tag/v0.19.0-rc.0)
upjet version: [v0.7.0](https://github.com/upbound/upjet/releases/tag/v0.7.0)

eljohnson92:
This change allows for types of references that are not strings. some terraform providers use "number" types for resource IDs which seem to get translated to float64 as part of upjet. when trying to use them as references previously only strings were supported, this PR along with a PR in crossplane-tools allows for "number" IDs to also be supported as references

Fixes #366 

I have:

- [ x ] Read and followed Crossplane's [contribution process].
- [ x ] Run `make reviewable test` to ensure this PR is ready for review.

This has been tested by adding unit tests to cover the new functions, and this code has also been tested by the [linode-provider](https://github.com/linode/provider-linode), which is using my forked version of this repo and the crossplane-tools repo

### DIFF
--- a/pkg/reference/reference.go
+++ b/pkg/reference/reference.go
@@ -18,6 +18,7 @@ package reference
 
 import (
 	"context"
+	"strconv"
 
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 
@@ -50,12 +51,32 @@ func FromPtrValue(v *string) string {
 	return *v
 }
 
+// FromFloatPtrValue adapts a float pointer field for use as a CurrentValue.
+func FromFloatPtrValue(v *float64) string {
+	if v == nil {
+		return ""
+	}
+	return strconv.FormatFloat(*v, 'f', 0, 64)
+}
+
 // ToPtrValue adapts a ResolvedValue for use as a string pointer field.
 func ToPtrValue(v string) *string {
 	if v == "" {
 		return nil
 	}
 	return &v
+}
+
+// ToFloatPtrValue adapts a ResolvedValue for use as a float64 pointer field.
+func ToFloatPtrValue(v string) *float64 {
+	if v == "" {
+		return nil
+	}
+	vParsed, err := strconv.ParseFloat(v, 64)
+	if err != nil {
+		return nil
+	}
+	return &vParsed
 }
 
 // FromPtrValues adapts a slice of string pointer fields for use as CurrentValues.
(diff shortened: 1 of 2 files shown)


eljohnson92:
@negz is this something you would consider merging or have other workarounds we could adopt?

jeanduplessis:
@ezgidemirel @phisco would you mind doing a review here before we pass on to a maintainer

phisco (commented):
Not super sure about the naming, the FromPtrValue function is just dereferencing, while the FromFloatPtrValue one is also formatting it to string while the ToFloatPtrValue is parsing it and hiding the error.

eljohnson92:
@phisco if it is just a matter of variable naming I'm happy to rename the variable. let me know if there is additionally any functionality you think I should change here

phisco (approved):
At second thought I think it could be ok with the current names. Maybe @negz or @turkenh can have a look at it next and get it merged.

turkenh (approved):

PR #372, opened 2023-01-23
Add DesiredStateChanged to Filter out Updates on Status

turkenh:

This PR adds a new predicate function named `DesiredStateChanged()` which does nothing but combines (with a `predicate.Or`) the following controller-runtime predicates:

- [GenerationChangedPredicate](https://github.com/kubernetes-sigs/controller-runtime/blob/9241bceb3098499d7243fa6aee0cd6930bfc03e8/pkg/predicate/predicate.go#L156)
  > // GenerationChangedPredicate implements a default update predicate function on Generation change.
  > //
  > // This predicate will skip update events that have no change in the object's metadata.generation field.
  > // The metadata.generation field of an object is incremented by the API server when writes are made to the spec field of an object.
  > // This allows a controller to ignore update events where the spec is unchanged, and only the metadata and/or status fields are changed.
  > //
  > // For CustomResource objects the Generation is only incremented when the status subresource is enabled.
  > //
  > // Caveats:
  > //
  > // * The assumption that the Generation is incremented only on writing to the spec does not hold for all APIs.
  > // E.g For Deployment objects the Generation is also incremented on writes to the metadata.annotations field.
  > // For object types other than CustomResources be sure to verify which fields will trigger a Generation increment when they are written to.
  > //
  > // * With this predicate, any update events with writes only to the status field will not be reconciled.
  > // So in the event that the status block is overwritten or wiped by someone else the controller will not self-correct to restore the correct status.

- [AnnotationChangedPredicate](https://github.com/kubernetes-sigs/controller-runtime/blob/9241bceb3098499d7243fa6aee0cd6930bfc03e8/pkg/predicate/predicate.go#L186)
  > // AnnotationChangedPredicate implements a default update predicate function on annotation change.
  > //
  > // This predicate will skip update events that have no change in the object's annotation.
  > // It is intended to be used in conjunction with the GenerationChangedPredicate, as in the following example:
  > //
  > //	Controller.Watch(
  > //		&source.Kind{Type: v1.MyCustomKind},
  > //		&handler.EnqueueRequestForObject{},
  > //		predicate.Or(predicate.GenerationChangedPredicate{}, predicate.AnnotationChangedPredicate{}))
  > //
  > // This is mostly useful for controllers that needs to trigger both when the resource's generation is incremented
  > // (i.e., when the resource' .spec changes), or an annotation changes (e.g., for a staging/alpha API).

- [LabelChangedPredicate](https://github.com/kubernetes-sigs/controller-runtime/blob/9241bceb3098499d7243fa6aee0cd6930bfc03e8/pkg/predicate/predicate.go#L217)
  > // LabelChangedPredicate implements a default update predicate function on label change.
  > //
  > // This predicate will skip update events that have no change in the object's label.
  > // It is intended to be used in conjunction with the GenerationChangedPredicate, as in the following example:
  > //
  > // Controller.Watch(
  > //
  > //	&source.Kind{Type: v1.MyCustomKind},
  > //	&handler.EnqueueRequestForObject{},
  > //	predicate.Or(predicate.GenerationChangedPredicate{}, predicate.LabelChangedPredicate{}))
  > //
  > // This will be helpful when object's labels is carrying some extra specification information beyond object's spec,
  > // and the controller will be triggered if any valid spec change (not only in spec, but also in labels) happens.

Fixes https://github.com/crossplane/crossplane/issues/3455

I have:

See https://github.com/crossplane-contrib/provider-aws/pull/1645 to see how this can be consumed by provider controllers and also the result of testing.

### DIFF
--- a/pkg/resource/predicates.go
+++ b/pkg/resource/predicates.go
@@ -146,3 +146,17 @@ func IsNamed(name string) PredicateFn {
 		return mo.GetName() == name
 	}
 }
+
+// DesiredStateChanged accepts objects that have changed their desired state, i.e.
+// the state that is not managed by the controller.
+// To be more specific, it accepts update events that have changes in one of the followings:
+// - `metadata.annotations`
+// - `metadata.labels`
+// - `spec`
+func DesiredStateChanged() predicate.Predicate {
+	return predicate.Or(
+		predicate.AnnotationChangedPredicate{},
+		predicate.LabelChangedPredicate{},
+		predicate.GenerationChangedPredicate{},
+	)
+}
--- a/pkg/resource/predicates.go
+++ b/pkg/resource/predicates.go
@@ -27,10 +27,14 @@ import (
 )
 
 // A PredicateFn returns true if the supplied object should be reconciled.
+// Deprecated: This type will be removed soon. Please use
+// controller-runtime's predicate.NewPredicateFuncs instead.
 type PredicateFn func(obj runtime.Object) bool
 
 // NewPredicates returns a set of Funcs that are all satisfied by the supplied
 // PredicateFn. The PredicateFn is run against the new object during updates.
+// Deprecated: This function will be removed soon. Please use
+// controller-runtime's predicate.NewPredicateFuncs instead.
 func NewPredicates(fn PredicateFn) predicate.Funcs {
 	return predicate.Funcs{
 		CreateFunc:  func(e event.CreateEvent) bool { return fn(e.Object) },
@@ -41,6 +45,8 @@ func NewPredicates(fn PredicateFn) predicate.Funcs {
 }
 
 // AnyOf accepts objects that pass any of the supplied predicate functions.
+// Deprecated: This function will be removed soon. Please use
+// controller-runtime's predicate.Or instead.
 func AnyOf(fn ...PredicateFn) PredicateFn {
 	return func(obj runtime.Object) bool {
 		for _, f := range fn {
@@ -53,6 +59,8 @@ func AnyOf(fn ...PredicateFn) PredicateFn {
 }
 
 // AllOf accepts objects that pass all of the supplied predicate functions.
+// Deprecated: This function will be removed soon. Please use
+// controller-runtime's predicate.And instead.
 func AllOf(fn ...PredicateFn) PredicateFn {
 	return func(obj runtime.Object) bool {
 		for _, f := range fn {
@@ -66,6 +74,7 @@ func AllOf(fn ...PredicateFn) PredicateFn {
 
 // HasManagedResourceReferenceKind accepts objects that reference the supplied
 // managed resource kind.
+// Deprecated: This function will be removed soon.
 func HasManagedResourceReferenceKind(k ManagedKind) PredicateFn {
 	return func(obj runtime.Object) bool {
 		r, ok := obj.(ManagedResourceReferencer)
@@ -82,6 +91,7 @@ func HasManagedResourceReferenceKind(k ManagedKind) PredicateFn {
 }
 
 // IsManagedKind accepts objects that are of the supplied managed resource kind.
+// Deprecated: This function will be removed soon.
 func IsManagedKind(k ManagedKind, ot runtime.ObjectTyper) PredicateFn {
 	return func(obj runtime.Object) bool {
 		gvk, err := GetKind(obj, ot)
@@ -94,6 +104,7 @@ func IsManagedKind(k ManagedKind, ot runtime.ObjectTyper) PredicateFn {
 
 // IsControlledByKind accepts objects that are controlled by a resource of the
 // supplied kind.
+// Deprecated: This function will be removed soon.
 func IsControlledByKind(k schema.GroupVersionKind) PredicateFn {
 	return func(obj runtime.Object) bool {
 		mo, ok := obj.(metav1.Object)
@@ -112,6 +123,7 @@ func IsControlledByKind(k schema.GroupVersionKind) PredicateFn {
 
 // IsPropagator accepts objects that request to be partially or fully propagated
 // to another object of the same kind.
+// Deprecated: This function will be removed soon.
 func IsPropagator() PredicateFn {
 	return func(obj runtime.Object) bool {
 		from, ok := obj.(metav1.Object)
@@ -125,6 +137,7 @@ func IsPropagator() PredicateFn {
 
 // IsPropagated accepts objects that consent to be partially or fully propagated
 // from another object of the same kind.
+// Deprecated: This function will be removed soon.
 func IsPropagated() PredicateFn {
 	return func(obj runtime.Object) bool {
 		to, ok := obj.(metav1.Object)
@@ -137,6 +150,7 @@ func IsPropagated() PredicateFn {
 }
 
 // IsNamed accepts objects that is named as the given name.
+// Deprecated: This function will be removed soon.
 func IsNamed(name string) PredicateFn {
 	return func(obj runtime.Object) bool {
 		mo, ok := obj.(metav1.Object)
--- a/pkg/resource/predicates_test.go
+++ b/pkg/resource/predicates_test.go
@@ -24,7 +24,10 @@ import (
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 
+	runtimev1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 	"github.com/crossplane/crossplane-runtime/pkg/errors"
 	"github.com/crossplane/crossplane-runtime/pkg/meta"
 	"github.com/crossplane/crossplane-runtime/pkg/resource/fake"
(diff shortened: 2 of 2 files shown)


negz:
I'm not sure about this. Two concerns come to mind:

* This seems to exclude most `metadata` fields. Do we need to queue a reconcile when the deletion timestamp changes, or finalizers, owner references, etc?
* Do we have cases in the managed resource reconciler where we're assuming that updating the status will trigger an immediate requeue? I vaguely recall we do.

liubog2008:
> This seems to exclude most metadata fields. Do we need to queue a reconcile when the deletion timestamp changes, or finalizers, owner references, etc?

Don't need. If we update finalizers, we can just return a result with requeue flag in `Reconcile`. Normally only annotations and labels in the metadata will be updated by the third-party.

turkenh:
@liubog2008 has a point. I believe we can act based on what is expected to be updated by users under `metadata`, and for others, we can tweak our controller logic. For example, I am already noticing that we have an [explicit Requeue](https://github.com/crossplane/crossplane-runtime/blob/1fc35a41fccae56fdc26f075e737f4a7dc4a7a58/pkg/reconciler/managed/reconciler.go#L834) in the managed reconciler after the removal of the finalizer. 

> Do we have cases in the managed resource reconciler where we're assuming that updating the status will trigger an immediate requeue? I vaguely recall we do.

I see in the comments of managed reconciler we have such [assumptions already](https://github.com/crossplane/crossplane-runtime/blob/1fc35a41fccae56fdc26f075e737f4a7dc4a7a58/pkg/reconciler/managed/reconciler.go#L785), but I believe we can avoid them by making them explicit if needed.

turkenh:
> Do we have cases in the managed resource reconciler where we're assuming that updating the status will trigger an immediate requeue? I vaguely recall we do.

I found such assumptions in the upjet, which would need to be removed if we decide to go with this approach. After an async operation is completed, [upjet relies on status updates](https://github.com/upbound/upjet/blob/615535f51088f690c02bbe0171c9a56d25acb45d/pkg/controller/api.go#L89) triggering a new reconciliation to reflect the state on the MR (e.g. marking resource as ready).

negz:
I feel okay with proceeding with this approach, but I would like to see it _extensively_ tested first. I'm concerned it may break assumptions we have in the managed resource reconciler. Historically we've found changes to the managed resource reconciler's requeue decisions too often introduced subtle bugs, and changes to this reconciler have a very broad impact.

One thing I don't understand - how do deletes work with this predicate? When we delete an MR the API Server sets the deletion timestamp in metadata. Which predicate includes such events?

turkenh:
> I feel okay with proceeding with this approach, but I would like to see it _extensively_ tested first. I'm concerned it may break assumptions we have in the managed resource reconciler. Historically we've found changes to the managed resource reconciler's requeue decisions too often introduced subtle bugs, and changes to this reconciler have a very broad impact.

Agreed, trying to come up with some scenarios that would make us feel comfortable. One note, though, this PR is just adding a predicate that still need to be consumed by individual controller setup calls like [this](https://github.com/crossplane-contrib/provider-aws/pull/1645/files#diff-3627e10d7274fe2d160686035f4b9b9b3de8d3e40659773984bbf0a10ca14149R69). So, unless we need to make changes in the managed reconciler logic after the experiments, this wouldn't be something that we wouldn't need to roll out altogether and do that gradually.

> One thing I don't understand - how do deletes work with this predicate? When we delete an MR, the API Server sets the deletion timestamp in metadata. Which predicate includes such events?

Kube API server explicitly [bumps generation](https://github.com/kubernetes/apiserver/blob/e289f0d9a0af88cd0c19d53fe0fd07b407f0212c/pkg/registry/rest/delete.go#L155) in case of a delete, so it is part of the `GenerationChangedPredicate`. I've just tested and validated this.

turkenh:
I have conducted the experiments below and observed no issues or noticeable behavior changes. I have used the native provider-aws not to interfere with another layer. FWIW, we already know we must fix [this issue](https://github.com/crossplane/crossplane-runtime/pull/372#issuecomment-1438328103) in upjet before consuming this change. 

@negz please let me know if there are other scenarios in your mind that worths testing.

### Experiment 1 - Happy Path CRUD of an AWS Bucket

With `crossplane-contrib/provider-aws`, create the following bucket

```yaml
apiVersion: s3.aws.crossplane.io/v1beta1
kind: Bucket
metadata:
  name: test-bucket
  annotations:
    crossplane.io/external-name: crossplane-example-bucket-by-hasan
spec:
  forProvider:
    acl: private
    locationConstraint: us-east-1
### Update with below
#    lifecycleConfiguration:
#      rules:
#        - status: Enabled
#          filter:
#            prefix: "test/"
#          expiration:
#            days: 15
```

No issues or differences observed (no predicate vs with predicate) with none of CRUD operation on the happy path. ✅ 

### Experiment 2 - Delete a non-empty AWS Bucket

**No Predicate**

Screen Shot 2023-03-07 at 17 16 55

**With Predicate**

Screen Shot 2023-03-07 at 17 24 45

### Experiment 3 - Reference Resolution - 3 MRs referencing on each other

```yaml
kind: VPC
metadata:
  name: sample-vpc
spec:
  forProvider:
    region: us-east-1
    cidrBlock: 10.0.0.0/16
    enableDnsSupport: true
    enableDnsHostNames: true
    instanceTenancy: default
---
kind: SecurityGroup
metadata:
  name: sample-cluster-sg
spec:
  forProvider:
    region: us-east-1
    vpcIdRef:
      name: sample-vpc
    groupName: my-cool-ekscluster-sg
    description: Cluster communication with worker nodes
    ingress:
      - fromPort: 80
        toPort: 80
        ipProtocol: tcp
        ipRanges:
          - cidrIp: 10.0.0.0/8
---
apiVersion: ec2.aws.crossplane.io/v1alpha1
kind: SecurityGroupRule
metadata:
  name: sample-sgr
spec:
  forProvider:
    region: us-east-1
    protocol: "tcp"
    fromPort: 443
    toPort: 443
    type: "ingress"
    securityGroupIdRef:
      name: sample-cluster-sg
    cidrBlock: "123.12.0.0/16" 
    description: "sample-sgr description"
```

**No Predicate**

All was ready within ~15 secs.

**With Predicate**

While VPC and SecurityGroupRule became ready within seconds, SecurityGroup stuck as READY=false (waited about 15mins). It become immediately READY after I've added a random label on it (to force reconcile).

Debugging it further I found it because of https://github.com/crossplane-contrib/provider-aws/issues/1516

In `No Predicate` case, this was passing by chance due to a different order of reconciliation where SecurityGroup could mark READY at least once before SecurityGroupRule creating a rule and causing [a diff](https://github.com/crossplane-contrib/provider-aws/blob/04858ec0d42823197b53894b3993ce85ba2f9885/pkg/controller/ec2/securitygroup/controller.go#L152) for SecurityGroup.

I've set `spec.forProvider.ignoreIngress=true` for `SecurityGroup` and retried the experiments. This time all 3 resources were ready in ~10 secs.

I also have a relatively subjective observation that, we are hitting the infamous `the object has been modified` error less frequently when we have predicates (which is kind of expected) and this way we have slightly faster provisioning (i.e. time to readiness). I didn't spend time to do enough controlled experiments to support this though.

### Experiment 4 - Compositions owning multiple MRs

As a complex enough experiment, I have used [this example](https://github.com/awslabs/crossplane-on-eks/blob/main/examples/aws-provider/composite-resources/vpc-subnets-eks/vpc-subnets-eks-claims.yaml) from [awslabs/crossplane-on-eks](https://github.com/awslabs/crossplane-on-eks) which contains 2x claims composing the following Managed resources:

```bash
2x routetable.ec2.aws.crossplane.io
1x internetgateway.ec2.aws.crossplane.io
1x natgateway.ec2.aws.crossplane.io
1x vpc.ec2.aws.crossplane.io
1x address.ec2.aws.crossplane.io
6x subnet.ec2.aws.crossplane.io
1x cluster.eks.aws.crossplane.io
1x nodegroup.eks.aws.crossplane.io
2x role.iam.aws.crossplane.io
7x rolepolicyattachment.iam.aws.crossplane.io
```

**No Predicate**

All was ready in ~11 mins.

**With Predicate**

All was ready in ~11 mins.

Observed no issues or differences between with or without predicates.

liubog2008:
Any progress about this PR now? @negz @jbw976

pkg/resource/predicates.go
```diff
@@ -146,3 +146,17 @@ func IsNamed(name string) PredicateFn {
...
 	}
 }
+
+// DesiredStateChanged accepts objects that have changed their desired state, i.e.
+// the state that is not managed by the controller.
+// To be more specific, it accepts update events that have changes in one of the followings:
+// - `metadata.annotations`
+// - `metadata.labels`
+// - `spec`
+func DesiredStateChanged() predicate.Predicate {
```
negz, line 156: I notice almost everything else in this package returns our `PredicateFn` type, whereas this returns an upstream `predicate.Predicate` interface. It seems like all three of the `predicate.Or`-ed predicates only implement `Update`.

I'm wondering whether it makes sense for this to return a `PredicateFn` for consistency/compatibility with the rest of this package.

Alternatively, I think a lot of what this package does has been implemented upstream since we wrote it. For example `predicate.AnyOf` seems about the same as `predicate.Or`. Perhaps we should be deprecating much of this package in favor of upstream functionality?

negz (commented):
Thanks for taking the time to think through the edge cases here @turkenh. One comment to consider, but I'm happy adding this predicate.

pkg/resource/predicates.go
turkenh: > I'm wondering whether it makes sense for this to return a PredicateFn for consistency/compatibility with the rest of this package.

This was something I've considered in the beginning; however, it looks like this is not possible since the predicates we are adding need both [the old object and new object](https://github.com/kubernetes-sigs/controller-runtime/blob/97acd6444ff1ec87caf7d2077a9056ea56b6ba0e/pkg/predicate/predicate.go#L171) (e.g. to see if generation changed) but the `PredicateFn` only receives a single object (i.e. `ObjectNew`).

> Alternatively, I think a lot of what this package does has been implemented upstream since we wrote it. For example predicate.AnyOf seems about the same as predicate.Or. Perhaps we should be deprecating much of this package in favor of upstream functionality?

I see some potential value in having this specific predicate in the runtime:

- Instead of expecting each provider to combine the three in the same way, they can rely on what runtime has.
- We may want to change things under the hood on how we define the "desired state" for a Crossplane MR. A good example of this is one of the proposed solutions [here](https://github.com/crossplane/crossplane/issues/3840#issuecomment-1461758879). If we decide that some specific annotation should not be part of the desired state, we can add exceptions in this central place.

Regarding the deprecation of the rest, I wanted to see how they are used today and searched them one by one using a query as below and couldn't find any usage for any of them: 

https://github.com/search?q=org%3Acrossplane+OR+org%3Aupbound+OR+org%3Acrossplane-contrib+IsManagedKind&type=code

Am I missing something, or are they not used at all already?

negz: > I see some potential value in having this specific predicate in the runtime

I agree - to be clear I was only suggesting removing (most of?) the _other_ stuff in this package. Not the new predicate.

I'm a bit surprised the other stuff aren't used, but I trust your search. Maybe they haven't been used since the old resource claim reconcilers or something? Given they aren't used, would you mind adding `// Deprecated` comments to them in a separate commit? Then we can go through and remove a bunch of deprecated things next release.

turkenh: And I trust Github's search :) 

Done. 
Some have a single usage, like [IsNamed](https://github.com/upbound/universal-crossplane/blob/7403c2fe233e5335fe1251a031df36399abd1d1e/internal/controllers/billing/setup.go#L54), and I still added deprecation notice since it can live in the relevant codebase, no need to have shared.

negz (approved):

PR #375, opened 2023-01-24
Bump Kubernetes dependencies to latest

negz:

This PR also updates the linter to bring it in line with c/c. I had to do this because my Go version was too new for the golangci-lint version this repo was using.

I have:

Opening as draft for now - I'd like to test by building c/c against this repo.


pkg/connection/store/vault/store.go
```diff
@@ -169,7 +170,7 @@ func (ss *SecretStore) DeleteKeyValues(_ context.Context, s *store.Secret, do ..
 	}
 
 	for _, o := range do {
-		if err = o(context.Background(), s); err != nil {
+		if err = o(ctx, s); err != nil {
```
negz, line 173: @turkenh I presume it's safe to pass this context on here. Was not passing it intentional?

pkg/connection/store/vault/store.go
```diff
@@ -211,7 +212,7 @@ func applyOptions(wo ...store.WriteOption) []kv.ApplyOption {
 				},
 				Data: keyValuesFromData(desired.Data),
 			}
-			if err := o(context.Background(), cs, ds); err != nil {
+			if err := o(ctx, cs, ds); err != nil {
```
negz, line 215: @turkenh Ditto here - I presume this is safe to pass?

pkg/connection/manager.go
```diff
@@ -142,7 +143,7 @@ func (m *DetailsManager) FetchConnection(ctx context.Context, so resource.Connec
 }
 
 // PropagateConnection propagate connection details from one resource to another.
-func (m *DetailsManager) PropagateConnection(ctx context.Context, to resource.LocalConnectionSecretOwner, from resource.ConnectionSecretOwner) (propagated bool, err error) { // nolint:interfacer
+func (m *DetailsManager) PropagateConnection(ctx context.Context, to resource.LocalConnectionSecretOwner, from resource.ConnectionSecretOwner) (propagated bool, err error) { //nolint:interfacer // See note below.
```
phisco, line 146: Interfacer has been deprecated, see https://github.com/mvdan/interfacer and golangcilint-list, I'd say it could be even dropped.

pkg/resource/unstructured/client.go
```diff
@@ -54,7 +54,7 @@ type WrapperClient struct {
 // Get retrieves an obj for the given object key from the Kubernetes Cluster.
 // obj must be a struct pointer so that obj can be updated with the response
 // returned by the Server.
-func (c *WrapperClient) Get(ctx context.Context, key client.ObjectKey, obj client.Object) error {
+func (c *WrapperClient) Get(ctx context.Context, key client.ObjectKey, obj client.Object, opts ...client.GetOption) error {
```
phisco, line 57: Is it intentional that you are not using opts?

negz: No, good catch.

pkg/connection/manager.go
negz: Hrm, I would expect `nolintlint` to complain if this statement was ineffective.

phisco: Nono, it's effective as it's still integrated into golangci-lint, but the original project was archived suggesting it's not so useful, which I kind of agree with

pkg/connection/store/vault/store.go
turkenh: I don’t remember that it was intentional, so looks good.

pkg/connection/store/vault/store.go
turkenh: 👍

pkg/connection/manager.go
negz: Nope, removed the nolint directive and did not get a linter warning. I've removed it.

hasheddan (approved):

PR #379, opened 2023-02-02
External Secret Store API and Client Implementation

ezgidemirel:

Fixes #

I have:

- [ ] Read and followed Crossplane's [contribution process].


pkg/connection/store/external/store.go
```diff
@@ -0,0 +1,186 @@
...
+		Certificates: []tls.Certificate{certificate},
+		RootCAs:      pool,
+	}
+
+	return credentials.NewTLS(tlsConfig), nil
+}
+
+// NewSecretStore returns a new External SecretStore.
+func NewSecretStore(ctx context.Context, kube client.Client, cfg v1.SecretStoreConfig) (*SecretStore, error) {
+	kp, err := loadKeyPair()
```
turkenh, line 91: I assume this client will use the same certs no matter which plugin server it talks to (when there are more than once). Should we move `loadKeyPair` to a previous stage (e.g., startup) so that we wouldn't need to do it each time we want to build a secret store, which is expected to happen in each reconcile?

pkg/connection/store/external/store.go
```diff
@@ -0,0 +1,186 @@
...
+		kubeClient: kube,
+		client:     cl,
+		config:     cfg.External.ConfigRef,
+	}, nil
+}
+
+// ReadKeyValues reads and returns key value pairs for a given Secret.
+func (ss *SecretStore) ReadKeyValues(ctx context.Context, n store.ScopedName, s *store.Secret) error {
+	sec := new(ess.Secret)
+	sec.ScopedName = n.Name
```
turkenh, line 113: Hmm, should we somehow include `n.Scope` in `sec.ScopedName` ?

pkg/connection/store/external/store.go
```diff
@@ -0,0 +1,186 @@
...
+
+	res, err := ss.client.GetSecret(ctx, &ess.GetSecretRequest{Secret: sec, Config: cfg})
+	if err != nil {
+		return errors.Wrap(err, errGet)
+	}
+
+	s.Data = make(map[string][]byte, len(res.Secret.Data))
+	for d := range res.Secret.Data {
+		res.Secret.Data[d] = s.Data[d]
+	}
```
turkenh, line 125: Shouldn't we fill `s.Data` ?

```suggestion
	for k, v := range res.Secret.Data {
		s.Data[k] = v
	}
```

pkg/connection/store/external/store.go
```diff
@@ -0,0 +1,186 @@
...
+
+	if res.Secret.Metadata != nil {
+		s.Metadata.Labels = make(map[string]string, len(res.Secret.Metadata))
+		for m := range res.Secret.Metadata {
+			res.Secret.Metadata[m] = s.Metadata.Labels[m]
+		}
```
turkenh, line 131: Similarly:

```suggestion
		for k, v := range res.Secret.Metadata {
			s.Metadata.Labels[k] = v
		}
```

pkg/connection/store/external/store.go
```diff
@@ -0,0 +1,186 @@
...
+	if err != nil {
+		return errors.Wrap(err, errGet)
+	}
+
+	s.Data = make(map[string][]byte, len(res.Secret.Data))
+	for d := range res.Secret.Data {
+		res.Secret.Data[d] = s.Data[d]
+	}
+
+	if res.Secret.Metadata != nil {
```
turkenh, line 127:
```suggestion
	if len(res.Secret.Metadata) > 0 {
```

pkg/connection/store/external/store.go
```diff
@@ -0,0 +1,186 @@
...
+func (ss *SecretStore) WriteKeyValues(ctx context.Context, s *store.Secret, wo ...store.WriteOption) (changed bool, err error) {
+	sec := new(ess.Secret)
+	sec.ScopedName = s.Name
+
+	cfg := ss.getConfigReference()
+
+	sec.Data = make(map[string][]byte, len(s.Data))
+	for d := range s.Data {
+		sec.Data[d] = s.Data[d]
+	}
```
turkenh, line 147: nit:

```suggestion
	for k, v := range s.Data {
		sec.Data[k] = v
	}
```

pkg/connection/store/external/store.go
```diff
@@ -0,0 +1,186 @@
...
+
+	if sec.Metadata != nil {
+		sec.Metadata = make(map[string]string, len(s.Metadata.Labels))
+		for m := range s.Metadata.Labels {
+			sec.Metadata[m] = s.Metadata.Labels[m]
+		}
```
turkenh, line 153: nit:

```suggestion
		for k, v := range s.Metadata.Labels {
			sec.Metadata[k] = v
		}
```

pkg/connection/store/external/store.go
```diff
@@ -0,0 +1,186 @@
...
+	sec.ScopedName = s.Name
+
+	cfg := ss.getConfigReference()
+
+	sec.Data = make(map[string][]byte, len(s.Data))
+	for d := range s.Data {
+		sec.Data[d] = s.Data[d]
+	}
+
+	if sec.Metadata != nil {
```
turkenh, line 149:
```suggestion
	if len(s.Metadata.Labels) > 0 {
```

apis/proto/v1alpha1/ess.proto
```diff
@@ -0,0 +1,61 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+syntax = "proto3";
+
+option go_package = "ess-grpc-2/proto/ess";
```
negz, line 16: I believe `buf` has a linter you can run against this file. I know it has some opinions about how `go_package` should be derived from the directory name.

apis/proto/v1alpha1/ess.proto
```diff
@@ -0,0 +1,61 @@
...
+
+// This service defines the APIs for an External Secret Store plugin.
+service ExternalSecretStoreService {
+  rpc GetSecret(GetSecretRequest) returns (GetSecretResponse) {}
+  rpc ApplySecret(ApplySecretRequest) returns (ApplySecretResponse) {}
+  rpc DeleteKeys(DeleteKeysRequest) returns (DeleteKeysResponse) {}
+}
+
+message ConfigReference {
```
negz, line 25: I'm not sure if the linter will require doc strings for messages/fields, but I think they would useful documentation for developers building secret stores.

apis/proto/v1alpha1/ess.proto
```diff
@@ -0,0 +1,61 @@
...
+  string api_version = 1;
+  string kind = 2;
+  string name = 3;
+}
```
negz, line 29: Is the idea that the plugin will connect to Kubernetes to read the config? Perhaps it would make sense to have plugins be unaware of Kubernetes - i.e. to have the caller read the config and pass it to the plugin via gRPC.

apis/proto/v1alpha1/ess.proto
```diff
@@ -0,0 +1,61 @@
...
+  bool changed = 1;
+}
+message DeleteKeysRequest {
+  ConfigReference config = 1;
+  Secret secret = 2;
+  DeleteOptions options = 3;
+}
+message DeleteKeysResponse {}
+message GetOptions {}
+message ApplyOptions {}
```
negz, line 58: If these aren't used, could we add them later if/when they turn out to be needed?

pkg/connection/store/external/store.go
```diff
@@ -0,0 +1,186 @@
...
+	errApply  = "cannot apply secret"
+	errDelete = "cannot delete secret"
+)
+
+var (
+	certsPathFmt = "/certs/%s"
+	caCertFile   = "ca.crt"
+	tlsCertFile  = "tls.crt"
+	tlsKeyFile   = "tls.key"
+)
```
negz, line 53: Could we make it so that these paths are configurable (e.g. via flags)?

pkg/connection/store/external/store.go
```diff
@@ -0,0 +1,186 @@
...
+
+// SecretStore is an External Secret Store.
+type SecretStore struct {
+	client     ess.ExternalSecretStoreServiceClient
+	kubeClient client.Client
+	ctx        context.Context
```
negz, line 59: I believe contexts in structs are considered an anti-pattern - see https://go.dev/blog/context-and-structs.

negz (commented):
I haven't done a full review, but this PR caught my eye so I've left a few suggestions. :)

apis/proto/v1alpha1/ess.proto
ezgidemirel: One scenario we considered was, deploying the plugin and the ESS to a different cluster where XP is running.  In that case, the `vaultconfig` CRD and the config resource will be installed in the ESS cluster and it won't be possible for clients (XP and providers) to read and send it to the gRPC server.

pkg/connection/store/external/store.go
ezgidemirel: I kept the file names constant and get the folder name via flag. Is it ok?

pkg/connection/store/external/store.go
```diff
@@ -0,0 +1,147 @@
...
+type SecretStore struct {
+	client     ess.ExternalSecretStoreServiceClient
+	kubeClient client.Client
+	config     *v1.Config
+}
+
+// NewSecretStore returns a new External SecretStore.
+func NewSecretStore(ctx context.Context, kube client.Client, tlsConfig *tls.Config, cfg v1.SecretStoreConfig) (*SecretStore, error) {
+	if cfg.Plugin.Endpoint == nil {
+		return nil, errors.New("endpoint is not provided")
```
negz, line 53: Should these error strings be constant? It seems like some are (e.g. `errGet`). I like having error strings be constants because you can then change them without having to update them in two places (i.e. the code and the tests).

pkg/connection/store/external/store.go
```diff
@@ -0,0 +1,147 @@
...
+	sec.ScopedName = filepath.Join(s.Scope, s.Name)
+
+	cfg := ss.getConfigReference()
+
+	_, err := ss.client.DeleteKeys(ctx, &ess.DeleteKeysRequest{Secret: sec, Config: cfg})
+	if err != nil {
+		return errors.Wrap(err, errDelete)
+	}
+
+	return nil
```
negz, line 137: Nit: This could be simplified. `errors.Wrap` will return `nil` if the wrapped error is `nil`.


```suggestion
	_, err := ss.client.DeleteKeys(ctx, &ess.DeleteKeysRequest{Secret: sec, Config: cfg})
	return errors.Wrap(err, errDelete)
```

pkg/controller/options.go
```diff
@@ -57,6 +58,9 @@ type Options struct {
 
 	// Features that should be enabled.
 	Features *feature.Flags
+
+	// TLSConfig for controllers to support External Secret Stores
+	TLSConfig *tls.Config
```
negz, line 63: Do we expect most/all Crossplane controllers to support external secret stores? Is there a chance controllers will need (separate) TLS configs for other features in future?

pkg/connection/manager.go
```diff
@@ -65,10 +66,11 @@ type DetailsManager struct {
 	client       client.Client
 	newConfig    func() StoreConfig
 	storeBuilder StoreBuilderFn
+	tlsConfig    *tls.Config
 }
 
 // NewDetailsManager returns a new connection DetailsManager.
-func NewDetailsManager(c client.Client, of schema.GroupVersionKind, o ...DetailsManagerOption) *DetailsManager {
+func NewDetailsManager(c client.Client, of schema.GroupVersionKind, tlsConfig *tls.Config, o ...DetailsManagerOption) *DetailsManager {
```
turkenh, line 73: What do you think about introducing the `tlsConfig` as an option? 

- Not all secret stores need it, e.g. `Kubernetes` 
- Not to have a breaking change in the function signature. 


```
func WithTLSConfig(tlsConfig *tls.Config) DetailsManagerOption {
	return func(m *DetailsManager) {
		m.tlsConfig = tlsConfig
	}
}
```

pkg/controller/options.go
turkenh: > Do we expect most/all Crossplane controllers to support external secret stores?
Yes, I believe so.

> Is there a chance controllers will need (separate) TLS configs for other features in future?
Yes, I would like to suggest moving this under a dedicated type for ESS.

pkg/connection/store/external/store.go
```diff
@@ -0,0 +1,147 @@
...
+		}
+	}
+
+	return nil
+}
+
+// WriteKeyValues writes key value pairs to a given Secret.
+func (ss *SecretStore) WriteKeyValues(ctx context.Context, s *store.Secret, wo ...store.WriteOption) (changed bool, err error) {
+	sec := new(ess.Secret)
+	sec.ScopedName = filepath.Join(s.Scope, s.Name)
```
turkenh, line 102: `s.Scope` is coming as the namespace of the owner resource. If the owner is a cluster scoped resource, than we have `s.Scope` as empty and should default it to `secretStoreConfig.defaultScope`. 

This should be done for all 3 methods.

apis/proto/v1alpha1/ess.proto
turkenh: While it sounds appealing to make the plugin unaware of Kubernetes, it introducing its own CRD, e.g. Vault specific configuration, makes it infeasible. The scenario that @ezgidemirel mentioned is a hypothetical example of this.

pkg/certificates/certificates.go
```diff
@@ -0,0 +1,74 @@
...
+	tlsKeyFileName  = "tls.key"
+)
+
+const (
+	errLoadCert  = "cannot load certificate"
+	errLoadCA    = "cannot load CA certificate"
+	errInvalidCA = "invalid CA certificate"
+)
+
+// Load loads TLS certificates in the given folder assuming certificate names are constant.
```
turkenh, line 41: nit: 

```suggestion
// Load loads TLS certificates in the given folder using well-defined filenames for certificates in a Kubernetes environment.
```

pkg/certificates/certificates.go
```diff
@@ -0,0 +1,74 @@
...
+func Load(certsFolderPath string, isServer bool) (*tls.Config, error) {
```
turkenh, line 42: Can we make this function name more descriptive? `certificates.Load` sounds too generic. Maybe something like, `TLSConfigForMTLS()`

apis/proto/v1alpha1/ess.proto
```diff
@@ -0,0 +1,70 @@
...
+*/
+
+syntax = "proto3";
+
+package ess.proto.v1alpha1;
+
+option go_package = "github.com/crossplane/crossplane-runtime/apis/proto/v1alpha1";
+
+// ExternalSecretStoreService defines the APIs for an External Secret Store plugin.
+service ExternalSecretStoreService {
```
turkenh, line 21: Should we include `Plugin` here?

```suggestion
service ExternalSecretStorePluginService {
```

apis/common/v1/connection_details.go
```diff
@@ -122,6 +125,30 @@ type SecretStoreConfig struct {
...
+}
+
+// ExternalStoreConfig represents configuration of an External Secret Store.
+type ExternalStoreConfig struct {
+	// Endpoint is the endpoint of the gRPC server.
+	// +optional
+	Endpoint *string `json:"endpoint,omitempty"`
+	// ConfigRef contains store config reference info.
+	// +optional
+	ConfigRef *Config `json:"configRef,omitempty"`
```
turkenh, line 141: Should both of these parameters be `required`?

pkg/connection/store/external/store.go
```diff
@@ -0,0 +1,147 @@
...
+// SecretStore is an External Secret Store.
+type SecretStore struct {
+	client     ess.ExternalSecretStoreServiceClient
+	kubeClient client.Client
+	config     *v1.Config
+}
+
+// NewSecretStore returns a new External SecretStore.
+func NewSecretStore(ctx context.Context, kube client.Client, tlsConfig *tls.Config, cfg v1.SecretStoreConfig) (*SecretStore, error) {
+	if cfg.Plugin.Endpoint == nil {
```
turkenh, line 52: This should be a required config, and we won't need this check.

pkg/connection/store/external/store.go
```diff
@@ -0,0 +1,147 @@
...
+		return nil, errors.New("endpoint is not provided")
+	}
+
+	creds := credentials.NewTLS(tlsConfig)
+	conn, err := grpc.Dial(*(cfg.Plugin.Endpoint), grpc.WithTransportCredentials(creds))
+	if err != nil {
+		return nil, errors.Wrap(err, fmt.Sprintf("cannot dial to the endpoint: %s", *cfg.Plugin.Endpoint))
```
turkenh, line 59:
```suggestion
		return nil, errors.Wrapf(err, "cannot dial to the endpoint: %s", *cfg.Plugin.Endpoint)
```

pkg/connection/store/external/store.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	cl := ess.NewExternalSecretStoreServiceClient(conn)
```
turkenh, line 62: nit: no need to define a new variable, has single usage.

pkg/connection/store/external/store.go
```diff
@@ -0,0 +1,147 @@
...
+	return &SecretStore{
+		kubeClient: kube,
+		client:     cl,
+		config:     cfg.Plugin.ConfigRef,
+	}, nil
+}
+
+// ReadKeyValues reads and returns key value pairs for a given Secret.
+func (ss *SecretStore) ReadKeyValues(ctx context.Context, n store.ScopedName, s *store.Secret) error {
+	sec := new(ess.Secret)
```
turkenh, line 73: nit:

```suggestion
	sec := &ess.Secret{}
```

I don't think there is something related in the styling docs/guides we are following, but I would like to mention that I am not familiar `new` keyword in XP codebase. You may consider using the suggested way just to be consistent.

pkg/connection/store/external/store.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	if s.Metadata != nil && len(s.Metadata.Labels) != 0 {
+		sec.Metadata = make(map[string]string, len(s.Metadata.Labels))
+		for k, v := range s.Metadata.Labels {
+			sec.Metadata[k] = v
+		}
+	}
+
+	cfg := ss.getConfigReference()
```
turkenh, line 115: nit: variable only used once, can be removed.

pkg/connection/store/external/store.go
negz: I _think_ `new` is actually preferred by Go (citation needed) but we tend to use `&Foo{}` syntax instead as you point out. We're committed now so good to stay consistent. :)

apis/proto/v1alpha1/ess.proto
negz: I see. Thanks for the context!

pkg/connection/store/external/store.go
negz: My preference would be to make them completely configurable. Is there a benefit to being opinionated about the filenames?

ezgidemirel: made them configurable

pkg/certificates/certificates.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+const (
+	errLoadCert  = "cannot load certificate"
+	errLoadCA    = "cannot load CA certificate"
+	errInvalidCA = "invalid CA certificate"
+)
+
+// LoadMTLSConfig loads TLS certificates in the given folder using well-defined filenames for certificates in a Kubernetes environment.
+func LoadMTLSConfig(certsFolderPath, caCertFileName, tlsCertFileName, tlsKeyFileName string, isServer bool) (*tls.Config, error) {
```
turkenh, line 36:
```suggestion
func LoadMTLSConfig(caPath, certPath, keyPath string, isServer bool) (*tls.Config, error) {
```

We can simplify parameters and get rid of one of them this way.

turkenh (approved):
Great work, thanks @ezgidemirel 💪

PR #384, opened 2023-02-23
Observe Only Management Policy Support

turkenh:

This PR implements the proposed changes in the [Observe Only design](https://github.com/crossplane/crossplane/pull/3531).

I have:

By consuming this PR from a provider and creating resources with `spec.managementPolicy: ObserveOnly`.

### DIFF
--- a/apis/common/v1/resource.go
+++ b/apis/common/v1/resource.go
@@ -211,7 +211,7 @@ type ResourceSpec struct {
 	// This field is planned to replace the DeletionPolicy field in a future
 	// release. Currently, both could be set independently and non-default
 	// values would be honored if the feature flag is enabled.
-	// See the design doc for more information: https://github.com/crossplane/crossplane/blob/master/design/design-doc-observe-only-resources.md#deprecation-of-deletionpolicy
+	// See the design doc for more information: https://github.com/crossplane/crossplane/blob/499895a25d1a1a0ba1604944ef98ac7a1a71f197/design/design-doc-observe-only-resources.md?plain=1#L223
 	// +optional
 	// +kubebuilder:default=FullControl
 	ManagementPolicy ManagementPolicy `json:"managementPolicy,omitempty"`
@@ -222,7 +222,7 @@ type ResourceSpec struct {
 	// This field is planned to be deprecated in favor of the ManagementPolicy
 	// field in a future release. Currently, both could be set independently and
 	// non-default values would be honored if the feature flag is enabled.
-	// See the design doc for more information: https://github.com/crossplane/crossplane/blob/master/design/design-doc-observe-only-resources.md#deprecation-of-deletionpolicy
+	// See the design doc for more information: https://github.com/crossplane/crossplane/blob/499895a25d1a1a0ba1604944ef98ac7a1a71f197/design/design-doc-observe-only-resources.md?plain=1#L223
 	// +optional
 	// +kubebuilder:default=Delete
 	DeletionPolicy DeletionPolicy `json:"deletionPolicy,omitempty"`


turkenh:
This PR is ready for review - leaving it as a draft until we get the design merged.

apis/common/v1/resource.go
```diff
@@ -203,6 +203,11 @@ type ResourceSpec struct {
 	// Deprecated: Please use ProviderConfigReference, i.e. `providerConfigRef`
 	ProviderReference *Reference `json:"providerRef,omitempty"`
 
+	// ManagementPolicy specifies how this managed resource should be managed.
```
lsviben, line 206: This seems a bit vague, I tried putting it in different words but feel free to discard/change
```suggestion
	// ManagementPolicy specifies the level of control Crossplane has over the managed external resource.  
    // The options are: 
    // FullControl (default) - fully controlled by Crossplane controllers, including deletion.
    // ObserveOnly - only observed by Crossplane controllers, not modified or deleted.
    // OrphanOnDelete - orphaned when its managed resource is deleted.
```

pkg/reconciler/managed/reconciler.go
```diff
@@ -791,6 +816,41 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		return reconcile.Result{Requeue: true}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
 	}
 
+	if r.managementPoliciesEnabled && managed.GetManagementPolicy() == xpv1.ManagementObserveOnly {
+		// In the observe-only mode, observation.ResourceExists will be an error
```
lsviben, line 820: NIT
```suggestion
		// In the observe-only mode, !observation.ResourceExists will be an error
```

apis/common/v1/policies.go
```diff
@@ -16,14 +16,33 @@ limitations under the License.
...
+// external resource.
+// +kubebuilder:validation:Enum=FullControl;ObserveOnly;OrphanOnDelete
+type ManagementPolicy string
+
+const (
+	// ManagementFullControl means the external resource is fully controlled
+	// by Crossplane controllers, including its deletion.
+	ManagementFullControl ManagementPolicy = "FullControl"
+
+	// ManagementObserveOnly means the external resource will only be observed
```
lsviben, line 29: Should we give here info that the external resource has to already exist if this ManagementPolicy is chose, or is there a better place for that?

apis/common/v1/resource.go
```diff
@@ -203,6 +203,11 @@ type ResourceSpec struct {
...
+	// +optional
+	// +kubebuilder:default=FullControl
+	ManagementPolicy ManagementPolicy `json:"managementPolicy,omitempty"`
```
negz, line 209: Let's mark this field as alpha. In core APIs we've done this by including the following text in the doctoring:

```go
// THIS IS AN ALPHA FIELD. Do not use it in production. It is not honored
// unless the relevant Crossplane feature flag is enabled, and may be
// changed or removed without notice.
```

I think it would also be worth explaining the interplay between `ManagementPolicy` and `DeletionPolicy` in one or both of the field comments.

pkg/reconciler/managed/reconciler.go
```diff
@@ -445,9 +448,10 @@ type Reconciler struct {
 	client     client.Client
 	newManaged func() resource.Managed
 
-	pollInterval        time.Duration
-	timeout             time.Duration
-	creationGracePeriod time.Duration
+	pollInterval              time.Duration
+	timeout                   time.Duration
+	creationGracePeriod       time.Duration
+	managementPoliciesEnabled bool
```
negz, line 454: I wonder if we should just pass some `feature.Flags` to the reconciler? Doesn't make much difference now but it would be handy if we ever end up having multiple feature flags that affect it.

turkenh: I have mixed feelings about this. Currently, AFAIK, we have two feature flags affecting the runtime code for this and external secret stores. While it makes sense to pass the feature flag for management policies, external secret store is a bit different since we would need to pass [another parameter](https://github.com/turkenh/upbound-provider-aws/blob/cd9a184a48b63e3bc4f36a35e3c8c08e8a6d829b/internal/controller/ec2/vpc/zz_controller.go#L34) other than the flag itself.

The most significant advantage would be not needing to update all Setup calls when we add a new feature flag, but this does not hold for 1 of the 2 examples we have today.  

[This](https://github.com/turkenh/upbound-provider-aws/blob/cd9a184a48b63e3bc4f36a35e3c8c08e8a6d829b/internal/controller/ec2/vpc/zz_controller.go#L32-L51) is how we are handling feature flags for those two features today, do you still think it makes sense?

apis/common/v1/policies.go
lsviben: Yeah it should be obvious.

lsviben (approved):

pkg/reconciler/managed/reconciler.go
```diff
@@ -675,10 +686,24 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
 
-	// If managed resource has a deletion timestamp and and a deletion policy of
+	// Check if the ManagementPolicy is set to a non-default value while the
+	// feature is not enabled. This is a safety check to let users know that
+	// they need to enable the feature flag before using the feature. For
+	// example, we wouldn't want someone to set the policy to ObserveOnly but
+	// not realize that the controller is still trying to reconcile
+	// (and modify or delete) the resource since they forgot to enable the
+	// feature flag.
+	if !r.managementPoliciesEnabled && (managed.GetManagementPolicy() == xpv1.ManagementObserveOnly || managed.GetManagementPolicy() == xpv1.ManagementOrphanOnDelete) {
```
negz, line 696: I don't love the complexity of this conditional, but I feel okay about it temporarily given it will go away once management policies are GA.

pkg/reconciler/managed/reconciler.go
negz: One other thought here - this will need to default to enable once we go beta (and the feature is still behind a flag, but on by default). I don't think that would impact this decision though (presumably it could just leave this code as is, and change the default flag value in the callers to enabled).

negz (approved):

PR #389, opened 2023-03-08
fix: properly validate max index

phisco:

Fixing previously introduced variable mismatch in tests and refactoring code.

I have:

Added unit tests.

### DIFF
--- a/pkg/fieldpath/paved.go
+++ b/pkg/fieldpath/paved.go
@@ -53,6 +53,7 @@ type Paved struct {
 	maxFieldPathIndex uint
 }
 
+// PavedOption can be used to configure a Paved behavior.
 type PavedOption func(paved *Paved)
 
 // PaveObject paves a runtime.Object, making it possible to get and set values
@@ -361,13 +362,13 @@ func (p *Paved) setValue(s Segments, value any) error {
 	// any per https://golang.org/pkg/encoding/json/#Unmarshal. We
 	// marshal our value to JSON and unmarshal it into an any to ensure
 	// it meets these criteria before setting it within p.object.
-	var v any
-	j, err := json.Marshal(value)
+	v, err := toValidJSON(value)
 	if err != nil {
-		return errors.Wrap(err, "cannot marshal value to JSON")
+		return err
 	}
-	if err := json.Unmarshal(j, &v); err != nil {
-		return errors.Wrap(err, "cannot unmarshal value from JSON")
+
+	if err := p.validateSegments(s); err != nil {
+		return err
 	}
 
 	var in any = p.object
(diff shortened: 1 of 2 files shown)


pkg/fieldpath/paved.go
```diff
@@ -404,6 +400,12 @@ func (p *Paved) setValue(s Segments, value any) error {
 				return nil
 			}
 
+			if !final && s[i+1].Type == SegmentIndex &&
```
turkenh, line 403: Looks like it is already `not final` (see the previous if).

```suggestion
			if s[i+1].Type == SegmentIndex &&
```

pkg/fieldpath/paved.go
```diff
@@ -404,6 +405,12 @@ func (p *Paved) setValue(s Segments, value any) error {
 				return nil
 			}
 
+			// If the next segment is an index we need to ensure the field is not above the max index, if set.
+			if s[i+1].Type == SegmentIndex &&
+				p.maxFieldPathIndexEnabled() && s[i+1].Index > p.maxFieldPathIndex {
```
turkenh, line 410: Isn't this case already covered with the above check 🤔 I mean case `SegmentIndex`


A small nit for readability
```suggestion
			if p.maxFieldPathIndexEnabled() && s[i+1].Type == SegmentIndex && s[i+1].Index > p.maxFieldPathIndex {
```

pkg/fieldpath/paved.go
```diff
@@ -404,6 +405,12 @@ func (p *Paved) setValue(s Segments, value any) error {
 				return nil
 			}
 
+			// If the next segment is an index we need to ensure the field is not above the max index, if set.
```
turkenh, line 408: Isn't this case already covered with the above check 🤔 I mean case `SegmentIndex` ?

pkg/fieldpath/paved.go
```diff
@@ -483,6 +484,17 @@ func (p *Paved) SetValue(path string, value any) error {
...
 
+func (p *Paved) validateSegments(s Segments) error {
+	if p.maxFieldPathIndexEnabled() {
+		for _, segment := range s {
+			if segment.Type == SegmentIndex && segment.Index > p.maxFieldPathIndex {
+				return errors.Errorf("index %v is greater than max allowed index %d", segment.Index, p.maxFieldPathIndex)
+			}
+		}
+	}
+	return nil
```
turkenh, line 495: nit for readability: 

```suggestion
        if !p.maxFieldPathIndexEnabled() {
            return nil
        }
        for _, segment := range s {
	        if segment.Type == SegmentIndex && segment.Index > p.maxFieldPathIndex {
		        return errors.Errorf("index %v is greater than max allowed index %d", segment.Index, p.maxFieldPathIndex)
	        }
        }
	return nil
```

turkenh (approved):

pkg/fieldpath/paved.go
phisco: generally agree with that, but in this case I wanted to keep it nested so that in the future additional checks could be added without changing the existing code

phisco: still, changed accordingly

PR #394, opened 2023-03-13
tests(fuzz): add FuzzParse test case

phisco:

While moving all the fuzz test cases over to `crossplane/crossplane` in https://github.com/crossplane/crossplane/pull/3671, we decided not to add `FuzzParse` because it belonged to this repo instead, as discussed [here](https://github.com/crossplane/crossplane/pull/3671%23discussion_r1089313171), so I'm adding it here and we'll configure [oss-fuzz](https://github.com/google/oss-fuzz/tree/master/projects/crossplane) to also build fuzzers for this project, as suggested by @AdamKorcz.

I have:

Running `go test -fuzz=FuzzParse` from the `pkg/parser` directory.

### DIFF
--- a/pkg/parser/parser_test.go
+++ b/pkg/parser/parser_test.go
@@ -19,6 +19,7 @@ package parser
 import (
 	"bytes"
 	"context"
+	"io"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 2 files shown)


AdamKorcz:
Great! If you want, you can also move the build script for the crossplane-runtime fuzzer over here, and we can invoke that script from OSS-Fuzz to make it easier to make modify the build.

negz (approved):

PR #399, opened 2023-03-17
Debugging `make check-diff`

negz:
Don't merge this - I'm just trying to reproduce some issues @ezgidemirel is seeing in https://github.com/crossplane/crossplane-runtime/pull/379.


negz:
Bumping the build submodule seems to have done the trick. I'll close this and bump that in https://github.com/crossplane/crossplane-runtime/pull/400.

PR #417, opened 2023-03-29
Filter out external create annotations from DesiredState

turkenh:

This PR filters out certain annotations from the Desired State by extending the recently introduced `DesiredStateChanged` predicate accordingly. 

This is to prevent backoff issues due to `crossplane.io/external-create-failed` and `crossplane.io/external-create-pending` annotations being constantly updated with timestamps as described in https://github.com/crossplane/crossplane/issues/3840

Fixes https://github.com/crossplane/crossplane/issues/3840

I have:

I have tested with this by creating a GCP bucket without enough permissions, as reported in the [issue](https://github.com/crossplane/crossplane/issues/3840). I could verify that the controller retries with proper back off with this change:

Screen Shot 2023-03-29 at 16 10 25

I have also validated that an annotation change other then ignored properly triggers a new reconcile.


lsviben (approved):
Looks good, thanks @turkenh

pkg/resource/predicates.go
```diff
@@ -164,13 +166,64 @@ func IsNamed(name string) PredicateFn {
...
 	return predicate.Or(
-		predicate.AnnotationChangedPredicate{},
+		AnnotationChangedPredicate{
+			ignored: []string{
+				// These annotations are managed by the controller and should
+				// not be considered as a change in desired state. The managed
+				// reconciler explicitly requests a new reconcile already after
+				// updating these annotations.
+				meta.AnnotationKeyExternalCreateFailed,
+				meta.AnnotationKeyExternalCreatePending,
```
negz, line 181: I'm not sure about hard coding these ignored annotations. Would it make sense to pass them in as arguments to `DesiredStateChanged`?

My thinking is:

* What if we want to use `DesiredStateChanged` in another context in future?
* What if we want to ignore other annotations in future, but not for 100% of callers?

Does every MR reconciler `Setup` call need to call `DesiredStateChanged` to create new predicates? If so I could understand if we didn't want to have to update all that boilerplate in future but I'm still on the fence.

pkg/resource/predicates.go
```diff
@@ -164,13 +166,64 @@ func IsNamed(name string) PredicateFn {
...
+	// to nil to ensure they are equal.
+	// https://github.com/golang/go/issues/16531
+	if len(na) == 0 {
+		na = nil
+	}
+	if len(oa) == 0 {
+		oa = nil
+	}
+
+	return !reflect.DeepEqual(na, oa)
```
negz, line 228: You could also use `cmp.Equal(na, oa, cmpopts.EquateEmpty())` here. We mostly use `cmp` in tests, but we do use it in 'real' code in a few places too.

https://pkg.go.dev/github.com/google/go-cmp/cmp/cmpopts#EquateEmpty

negz: Actually, we probably don't need reflection at all here. Wouldn't this work?


```go
na := e.ObjectNew.GetAnnotations()
oa := e.ObjectOld.GetAnnotations()

for _, k := range a.ignored {
	delete(na, k)
	delete(oa, k)
}

if len(na) != len(oa) {
	return false
}

for k, v := range na {
	if oa[k] != v {
		return false
	}
}

return true
```

I'm probably prematurely optimizing here so feel free to disregard. 😉

negz (approved):
A few comments but none of them are blocking. Thanks!

pkg/resource/predicates.go
turkenh: > Does every MR reconciler Setup call need to call DesiredStateChanged to create new predicates? 

Yes, please see as an example: https://github.com/crossplane-contrib/provider-aws/blob/44b8d0f3bb46c179aa5c69e0a2c13dfeb6f4d863/pkg/controller/database/dbsubnetgroup/controller.go#L71

This is the main reason why I didn't want to introduce them as arguments. 

> - What if we want to use DesiredStateChanged in another context in future?
> - What if we want to ignore other annotations in future, but not for 100% of callers?

Once we have some concrete use cases like these, we can introduce some optional arguments where the default behavior can be overridden.

pkg/resource/predicates.go
turkenh: Agreed that this is better than using reflect.
Updated accordingly, thanks!

PR #439, opened 2023-05-09
enable Kubernetes based authentication for Vault as an External Secret Store

aaronschweig:

Hello Team,

I am currently developing a bigger internal provider for my company and we have a hard requirement of using vault as a secret store and not have the secret live inside kubernetes itself. I was very pleased to see that you already have a concept of [Vault as an External Secret Store](https://github.com/crossplane/crossplane/blob/master/design/design-doc-external-secret-stores.md). Unfourtunately it would require me to use a vault token which is not possible within our setup. Then I looked a little bit through to pull requests and found a little note in https://github.com/crossplane/crossplane-runtime/pull/322 that it was actually intended to enable vault kubernetes auth at a later point in time. As we had used plain `controller-runtime` scaffolded by `operator-sdk` before we started transitioning to crossplane we already have our enterprise vault setup to be able to authenticate service accounts from within our kubernetes clusters. So having the `crossplane-runtime` natively support vault kubernetes auth would be an amazing feature for us and would help us tremendously in securing the credentials that get created and not expose them accidentially.

The spec changes I made to the `StoreConfig` were inspired by the design document linked earlier and mainly include the addition of the `method: Kubernetes` which itself allows further configuration under the `kubernetes` key.

There the user is required to provide a `role` as it is required to provide a role which correlates to the service account by vault (https://developer.hashicorp.com/vault/docs/auth/kubernetes#via-the-api). Additionally the user can override the default mountPath of the kubernetes auth engine as well.

```yaml
apiVersion: automaticd.crossplane.io/v1alpha1
kind: StoreConfig
metadata:
  name: vault
spec:
  type: Vault
  vault:
    auth:
      method: Kubernetes
      kubernetes:
        role: the-role
        mountPath: "k8s"
    server: "https://vault.acme.corp"
    version: v2
```

The above example would allow the provider to authenticate with the specified vault instance and the service account token mounted to the pod the operator is running on.

To further customise the experience (I allowed more customisation with optional additional configuration). Most importantly the user is able to specify the namespace of the targeted vault instance as this feature for enterprise vaults was not implemented in the initial draft of the `connection_details.go` API. I think this would be a great addition which would not break any existing stuff out there.

Additionally I also allowed more fine grained control on how to retrieve the service account token that is used to authenticate against vault. I would justify this decision because the default way on how the kubernetes auth method from vault tries to receive the service account token is via looking up the file mounted to `/var/run/secrets/kubernetes.io/serviceaccount/token` which is not the case in every deployment and configuration. So I reused the idea of the `VaultAuthTokenConfig` and introduced the `ServiceAccountTokenSourceConfig` which allows the `StoreConfig` to point to the correct service account secret which contains the token.

A full example of the proposed spec would look like below:

```yaml
apiVersion: automaticd.crossplane.io/v1alpha1
kind: StoreConfig
metadata:
  name: vault
  namespace: the-k8s-namespace
spec:
  type: Vault
  vault:
    auth:
      method: Kubernetes
      kubernetes:
        role: the-role
        mountPath: "k8s"
        serviceAccountTokenSource:
          source: Secret
          secretRef:
            key: token
            name: the-k8s-secret
            namespace: the-k8s-namespace
    mountPath: "piper"
    server: "https://vault.acme.corp"
    version: v2
    namespace: some-ns
```

I could not find an existing Issue following up the comment made in the initial PR so I thought I might open a PR directly as the changes weren't to complicated.

Thanks already for some feedback and thoughts and if you think these changes could be beneficial to the project.

Fixes #

I have:

I have added unit tests for the newly added functionality in the `vault` package. I also ran a provider locally and tried out if the writing and authentication with vault actually works


ezgidemirel:
Hi @aaronschweig, thanks a lot for the PR! This is a great improvement, and it will help more people to use Vault in their enterprise environments.

I don't know if you've seen, but we recently implemented and merged the new external secret store [design](https://github.com/crossplane/crossplane/blob/master/design/one-pager-pluggable-secret-stores.md), which introduces `SecretStoreType: Plugin`. For the sake of this new type, we deprecated `SecretStoreType: Vault` and it will be removed from the code base in future releases. Therefore, I'd like to ask whether you can move this code to the [ess-plugin-vault](https://github.com/crossplane-contrib/ess-plugin-vault) repository. 

With the new design, authentication and CRUD operations of secrets are done by the Plugin. MTLS is enabled between Crossplane/provider pods and the Plugin itself. You can check the [documentation](https://docs.crossplane.io/knowledge-base/integrations/vault-as-secret-store/) for further details and how you can setup an environment.

Again, thanks for your contribution and please let me know if you have any questions/concerns.

aaronschweig:
Thanks a lot for pointing me in the right direction @ezgidemirel. As the code looked pretty similar in the mentioned repository I quickly ported the changes over in a [new PR](https://github.com/crossplane-contrib/ess-plugin-vault/pull/4) to the correct repository 😉

phisco:
@aaronschweig, thanks for your contribution! Looks like this PR can be closed now that https://github.com/crossplane-contrib/ess-plugin-vault/pull/4 has been merged, right?

aaronschweig:
Yes we can close this PR!

PR #446, opened 2023-05-30
Add Waiting condition reason

jeanduplessis:

Adds a new Waiting ConditionReason to avoid adding a free text value when indicating a waiting condition.
To be used in Crossplane to address https://github.com/crossplane/crossplane/issues/4094

I have:

Expanded on unit test. Will be tested further as part of fix in Crossplane.

PR in Crossplane: https://github.com/crossplane/crossplane/pull/4125

### DIFF
--- a/apis/common/v1/condition.go
+++ b/apis/common/v1/condition.go
@@ -45,6 +45,7 @@ const (
 	ReasonUnavailable ConditionReason = "Unavailable"
 	ReasonCreating    ConditionReason = "Creating"
 	ReasonDeleting    ConditionReason = "Deleting"
+	ReasonWaiting     ConditionReason = "Waiting"
 )
 
 // Reasons a resource is or is not synced.


phisco (approved):

apis/common/v1/condition.go
```diff
@@ -262,3 +263,16 @@ func ReconcilePaused() Condition {
 		Reason:             ReasonReconcilePaused,
 	}
 }
+
+// ReconcileWaiting returns a condition that indicates the composite resource
+// claim is currently waiting for its composite resource to become ready.
+// The message should contain a human readable reason for the wait.
```
negz, line 269: How strongly do you feel that we should add this to crossplane-runtime? I believe it's only used in one place (the c/c claim reconciler). Most of the other conditions here are very widely used - i.e. by every managed resource implementation.

I'm not against adding it here, but we have in the past found that doing so is likely to slow things down due to the coupling. i.e. You need to update the code here, then bump c/c's dependency on c/crossplane-runtime, etc. It may not be worth that cost if we don't think we'll end up using it in more than one place.

jeanduplessis: I don't feel strongly about it. I tried to follow the existing pattern in the file, but with it used only in one place, it's probably not worth making it DRY. Will revert.

jeanduplessis:
I will close this PR since adding `ReasonWaiting     ConditionReason = "Waiting"` when it's only used in one place isn't worth it right now. Will define it in the claim reconciler directly in c/c.

PR #456, opened 2023-06-20
Granular management policies

### LINKED ISSUE #453: Granular management policies
Rework the management policies alpha feature based on the new [design](https://github.com/crossplane/crossplane/blob/1f2a0bbd0a6d206fe719cc7aff7743159fefd526/design/one-pager-ignore-changes.md#proposal) which is based on granular management policies.

In practice this means changing from:
```
spec:
  managmentPolicy: FullControl/ObserveOnly/OrphanOnDelete
```
to:
```
spec:
  managmentPolicy: ["*", "Create", "Update", "Observe", "Delete", "LateInitialize"]
```
More details are in the proposal linked above.

lsviben:

Implements new granular management policies based on the [accepted design doc](https://github.com/crossplane/crossplane/blob/d0386c2c18c9719fc8b88caaa55c48840bc8c07a/design/one-pager-ignore-changes.md). 

In essence, it replaces the old `spec.managementPolicy` that was introduced with [ObserveOnly](https://github.com/crossplane/crossplane/blob/2c836060c5251db33b9539c5786812abca75693a/design/design-doc-observe-only-resources.mdshould) with an new array of granular management policies `spec.managementPolicies`

In addition as there are more options now, the various reconciler steps are now conditioned based on weather the management policies feature is in use and the combination of actions.

NOTE: As it is an alpha feature, we are just replacing the old `managementPolicy` field with the new one. 

Fixes #453

I have:

Deployed with updated upjet (PR TBD) and provider-aws in a local kind cluster.

1. Used the [Crossplane repo](https://github.com/crossplane/crossplane/) tools to create a kind cluster and install Crossplane on it:

```
./cluster/local/kind.sh up
./cluster/local/kind.sh helm-install
```

2. Check out the [aws-provider repo](https://github.com/upbound/provider-aws/) , add a replace to the go.mod which points to the branch of the [upjet changes PR](https://github.com/upbound/upjet/pull/224) and this PR changes.

```
replace github.com/crossplane/crossplane-runtime => github.com/lsviben/crossplane-runtime v0.0.0-20230712210651-dae13a90cf2f

replace github.com/upbound/upjet => github.com/lsviben/upjet v0.0.0-20230712211824-291bef216c84

replace github.com/crossplane/crossplane-tools => github.com/lsviben/crossplane-tools v0.0.0-20230712201434-f65dadcb91b1

require (
...
```

run 
```
go mod tidy
go mod vendor
```

Because of the controller-runtime version change, make a small change in the code at [internal/controller/providerconfig/config.go:31](https://github.com/upbound/provider-aws/blob/f296aa59412dba7003be40b3820a5a8e8fa0ccc5/internal/controller/providerconfig/config.go#L31)

```diff
                Named(name).
                WithOptions(o.ForControllerRuntime()).
                For(&v1beta1.ProviderConfig{}).
-               Watches(&source.Kind{Type: &v1beta1.ProviderConfigUsage{}}, &resource.EnqueueRequestForProviderConfig{}).
+               Watches(&v1beta1.ProviderConfigUsage{}, &resource.EnqueueRequestForProviderConfig{}).
                Complete(providerconfig.NewReconciler(mgr, of,
                        providerconfig.WithLogger(o.Logger.WithValues("controller", name)),
                        providerconfig.WithRecorder(event.NewAPIRecorder(mgr.GetEventRecorderFor(name)))))

```

3. Generate the CRDs
```
make generate
```

4. Apply the CRDs
```
kubectl apply -f package/crds/
```

5. Apply the provider config

```
apiVersion: aws.upbound.io/v1beta1
kind: ProviderConfig
metadata:
  name: default
spec:
  credentials:
    source: Secret
    secretRef:
      namespace: crossplane-system
      name: aws-creds
      key: creds
```

6. Create the AWS credentials secret
7. Run the provider with `--enable-management-policies`.

   You can update the `run` target in the Makefile as below

    ```diff
    diff --git a/Makefile b/Makefile
    index d529a0d6..84411669 100644
    --- a/Makefile
    +++ b/Makefile
    @@ -111,7 +111,7 @@ submodules:
     run: go.build
            @$(INFO) Running Crossplane locally out-of-cluster . . .
            @# To see other arguments that can be provided, run the command with --help instead
    -       UPBOUND_CONTEXT="local" $(GO_OUT_DIR)/provider --debug
    +       UPBOUND_CONTEXT="local" $(GO_OUT_DIR)/provider --debug --enable-management-policies
    
     # NOTE(hasheddan): we ensure up is installed prior to running platform-specific
     # build steps in parallel to avoid encountering an installation race condition.
    ```
   
    and run with: 

    ```shell
    make run
    ```

8. Apply some of the test cases below.

**NOTE**
I used the monolith Upbound AWS provider for simplicity, if your machine is having issues with the ~900 CRDs then consider running the smaller provider families for testing. 

### No management policy set - defaults to ["*"]
```
apiVersion: ec2.aws.upbound.io/v1beta1
kind: VPC
metadata:
  name: training-vpc-test
spec:
  forProvider:
    region: us-west-1
    cidrBlock: 172.16.0.0/16
    tags:
      Name: TrainingVpc
```

### Pause

Create a standard VPC (check above) and then edit the `spec.managementPolicies` to `[]`

```
apiVersion: ec2.aws.upbound.io/v1beta1
kind: VPC
metadata:
  name: training-vpc-test
spec:
  managementPolicies: []
  forProvider:
    region: us-west-1
    cidrBlock: 172.16.0.0/16
    tags:
      Name: TrainingVpc
```

The resource should be paused, you can test by trying to make it reconcile, by for example removing an annotation.

### Observe Only test

Create an external resource and make sure you use its name as the MR name. For AWS, we also need to provide some indentifier data (`forProvider.region`)
```
apiVersion: rds.aws.upbound.io/v1beta1
kind: Instance
metadata:
  name: an-existing-dbinstance
spec:
  managementPolicies: ["Observe"]
  forProvider:
    region: us-west-1
```

### Orphan through managementPolicies

As the `spec.managementPolicies` does not have the Delete action, after creating the resource below, waiting for it to be Ready
and then deleting it, the external resource will be orphaned.

```
apiVersion: ec2.aws.upbound.io/v1beta1
kind: VPC
metadata:
  name: training-vpc-test
spec:
  deletionPolicy: Delete
  managementPolicies: ["Observe", "Create", "Update", "LateInitialize"]
  forProvider:
    region: us-west-1
    cidrBlock: 172.16.0.0/16
    tags:
      Name: TrainingVpc
```

### Orphan through deletionPolicy

Same process as above, but here we check that the `spec.deletionPolicy: Orphan` is the one determining the external resource deletion as the `spec.managementPolicies` is a default value. 

```
apiVersion: ec2.aws.upbound.io/v1beta1
kind: VPC
metadata:
  name: training-vpc-test
spec:
  deletionPolicy: Orphan
  managementPolicies: ["*"]
  forProvider:
    region: us-west-1
    cidrBlock: 172.16.0.0/16
    tags:
      Name: TrainingVpc
```      

### Delete through managementPolicies, deletionPolicy is orphan

Same setup as above, but in this case, both `spec.deletionPolicy` and `spec.managementPolicies` are non-default so `spec.managementPolicies` trumps the `spec.deletionPolicy` and the external resource gets deleted.

```
apiVersion: ec2.aws.upbound.io/v1beta1
kind: VPC
metadata:
  name: training-vpc-test
spec:
  deletionPolicy: Orphan
  managementPolicies: ["Observe", "Create", "Update", "Delete"]
  forProvider:
    region: us-west-1
    cidrBlock: 172.16.0.0/16
    tags:
      Name: TrainingVpc
```

### No late initialization

Tested using a use-case from the [design doc](https://github.com/crossplane/crossplane/blob/5c97a1937335218b158e8ae204521efdfa588d92/design/one-pager-ignore-changes.md#autoscalinggroup):

Prerequisite:
```
apiVersion: autoscaling.aws.upbound.io/v1beta1
kind: LaunchConfiguration
metadata:
  annotations:
    upjet.upbound.io/manual-intervention: "This resource refers to an AMI ID."
    meta.upbound.io/example-id: autoscaling/v1beta1/launchconfiguration
  labels:
    testing.upbound.io/example-name: bob
  name: bob
spec:
  forProvider:
    imageId: ami-0f6ded6fd335abde3
    instanceType: t2.micro
    region: us-west-1
```

The Autoscaling group needs a Launch Configuration or a Launch Template to function, in this case I opted for the Launch Configuration. When its ready, create an AutoscalingGroup

```
apiVersion: autoscaling.aws.upbound.io/v1beta1
kind: AutoscalingGroup
metadata:
  annotations:
    meta.upbound.io/example-id: ecs/v1beta1/capacityprovider
    upjet.upbound.io/manual-intervention: This resource depends on LaunchConfiguration with manual intervention.
  labels:
    testing.upbound.io/example-name: example
  name: example
spec:
  managementPolicies: ["Observe", "Create", "Update", "Delete"]  
  forProvider:
    availabilityZones:
      - us-west-1b
    launchConfiguration: bob
    maxSize: 1
    minSize: 1
    region: us-west-1
    tag:
      - key: AmazonECSManaged
        propagateAtLaunch: true
        value: "true"
```

The resources are taken from the [provider-aws examples.](https://github.com/upbound/provider-aws/tree/main/examples/autoscaling)

When the resource is created and ready, we can see that the `spec.forProvider.desiredCapacity` is not being late-initialized. We can check its value in `status.atProvider.desiredCapacity`. This allows the `desiredSize` to be managed by an external controller (autoscaler).


pkg/reconciler/managed/reconciler.go
```diff
@@ -693,7 +694,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	// not realize that the controller is still trying to reconcile
 	// (and modify or delete) the resource since they forgot to enable the
 	// feature flag.
-	if !r.managementPoliciesEnabled && (managed.GetManagementPolicy() == xpv1.ManagementObserveOnly || managed.GetManagementPolicy() == xpv1.ManagementOrphanOnDelete) {
+	if !r.managementPoliciesEnabled && (managed.GetManagementPolicy() != nil && !ContainsOnlyManagementAction(managed.GetManagementPolicy(), xpv1.ManagementActionAll)) {
```
turkenh, line 697: Starting to worry about readability in general. I am wondering if we could improve it by wrapping the logic in function(s) and also defining some internal (whitelisted) combinations as states. 

I don't have a great example of what it could look like yet :)

pkg/reconciler/managed/reconciler.go
```diff
@@ -676,13 +676,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	)
 
 	// Check the pause annotation and return if it has the value "true"
-	// after logging, publishing an event and updating the SYNC status condition
-	if meta.IsPaused(managed) {
-		log.Debug("Reconciliation is paused via the pause annotation", "annotation", meta.AnnotationKeyReconciliationPaused, "value", "true")
-		record.Event(managed, event.Normal(reasonReconciliationPaused, "Reconciliation is paused via the pause annotation"))
+	// or if the management policies are enabled and the ManagementPolicy is empty.
+	// Log, publish an event and update the SYNC status condition.
+	if IsPaused(managed, r.managementPoliciesEnabled) {
```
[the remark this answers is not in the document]
lsviben: This was my first thought as well, but then I didnt like leaving the fact of leaving a public API `meta.IsPaused` which does not really check if its paused as it doesnt have full info (management policies). Thats why i turned to moving it all here and deprecating the `meta.IsPaused`

pkg/meta/meta.go
```diff
@@ -423,6 +423,8 @@ func AllowsPropagationTo(from metav1.Object) map[types.NamespacedName]bool {
 
 // IsPaused returns true if the object has the AnnotationKeyReconciliationPaused
 // annotation set to `true`.
+// Deprecated: Use reconciler.IsPaused instead to check if the Crossplane managed resource is paused as it includes
+// the check for management policies.
```
negz, line 427: Nit: We wrap comments at 80 chars (I forget why, but we do 😄).

pkg/reconciler/managed/reconciler.go
negz, line 681: I know you didn't introduce this field, but wonder whether it would be better to just embed `feature.Flags` in `Reconciler` rather than using a `bool` field per flag. i.e.:

```go
type Reconciler struct {
	// Various fields redacted...

	// Replaces managementPoliciesEnabled
	features feature.Flags
}
```

This would:

* Make it easier to grep the codebase for the pattern - e.g. `r.features.Enabled(feature.SomeCoolFeature)`
* Make it easier to handle multiple feature flags in this reconciler in future.

lsviben:
Discussed with @turkenh to add an allowlist of supported combinations of ManagementPolicies and check that in the reconciler.

What we though about is initially supporting:

Default:
```
managememntPolicy: ["*"]
```
All - alternative to default
```
managememntPolicy: ["Observe", "Create", "Update", "LateInitialize", "Delete"]
```
Observe Only
```
managememntPolicy: ["Observe"] 
```
Pause
```
managememntPolicy: [""]
```
No late init
```
managememntPolicy: ["Observe", "Create", "Update", "Delete"]
```
No late init, orphan
```
managememntPolicy: ["Observe", "Create", "Update"]
```
Orphan
```
managememntPolicy: ["Observe", "Create", "Update", "LateInitialize"]
```

These are some combinations that we thought make sense now. Possibly there are other combinations that we could add to the initial list. 

One issue with this is that if we want to add a combination, we need to do a release of crossplane-runtime, and then the providers need to be updated and released.

But still, it gives a clear picture of what kind of managed policy behaviors we are supporting, and can say that we tested and know they work.

lsviben:
I have tested migrating from the previous `managmenetPolicy: FullControl/OrphanOnDelete/Observe` to the new `managementPolicy: ["*"]`.

For some reason (also based on some early testing) I expected that after applying the new CRD with the updated OpenAPI spec, that the old policy will get replaced by the default new managementPolicy (`spec.managmementPolicy: [*]`). Well I was wrong :D. 

What happens is that the old `managementPolicy` doesnt get removed and starting the providers fails because of the unexpected value in the field. 

If I dont find an easy and elegant solution for that I am considering simply renaming the new `spec.managementPolicy` to `spec.managmentPolicies`. This will remove the old field and default the new one when the new CRD is applied 

cc @turkenh

chlunde:
Nice! I am really looking forward to this! 👍  A couple of late comments, I missed the proposal.

## If limiting policy combinations, consider supporting immutable resources `["Create", "Delete"]`
> Discussed with @turkenh to add an allowlist of supported combinations of ManagementPolicies and check that in the reconciler.
> 
> What we though about is initially supporting:
...

Immutable resources (like Kubernetes Job) would work better if we have support for the following combinations:
 
```
managementPolicy: ["Create", "Delete"]
managementPolicy: ["Create", "Observe", "Delete"]
```

I don't see any technical reason why "LateInitialize" could not be part of it, which makes four meaningful, valid policies for immutable resources alone.

(Have you written what combinations you worry about supporting and what issues you expect? With the combinations mentioned above allowed, I wonder what is *not* there 😄 )

---

## Keeping LateInitialize
Are there any good use cases for LateInitialize when Observe exists?

As a user, I think `LateInitialize` might have two paths forward:
* Disable and deprecate it, as it is a major source of issues. Just keep it around for legacy use cases for a while.
* A mode where LateInitialize keeps updating the field, by using managed fields and checking that the provider is the manager of that specific field.
Am I missing some use cases where it is helpful?

---

## initProvider name
It might make sense to consider the name "initProvider" together with deletion options, to see if there's a way to make it intuitive that these are API attributes specific to create and delete.

I know of three kinds of additional deletion behaviors:
* the external API has force and/or backup flags: (ECR, [forceDelete](https://github.com/crossplane-contrib/provider-aws/pull/743). RDS: final snapshot name)
* the provider, like terraform AWS provider, [directly deletes data, for S3](https://github.com/crossplane-contrib/provider-aws/issues/495).
* Orphan

I wonder if the design should consider this to ensure the API is orthogonal and consistent in naming.

---

## Use of array - usability in compositions?

I wonder how well this will work with compositions, if you want to have one or more feature flags in a definition to allow the user to control if a resource should be orphaned, for example.

For example, map transforms do not work as they output single strings. This might require a new transform like "add to set", or a map transform returning a list. Have you tested a use case like this end-to-end?

---

I hope the tone of this comment does not look negative, as I really like this proposal.

lsviben:
Thanks for the feedback @chlunde , its really appreciated! :+1: 

> Nice! I am really looking forward to this! 👍 A couple of late comments, I missed the proposal.
> 
> ## If limiting policy combinations, consider supporting immutable resources `["Create", "Delete"]`
> > Discussed with @turkenh to add an allowlist of supported combinations of ManagementPolicies and check that in the reconciler.
> > What we though about is initially supporting:
> > ...
> 
> Immutable resources (like Kubernetes Job) would work better if we have support for the following combinations:
> 
> ```
> managementPolicy: ["Create", "Delete"]
> managementPolicy: ["Create", "Observe", "Delete"]
> ```
> 
> I don't see any technical reason why "LateInitialize" could not be part of it, which makes four meaningful, valid policies for immutable resources alone.
> 
> (Have you written what combinations you worry about supporting and what issues you expect? With the combinations mentioned above allowed, I wonder what is _not_ there smile )

Thanks for the suggestion. The idea is to support valid use cases, and to have confidence that it works properly. 
We could add the "immutable" use cases, seams reasonable to me.

One combination I am not sure if its neccessary:
```
managementPolicy: ["Create", "Delete"]
```

as I dont see a reason why not to add the Observe. Having combinations with LateInitialize and Delete make sense. 

We dont have to add all valid combinations right away, until we see a use case for it. As this is an alpha feature, IMO its better to start small, and iterate through discovering use cases we want to solve. There could be valid combinations as well which are omitting "Create" for importing cases, but while its not needed, I am hesitant to add it right away and increase the load of what we are supporting.

Of course there are also combinations which dont make sense like ["LateInitialize"].

> ## Keeping LateInitialize
> Are there any good use cases for LateInitialize when Observe exists?
> 
> As a user, I think `LateInitialize` might have two paths forward:
> 
> * Disable and deprecate it, as it is a major source of issues. Just keep it around for legacy use cases for a while.
> * A mode where LateInitialize keeps updating the field, by using managed fields and checking that the provider is the manager of that specific field.
>   Am I missing some use cases where it is helpful?

TBH, I also see that going forward it may be that Late initialization gets deprecated. When I started to work on this issue, I also thought about deprecating/removing it alltogather, but as its a feature that has been around for a while, people are already using late initialized fields in patches and such. As we now have `status.atProvider` I think late initialization is losing its usefulness and we could rethink if its needed. I may be also missing some other use cases where it is helpful.

> ## initProvider name
> It might make sense to consider the name "initProvider" together with deletion options, to see if there's a way to make it intuitive that these are API attributes specific to create and delete.
> 
> I know of three kinds of additional deletion behaviors:
> 
> * the external API has force and/or backup flags: (ECR, [forceDelete](https://github.com/crossplane-contrib/provider-aws/pull/743). RDS: final snapshot name)
> * the provider, like terraform AWS provider, [directly deletes data, for S3](https://github.com/crossplane-contrib/provider-aws/issues/495).
> * Orphan
> 
> I wonder if the design should consider this to ensure the API is orthogonal and consistent in naming.

Ehh, naming is hard. For me, `initProvider` makes sense as it has a connotation to `forProvider` which its complementing and the `init` in the name relates to me that is just used at the start of the resources lifecycle ( create). I don't have a better idea right now, `createProvider` or something like that does'nt sound so good to me.

Do you have some suggestion for the naming? 

> ## Use of array - usability in compositions?
> I wonder how well this will work with compositions, if you want to have one or more feature flags in a definition to allow the user to control if a resource should be orphaned, for example.
> 
> For example, map transforms do not work as they output single strings. This might require a new transform like "add to set", or a map transform returning a list. Have you tested a use case like this end-to-end?

I see your point, it would be useful to have a map transform with a list for this case, not only for the management policies, but in general for arrays. I haven't tested this in a composition yet btw.

lsviben:
> I have tested migrating from the previous `managmenetPolicy: FullControl/OrphanOnDelete/Observe` to the new `managementPolicy: ["*"]`.
> 
> For some reason (also based on some early testing) I expected that after applying the new CRD with the updated OpenAPI spec, that the old policy will get replaced by the default new managementPolicy (`spec.managmementPolicy: [*]`). Well I was wrong :) .
> 
> What happens is that the old `managementPolicy` doesnt get removed and starting the providers fails because of the unexpected value in the field.
> 
> If I dont find an easy and elegant solution for that I am considering simply renaming the new `spec.managementPolicy` to `spec.managmentPolicies`. This will remove the old field and default the new one when the new CRD is applied
> 
> cc @turkenh

Changed the field name from `spec.managmentPolicy` to `spec.managementPolicies`. When the newly generated CRD is applied, the following will happen:

Old resource:
```
spec:
  managmentPolicy: FullControl
```
apply new CRD:
```
spec:
  managmentPolicies: ["*"]
```

If we keep the name the same, what would happen is that the value will stay the same, even though the CRD spec changed:

Applying a resource with the old management policies CRD:
```
...
spec:
  managementPolicy: FullControl
```
applying the new CRD withouth changing the mane:
```
...
spec:
  managementPolicy: FullControl
```
It stays the same even though the CRD spec changed. I tried triggering an update or something to see if it gets changed, but the edit doesnt even pass because of the validation failing (it now expects an array instead of a string). Of course new resources are created correctly. 

Decided that its better to go with the name change as it achieves the effect we are looking for.

pkg/reconciler/managed/reconciler.go
```diff
@@ -83,6 +87,31 @@ const (
 	reasonReconciliationPaused event.Reason = "ReconciliationPaused"
 )
 
+var supportedManagementPolicies = []sets.Set[xpv1.ManagementAction]{
+	// default
+	sets.New[xpv1.ManagementAction](xpv1.ManagementActionAll),
+	//// all
```
negz, line 93:
```suggestion
	// all
```

Should these have just two `/`?

Also, I wonder if these comments should be a little more detailed? Or dropped? Reading through them right now I'm not sure they're providing much extra context over the code.

https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

pkg/reconciler/managed/reconciler.go
```diff
@@ -693,17 +726,25 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+	if isNonDefaultManagementPolicy(managementPoliciesEnabled, managed.GetManagementPolicy()) {
+		log.Debug(errManagementPolicyNonDefault, "policy", managed.GetManagementPolicy())
+		record.Event(managed, event.Warning(reasonManagementPolicyNotEnabled, errors.New(errManagementPolicyNonDefault)))
+		managed.SetConditions(xpv1.ReconcileError(errors.New(errManagementPolicyNonDefault)))
+		return reconcile.Result{}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
+	}
+
+	if !isManagementPolicySupported(managementPoliciesEnabled, managed.GetManagementPolicy()) {
+		log.Debug(errManagementPolicyNotSupported, "policy", managed.GetManagementPolicy(), "supported",
+			supportedManagementPolicies)
```
negz, line 738: Not: Any reason to wrap this? I don't think its longer than other lines in the file.

negz: Also, what does `supportedManagementPolicies` look like when logged? Is it huge?

pkg/reconciler/managed/reconciler.go
```diff
@@ -693,17 +726,25 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
-	if !r.managementPoliciesEnabled && (managed.GetManagementPolicy() == xpv1.ManagementObserveOnly || managed.GetManagementPolicy() == xpv1.ManagementOrphanOnDelete) {
-		log.Debug(errManagementPolicy, "policy", managed.GetManagementPolicy())
-		record.Event(managed, event.Warning(reasonManagementPolicyNotEnabled, errors.New(errManagementPolicy)))
-		managed.SetConditions(xpv1.ReconcileError(errors.New(errManagementPolicy)))
+	if isNonDefaultManagementPolicy(managementPoliciesEnabled, managed.GetManagementPolicy()) {
+		log.Debug(errManagementPolicyNonDefault, "policy", managed.GetManagementPolicy())
+		record.Event(managed, event.Warning(reasonManagementPolicyNotEnabled, errors.New(errManagementPolicyNonDefault)))
+		managed.SetConditions(xpv1.ReconcileError(errors.New(errManagementPolicyNonDefault)))
+		return reconcile.Result{}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
+	}
```
negz, line 734: I wonder whether there's any value in debug logging the management policy even in the happy path? I could imagine adding it to all logs using `log.WithValues`.

pkg/reconciler/managed/reconciler.go
```diff
@@ -693,17 +726,25 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	// not realize that the controller is still trying to reconcile
 	// (and modify or delete) the resource since they forgot to enable the
 	// feature flag.
-	if !r.managementPoliciesEnabled && (managed.GetManagementPolicy() == xpv1.ManagementObserveOnly || managed.GetManagementPolicy() == xpv1.ManagementOrphanOnDelete) {
-		log.Debug(errManagementPolicy, "policy", managed.GetManagementPolicy())
-		record.Event(managed, event.Warning(reasonManagementPolicyNotEnabled, errors.New(errManagementPolicy)))
-		managed.SetConditions(xpv1.ReconcileError(errors.New(errManagementPolicy)))
+	if isNonDefaultManagementPolicy(managementPoliciesEnabled, managed.GetManagementPolicy()) {
```
negz, line 729: Would these sanity checks (e.g. using management policies without the feature flag, using invalid management policies) be suited to a webhook? Or perhaps an Initializer? Basically wondering if we can push them out of this (already very complex) reconcile loop.

pkg/reconciler/managed/reconciler.go
```diff
@@ -1096,28 +1115,123 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+		if p.Equal(policySet) {
+			return true
+		}
+	}
+	return false
+}
+
+// ContainsManagementAction returns true if the management policy contains
+// the management action or ManagementActionAll.
+func ContainsManagementAction(managementPolicy xpv1.ManagementPolicy, action xpv1.ManagementAction) bool {
```
negz, line 1136: Any reason why some of these are public and others private?

pkg/reconciler/managed/reconciler.go
```diff
@@ -1096,28 +1115,123 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+}
+
+// isManagementPolicyObserveOnly returns true if the management policy is
+// enabled, and contains only the Observe action.
+func isManagementPolicyObserveOnly(managementPolicyEnabled bool, policy xpv1.ManagementPolicy) bool {
+	if !managementPolicyEnabled {
+		return false
+	}
+	return ContainsOnlyManagementAction(policy, xpv1.ManagementActionObserve)
+}
```
negz, line 1185: A lot of these functions take a policy and whether management policies are enabled as arguments. Would it make sense to introduce a `ManagementPolicyChecker` type? Might result in more succinct calls, and less refactoring when `managementPolicyEnabled` goes away (i.e. when its GA).

I'm thinking something like:

```go
type ManagementPolicyChecker interface {
	ShouldUpdate() bool
	ShouldLateInit() bool
	IsObserveOnly() bool  // Could this be phrased in some kind of ShouldFoo way?
}
```

You could then have one implementation that had both enabled and policy, e.g.:

```go
// This is a very dumb name.
type PolicyBasedManagementPolicyChecker struct {
	policy xpv1.ManagementPolicy
	enabled bool
}
```

Or one that embeds just `xpv1.ManagementPolicy` and a different implementation that is used whenever management policies aren't actually enabled.

negz (commented):
@lsviben Apologies, super quick/shallow review because I didn't get around to this until the end of my day and I have to run. I have another reminder to look again tomorrow.

pkg/reconciler/managed/reconciler.go
lsviben: Added it to log.WithValues, but only if the management policies feature is enabled. Makes sense to add it to the logs in my opinion.

pkg/reconciler/managed/reconciler.go
lsviben: If it was only up to the managed resources, then yeah, we could. But as its also dependent on the flag in the provider, which can change over time, I am not sure if we can catch for example a case:
1. Management policies enabled
2. Add resource with non-default management policy - passes webhook/whatever
3. Management policies disabled - we have no way now to sanity check the above resource

Here in the reconciler we have the only place where both the flag and the resource come together live.

Probably something like OPA would be able to validate the provider flag change and warn/disallow it if resources are using custom managment policies, but that seems like an overkill. The code will get simpler when we reach GA, although I realize that is not soon.

pkg/reconciler/managed/reconciler.go
lsviben: Updated the comments with some more context

pkg/reconciler/managed/reconciler.go
lsviben: These ones are the most generic and used by the providers (at least I used it in the [upjet PR](https://github.com/upbound/upjet/pull/224)). For the others, it seemed to me that they are very specific to the reconciler so I left them private. 

They will probably change/go away as I work on [the comment below](https://github.com/crossplane/crossplane-runtime/pull/456#discussion_r1259102918).

apis/common/v1/policies.go
```diff
@@ -16,23 +16,39 @@ limitations under the License.
...
 const (
-	// ManagementFullControl means the external resource is fully controlled
-	// by Crossplane controllers, including its deletion.
-	ManagementFullControl ManagementPolicy = "FullControl"
+	// ManagementActionObserve means that the managed resource status.atProvider
+	// will be updated with the external resource state.
+	ManagementActionObserve ManagementAction = "Observe"
+
+	// ManagementActionCreate means that the external resource will be created
+	// when using the managed resource spec.initProvider and spec.forProvider.
```
turkenh, line 34: nit: do we need `when` ? also for update.
```suggestion
	// using the managed resource spec.initProvider and spec.forProvider.
```

pkg/reconciler/managed/reconciler.go
```diff
@@ -83,6 +87,46 @@ const (
...
+	sets.New[xpv1.ManagementAction](xpv1.ManagementActionObserve, xpv1.ManagementActionCreate, xpv1.ManagementActionLateInitialize),
+	// No Update and no LateInitialize, the external resource is not updated
+	// when the managed resource is updated and the spec.forProvider is not
+	// late initialized.
+	sets.New[xpv1.ManagementAction](xpv1.ManagementActionObserve, xpv1.ManagementActionCreate, xpv1.ManagementActionDelete),
+	// No Update, no Delete and no LateInitialize, the external resource is
+	// not updated when the managed resource is updated, the external resource
+	// is not deleted when the managed resource is deleted and the
+	// spec.forProvider is not late initialized.
+	sets.New[xpv1.ManagementAction](xpv1.ManagementActionObserve, xpv1.ManagementActionCreate),
```
turkenh, line 127: Should we also add Observe + Update ? https://github.com/crossplane-contrib/provider-kubernetes/issues/115

apis/common/v1/resource.go
```diff
@@ -205,15 +205,17 @@ type ResourceSpec struct {
...
-	// values would be honored if the feature flag is enabled.
+	// values would be honored if the feature flag is enabled. If both are
+	// custom, the DeletionPolicy field will be ignored.
 	// See the design doc for more information: https://github.com/crossplane/crossplane/blob/499895a25d1a1a0ba1604944ef98ac7a1a71f197/design/design-doc-observe-only-resources.md?plain=1#L223
+	// and this one: https://github.com/crossplane/crossplane/blob/444267e84783136daa93568b364a5f01228cacbe/design/one-pager-ignore-changes.md
 	// +optional
-	// +kubebuilder:default=FullControl
-	ManagementPolicy ManagementPolicy `json:"managementPolicy,omitempty"`
+	// +kubebuilder:default={"*"}
+	ManagementPolicy ManagementPolicy `json:"managementPolicies,omitempty"`
```
turkenh, line 218: Not being consistent with the json tag here is bugging me. Should we?

```suggestion
	ManagementPolicies ManagementPolicies `json:"managementPolicies,omitempty"`
```

pkg/reconciler/managed/reconciler.go
turkenh: I like this. 

The interface could be extended with `IsNonDefault` and `IsSupported` as well.

pkg/meta/meta.go
```diff
@@ -423,6 +423,8 @@ func AllowsPropagationTo(from metav1.Object) map[types.NamespacedName]bool {
 
 // IsPaused returns true if the object has the AnnotationKeyReconciliationPaused
 // annotation set to `true`.
+// Deprecated: Use reconciler.IsPaused instead to check if the Crossplane
+// managed resource is paused as it includes the check for management policies.
```
negz, line 427: Could we make this call `reconciler.IsPaused`? My guess is that might not be possible/desirable due to an import cycle.

pkg/reconciler/managed/reconciler.go
negz: Nice, thanks for updating these comments

I'd prefer to avoid this giant set (no pun intended) of package scoped state. Perhaps we could make it a member of `ManagementPolicyActionChecker` and move its declaration into the  `NewManagementPolicyActionChecker`?

pkg/reconciler/managed/reconciler.go
```diff
@@ -675,14 +764,21 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		"external-name", meta.GetExternalName(managed),
 	)
 
+	managementPoliciesEnabled := r.features.Enabled(feature.EnableAlphaManagementPolicies)
+	if managementPoliciesEnabled {
+		log.WithValues("managementPolicies", managed.GetManagementPolicy())
+	}
+	actionChecker := NewManagementPolicyActionChecker(managementPoliciesEnabled, managed.GetManagementPolicy())
```
negz, line 771: Would there be any benefit to injecting this as a dependency (i.e. making it a field of `Reconciler`? In practice I doubt it would ever be swapped out in production, but maybe it would help to be able to swap it out in tests?

pkg/reconciler/managed/reconciler.go
```diff
@@ -675,14 +764,21 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+
 	// Check the pause annotation and return if it has the value "true"
-	// after logging, publishing an event and updating the SYNC status condition
-	if meta.IsPaused(managed) {
-		log.Debug("Reconciliation is paused via the pause annotation", "annotation", meta.AnnotationKeyReconciliationPaused, "value", "true")
-		record.Event(managed, event.Normal(reasonReconciliationPaused, "Reconciliation is paused via the pause annotation"))
+	// or if the management policies are enabled and the ManagementPolicy is empty.
+	// Log, publish an event and update the SYNC status condition.
+	if IsPaused(managed, managementPoliciesEnabled) {
```
negz, line 776: Should IsPaused be a method of the `ManagementPolicyChecker`?

pkg/reconciler/managed/reconciler.go
```diff
@@ -675,14 +764,21 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+		log.Debug("Reconciliation is paused through the management policy or the pause annotation")
+		record.Event(managed, event.Normal(reasonReconciliationPaused, "Reconciliation is paused"))
```
negz, line 778: These might be more helpful if we named the annotation explicitly. Naming the management policy might be a bit verbose - perhaps we could at least say `spec.managementPolicy` to make it easier to find?

pkg/reconciler/managed/reconciler.go
```diff
@@ -83,12 +87,96 @@ const (
...
+	}
+}
+
+// ActionAllowed returns true if the supplied action is contained in the
+// management policy. Returns true if the management policy feature is disabled.
+func (m *ManagementPolicyActionChecker) ActionAllowed(action xpv1.ManagementAction) bool {
+	if !m.managementPolicyEnabled {
+		return true
+	}
+	return ContainsManagementAction(m.managementPolicy, action)
```
negz, line 167: Should we just inline these functions into their matching methods? Do they need to exist separately?

pkg/reconciler/managed/reconciler.go
negz: Yeah, this is probably wishful thinking on my part.

pkg/reconciler/managed/reconciler.go
negz: @lsviben Do you think you'll add `IsNonDefault` and `IsSupported` to this as well? I like @turkenh's suggestion - it would be nice for this type to be a one-stop-shop for all things management-policy-checking.

If we wanted to get reaaaally fancy it would be pretty clean to have a type that essentially loaded up all the context we need to make branching decisions (e.g management policies, deletion policy, paused annotation, etc etc). Something like:

```go
policy, err := NewManagementPolicyResolver(managed)
if err != nil {
    // NewManagementPolicyResolver returns an error if you're doing anything "wrong",
    //i.e. if you're using an unsupported management policy, or if you're trying to use
    // management policies but the feature flag isn't enabled.
}

type ManagementPolicyResolver interface {
    ShouldPause() bool
    ShouldOnlyObserve() bool

    ShouldCreate() bool
    ShouldDelete() bool  // Equivalent to !shouldOrphan
    ShouldLateInit() bool
    ShouldUpdate() bool
}
```

All the `ShouldFoo` functions don't feel quite as clean as the nice `Action` and `OnlyAction` methods you have that take a policy as arguments, but I think I still prefer them because they let us push all of the context needed to make "should I do this thing" checks (not only management policies) into the policy resolver.

pkg/reconciler/managed/reconciler.go
lsviben: I aimed for that initially, but it needs a management policy, which we get from the managed resource we GET somewhere at the beginning of the reconcile loop. So we could inject it without the policy and then pass it later as an method argument,  or inject it without and set it later, but that seemed like not the best solution to me.

Maybe there is another way that I am not seeing.

pkg/meta/meta.go
lsviben: yeah there is an import cycle. Plus I am not sure if the meta package should be aware of management policies.

pkg/reconciler/managed/reconciler.go
lsviben: sure, i can add something like that, also like the `Resolver` name. Also like how it hides the checks for supported and the sanity check

lsviben: Added something similar:
```go
// ManagementPoliciesChecker is used to perform checks on management policies
// to determine specific actions are allowed, or if they are the only allowed
// action.
type ManagementPoliciesChecker interface {
	// IsSupported returns true if the management policy is supported.
	IsSupported() bool
	// IsDisabledNonDefault returns true if the management policy is set to a non-default
	// value, and the management policy feature is disabled. Used as a sanity
	// check.
	IsDisabledNonDefault() bool
	// IsPaused returns true if the resource is paused based
	// on the management policy.
	IsPaused() bool

	// ShouldOnlyObserve returns true if only the Observe action is allowed.
	ShouldOnlyObserve() bool
	// ShouldCreate returns true if the Create action is allowed.
	ShouldCreate() bool
	// ShouldLateInitialize returns true if the LateInitialize action is
	// allowed.
	ShouldLateInitialize() bool
	// ShouldUpdate returns true if the Update action is allowed.
	ShouldUpdate() bool
	// ShouldDelete returns true if the Delete action is allowed.
	ShouldDelete() bool
}
```

Added the IsPaused to it, but in the end this IsPaused just checks if its paused based on the management policies, and I left the meta.IsPaused to check paused condition based on annotations. Felt wrong to mix annotations in the ManagementPolicies checking tool, so i left them separate. 

Checking IsDisabledNonDefault and IsSupported through calls instead of an error in `NewManagement...` due to:
- if we want to have IsPaused there, we need to initialize it before it, so to avoid changing the order of checks from `IsPaused->IsDisabledNonDefault->IsSupported` to `IsDisabledNonDefault->IsSupported->IsPaused`
- simpler for testing

Just noticed that I didnt change naming, ill check that.

pkg/reconciler/managed/reconciler.go
lsviben: moved the giant set into an func returning default supported policies and made it injectable for the `ManagedPoliciesResolver`, so that its not package-scoped anymore and we can still easily change it for tests.

apis/common/v1/resource.go
lsviben: Agreed, changed it

pkg/reconciler/managed/reconciler.go
lsviben: answered here: https://github.com/crossplane/crossplane-runtime/pull/456#discussion_r1261465502

pkg/reconciler/managed/reconciler.go
lsviben: They were used both in the reconciler and as helper functions for the providers. I removed them now as they are not used anymore in the reconciler. And I think the check will be simple enough to not need a helper function for the providers here.

pkg/reconciler/managed/reconciler.go
```diff
@@ -89,6 +96,217 @@ func ControllerName(kind string) string {
...
+}
+
+// ManagementPoliciesResolver is used to perform management policy checks
+// based on the management policy and if the management policy feature is enabled.
+type ManagementPoliciesResolver struct {
+	managementPolicyEnabled     bool
+	supportedManagementPolicies []sets.Set[xpv1.ManagementAction]
+	managementPolicy            sets.Set[xpv1.ManagementAction]
+	deletionPolicy              xpv1.DeletionPolicy
+}
```
negz, line 133: Wonder if it we should move all of the management policy related stuff into its own file since this one is pretty big. I don't feel strongly though.

pkg/reconciler/managed/reconciler.go
```diff
@@ -675,35 +894,55 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
-		record.Event(managed, event.Normal(reasonReconciliationPaused, "Reconciliation is paused via the pause annotation"))
+	managementPoliciesEnabled := r.features.Enabled(feature.EnableAlphaManagementPolicies)
+	if managementPoliciesEnabled {
+		log.WithValues("managementPolicies", managed.GetManagementPolicies())
+	}
+
+	// Create the management policy resolver which will assist us in determining
+	// what actions to take on the managed resource based on the management
+	// and deletion policies.
+	policyResolver := NewManagementPoliciesResolver(managementPoliciesEnabled, managed.GetManagementPolicies(), managed.GetDeletionPolicy())
```
negz, line 905:
```suggestion
	policy := NewManagementPoliciesResolver(managementPoliciesEnabled, managed.GetManagementPolicies(), managed.GetDeletionPolicy())
```

Naming nit: I'd prefer this because `policy.ShouldDelete()` reads more like "the policy is 'should delete'" to me. Not something I feel very strong about though.

pkg/reconciler/managed/reconciler.go
negz: > Felt wrong to mix annotations in the ManagementPolicies checking tool, so i left them separate.

FWIW despite using the name "management policy checker" in my examples I was thinking more of an "all the policies" checker as a single place to calculate the policy based on all the policy inputs, which right now I think are:

* Paused annotation
* Deletion policy
* Management policies

Though to be fair the first two will go away eventually at which point this would purely be checking management policies. I see your point WRT ordering of the paused/disabled/supported calls though. Don't feel super strongly on this one.

Another thought: Could we merge `IsDisabledNonDefault` and `IsSupported` into a `Validate() error` method that returns an error indicating what's wrong with the policy config (i.e. its set despite the flag being enabled, or its set to an invalid value)?

negz (approved):
This looks good to me, but I'd like @turkenh to approve too before we merge.

I left a few comments for possible improvements, but I think we're into optimizations at this point so feel free to take them or leave them.

pkg/reconciler/managed/reconciler.go
```diff
@@ -89,6 +96,217 @@ func ControllerName(kind string) string {
...
+	ShouldUpdate() bool
+	// ShouldDelete returns true if the Delete action is allowed.
+	ShouldDelete() bool
+}
+
+// ManagementPoliciesResolver is used to perform management policy checks
+// based on the management policy and if the management policy feature is enabled.
+type ManagementPoliciesResolver struct {
+	managementPolicyEnabled     bool
+	supportedManagementPolicies []sets.Set[xpv1.ManagementAction]
```
turkenh, line 130: nit for simplification:

```suggestion
	enabled     bool
	supportedPolicies []sets.Set[xpv1.ManagementAction]
```

pkg/reconciler/managed/reconciler.go
turkenh: So, we don't want to add `Observe` + `Update` for now but enable a way for individual providers to extend the supported list? When I check the code, this doesn't seem to be the case though.

pkg/reconciler/managed/reconciler.go
```diff
@@ -89,6 +96,217 @@ func ControllerName(kind string) string {
...
+	managementPolicy            sets.Set[xpv1.ManagementAction]
```
turkenh, line 131:
```suggestion
	managementPolicies            sets.Set[xpv1.ManagementAction]
```

just to be consistent

pkg/reconciler/managed/reconciler.go
```diff
@@ -1062,6 +1272,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		log.Debug("External resource differs from desired state", "diff", observation.Diff)
 	}
 
+	// skip the update if the management policy is set to ignore updates
+	if !policyResolver.ShouldUpdate() {
+		log.Debug("Reconciliation succeeded", "requeue-after", time.Now().Add(r.pollInterval))
```
turkenh, line 1277: Should we enhance the debug message to indicate the update is skipped?

turkenh (approved):
Looking good, great work @lsviben 💪

pkg/reconciler/managed/reconciler.go
lsviben: Actually I want to add `Observe + Update` but didnt yet have a chance to add and test it first. Actually maybe it would be better to add them in a next PR, so that this can be merged?

I havent thought about allowing providers to set their own supported policies, now they are injected just in the resovler which actually just helps with tests and not having the supported list package scoped. But I like the idea, we can inject the list into the Reconciler, so that the providers can change it easier if needed

pkg/reconciler/managed/reconciler.go
lsviben: moved it!

pkg/reconciler/managed/reconciler.go
lsviben: @turkenh now the supported policies can be injected into the Reconciler, so if some provider devs would like to test something new or have their own set, its possible :tada:

rbrunan:
Hey @lsviben, thank you for your work. 
I've been testing this PR in our environment and unfortunately, it is not working as expected or I'm doing something wrong.
The test that we did is related to [this example in the design doc](https://github.com/crossplane/crossplane/blob/d0386c2c18c9719fc8b88caaa55c48840bc8c07a/design/one-pager-ignore-changes.md#eks-nodegroup). 
The example seems wrong to me since the `.spec.scalingConfig` content is an [array](https://marketplace.upbound.io/providers/upbound/provider-aws-eks/v0.37.0/resources/eks.aws.upbound.io/NodeGroup/v1beta1#doc:spec-forProvider-scalingConfig), not a map. When I apply that in a composition I receive the following error:
```
composed resource "eks-nodepool": cannot apply the patch at index 12: spec.forProvider.scalingConfig is not an array
```

Changing that to:
```
        spec:
          managementPolicy: ["Create", "Update", "Delete", "Observe"]
          initProvider:
            scalingConfig:
              - desiredSize: 1
          forProvider:
            scalingConfig:
              - minSize: 1 # <spec.parameters.scalingConfig.minSize>
                maxSize: 20 # <spec.parameters.scalingConfig.maxSize>
```

Throws another error:
```
cannot compose resources: cannot apply composed resource: cannot create object: NodeGroup.eks.aws.upbound.io "aws-dev-eu-west-1-test-cas-eks-np-mgmt-a-7jdtj" is invalid: spec.forProvider.scalingConfig.desiredSize: Required value
```

Finally, If I remove all entries for `.spec.forProvider.scalingConfig` and I use only the `.spec.initProvider.scalingConfig` with fixed values, the MR is generated but without the scaling config block:
```
cannot run refresh: refresh failed: Insufficient scaling_config blocks: At least 1 "scaling_config" blocks are required.
So my understanding is that the initiProvider is not yet implemented in that PR.
```

Am I doing something wrong or the initProvider is not passing the parameters to the MR?

rbrunan:
>When the resource is created and ready, we can see that the spec.forProvider.desiredSize is not being late-initialized. We can check its value in status.atProvider.desiredSize. This allows the desiredSize to be managed by an external controller (autoscaler).

Note that the parameter in the Autoscaling group is called [desiredCapacity](https://marketplace.upbound.io/providers/upbound/provider-aws-autoscaling/v0.37.0/resources/autoscaling.aws.upbound.io/AutoscalingGroup/v1beta1#doc:spec-forProvider-desiredCapacity) the `desiredSize` one is for the NodeGroup resource.

lsviben:
> 

Hey, @rbrunan, thanks for the feedback.

This PR does not introduce the `initProvider` functionality, just the basic Granular Management Polices, you can check the examples in the description of the this PR. 

I am currently working on introducing `initProvider` to [upjet](https://github.com/upbound/upjet/issues/225), and upjet based providers (upbound provider-aws/gcp/azure). 

As for the example in the design doc, sorry for that, I handwrote them and probably missed the map/array thing. I see also that an old versuion of `managementPolicy: PartialControl` is there instead of `managementPolicy:["Observe", "Create", "Update", "Delete"]`

So for now, only examples which do not use `initProviders` should work, hope that makes it clearer

rbrunan:
> > 
> 
> Hey, @rbrunan, thanks for the feedback.
> 
> This PR does not introduce the `initProvider` functionality, just the basic Granular Management Polices, you can check the examples in the description of the this PR.
> 
> I am currently working on introducing `initProvider` to [upjet](https://github.com/upbound/upjet/issues/225), and upjet based providers (upbound provider-aws/gcp/azure).
> 
> As for the example in the design doc, sorry for that, I handwrote them and probably missed the map/array thing. I see also that an old versuion of `managementPolicy: PartialControl` is there instead of `managementPolicy:["Observe", "Create", "Update", "Delete"]`
> 
> So for now, only examples which do not use `initProviders` should work, hope that makes it clearer

Thank you, now everything makes sense.

PR #478, opened 2023-07-11
pkg/errors: add multi error

sttts:

Follow-up of https://github.com/crossplane/crossplane/pull/4311.

Add `errors.Join(errs ...error)` and `errors.MultiError` interface. It is based on kube apimachinery
`NewAggregate` as the output is not multi-line like with Golang's `errors.Join`.

Fixes #

I have:

### DIFF
--- a/pkg/errors/errors.go
+++ b/pkg/errors/errors.go
@@ -22,6 +22,8 @@ package errors
 import (
 	"errors"
 	"fmt"
+
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
 )
 
 // New returns an error that formats as the given text. Each call to New returns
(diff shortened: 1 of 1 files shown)


phisco (approved):
LGTM, if down the line upstream `errors.Join` will allow setting a custom separator instead of forcing `\n` we could easily switch to that.

pkg/errors/errors.go
```diff
@@ -124,3 +126,34 @@ func Cause(err error) error {
...
 	return err
 }
+
+// MultiError is an error that wraps multiple errors.
+type MultiError interface {
+	error
+	Unwrap() []error
+}
+
+// Join returns an error that wraps the given errors.
```
negz, line 136: > aggregate errors if there is clear use for the user, but not in general, i.e. fail fast is the default.

I strongly agree with this guidance you added in https://github.com/crossplane/crossplane/pull/4311. My concern with wrapping errors is that folks will reach for the tool too often. What do you think about including this recommendation here in the docstring?

pkg/errors/errors.go
```diff
@@ -22,6 +22,8 @@ package errors
 import (
 	"errors"
 	"fmt"
+
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
```
negz, line 26: Out of scope for this PR, but I think we use a few other things from `kerrors` pretty frequently that may be worth aliasing in this package. Can't remember what they are right now. :)

negz (approved):

pkg/errors/errors.go
sttts: IsNotFound and the like?

pkg/errors/errors.go
sttts: done

pkg/errors/errors.go
```diff
@@ -124,3 +126,38 @@ func Cause(err error) error {
...
+// Join returns an error that wraps the given errors. Any nil error values are
+// discarded. Join returns nil if errs contains no non-nil values. The error
+// formats as the concatenation of the strings obtained by calling the Error
+// method of each element of errs and formatting like:
+//
+//	[first error, second error, third error]
+//
+// Note: aggregating errors should not be the default. Usually, return only the
+//
+//	first error, and only aggregate if there is clear value to the user.
```
negz, line 145: Looks like the formatting got a little messed up here?

pkg/errors/errors.go
negz: Yeah. No need to add them now though, can follow up later.

pkg/errors/errors.go
sttts: indeed

sttts: fixed

phisco:
@sttts the linter is complaining because of the imports ordering, you can fix it automatically by running `GO_LINT_ARGS=" --fix " make reviewable` (which under the hood runs `golangci-lint run --fix` with the right version of golangci-lint)

PR #489, opened 2023-07-27
reconciler/managed: add crossplane_resource_drift_seconds metric

sttts:

Adds a metric `crossplane_resource_drift_seconds` that records the time since the previous reconcile when a resource was found to be out of sync.
As we don't record the last reconcile on the object (for reasons, we cannot update an object on every observe), we count from the last reconcile in the same process. This means obviously that we don't emit any metric when the provider restarts.

I have:

### DIFF
--- a/pkg/reconciler/managed/metrics.go
+++ b/pkg/reconciler/managed/metrics.go
@@ -1,0 +1,100 @@
+/*
+Copyright 2023 The Crossplane Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package managed
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/tools/cache"
+	kmetrics "k8s.io/component-base/metrics"
+	"sigs.k8s.io/controller-runtime/pkg/cluster"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+
+	"github.com/crossplane/crossplane-runtime/pkg/errors"
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
+)
+
+func init() {
+	metrics.Registry.MustRegister(drift)
+}
+
+var subSystem = "crossplane"
+
+var (
+	drift = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Subsystem: subSystem,
+		Name:      "resource_drift_seconds",
+		Help:      "ALPHA: How long since the previous successful reconcile when a resource was found to be out of sync; excludes restart of the provider",
+		Buckets:   kmetrics.ExponentialBuckets(10e-9, 10, 10),
+	}, []string{"group", "kind"})
+)
+
+// driftRecorder records the time since the last observation of a resource
+// and records the time since on update as a metric. This represents an upper
+// bound for the duration the drift existed.
+type driftRecorder struct {
+	lastObservation sync.Map
+	gvk             schema.GroupVersionKind
+
+	cluster cluster.Cluster
+}
+
+var _ manager.Runnable = &driftRecorder{}
+
+func (r *driftRecorder) Start(ctx context.Context) error {
+	inf, err := r.cluster.GetCache().GetInformerForKind(ctx, r.gvk)
+	if err != nil {
+		return errors.Wrapf(err, "cannot get informer for drift recorder for resource %s", r.gvk)
+	}
+
+	registered, err := inf.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		DeleteFunc: func(obj interface{}) {
+			if final, ok := obj.(cache.DeletedFinalStateUnknown); ok {
+				obj = final.Obj
+			}
+			managed := obj.(resource.Managed)
+			r.lastObservation.Delete(managed.GetName())
+		},
+	})
+	if err != nil {
+		return errors.Wrap(err, "cannot add delete event handler to informer for drift recorder")
+	}
+	defer inf.RemoveEventHandler(registered) //nolint:errcheck // this happens on destruction. We cannot do anything anyway.
+
+	<-ctx.Done()
+
+	return nil
+}
+
+func (r *driftRecorder) recordUnchanged(name string) {
+	r.lastObservation.Store(name, time.Now())
+}
+
+func (r *driftRecorder) recordUpdate(name string) {
+	last, ok := r.lastObservation.Load(name)
+	if !ok {
+		return
+	}
+	drift.WithLabelValues(r.gvk.Group, r.gvk.Kind).Observe(time.Since(last.(time.Time)).Seconds())
+
+	r.lastObservation.Store(name, time.Now())
+}
--- a/pkg/reconciler/managed/reconciler.go
+++ b/pkg/reconciler/managed/reconciler.go
@@ -482,6 +482,8 @@ type Reconciler struct {
 
 	features feature.Flags
 
+	driftRecorder driftRecorder
+
 	// The below structs embed the set of interfaces used to implement the
 	// managed resource reconciler. We do this primarily for readability, so
 	// that the reconciler logic reads r.external.Connect(),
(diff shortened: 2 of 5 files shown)


pkg/reconciler/managed/metrics.go
```diff
@@ -0,0 +1,98 @@
...
+		DeleteFunc: func(obj interface{}) {
+			if final, ok := obj.(cache.DeletedFinalStateUnknown); ok {
+				obj = final.Obj
+			}
+			managed := obj.(resource.Managed)
+			r.lastObservation.Delete(managed.GetName())
+		},
+	})
+	if err != nil {
+		return err
```
turkenh, line 78: nit: should we wrap these (this and the above) errors with more details?

pkg/reconciler/managed/metrics.go
```diff
@@ -0,0 +1,98 @@
...
+	metrics.Registry.MustRegister(drift)
+}
+
+var subSystem = "crossplane"
+
+var (
+	drift = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Subsystem: subSystem,
+		Name:      "resource_drift_seconds",
+		Help:      "ALPHA: How long since the previous reconcile when a resource was found to be out of sync; excludes restart of the provider",
```
turkenh, line 45: > since the previous reconcile

Should this be "since the previous successful reconcile" or even "since the previous successful reconcile where the resource was observed as synced" ?

pkg/reconciler/managed/reconciler.go
```diff
@@ -1106,6 +1121,9 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		return reconcile.Result{Requeue: true}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
 	}
 
+	// record the drift after the successful update.
+	r.driftRecorder.recordUpdate(managed.GetName())
```
turkenh, line 1125: Should we also reset the timer here (i.e. call `recordUnchanged`)? I believe we rely on an upcoming reconciliation which we expect to happen after a successful update, right?

turkenh (approved):
Thank you @sttts, this is a great first step toward improving observability 💪 

Left a couple of commands, but none is blocking. 

Could you please also populate the `How has this code been tested` section in the PR description 🙏

pkg/reconciler/managed/metrics.go
sttts: fixed

pkg/reconciler/managed/reconciler.go
sttts: there would be an unchanged case afterwards. But I agree, we can directly reset.

pkg/reconciler/managed/metrics.go
sttts: fixed

ulucinar (approved):
Thanks @sttts. I will bump the crossplane-runtime version across the official providers once this PR is merged.

pkg/reconciler/managed/metrics.go
```diff
@@ -0,0 +1,100 @@
...
+
+	cluster cluster.Cluster
+}
+
+var _ manager.Runnable = &driftRecorder{}
+
+func (r *driftRecorder) Start(ctx context.Context) error {
+	inf, err := r.cluster.GetCache().GetInformerForKind(ctx, r.gvk)
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
sttts: done

pkg/reconciler/managed/reconciler.go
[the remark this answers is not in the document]
sttts: background: this metric is really only useful if it is an upper bound.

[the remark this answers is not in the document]
sttts: @ulucinar see second commit. Is that reasonable?

pkg/reconciler/managed/metrics.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+func init() {
+	metrics.Registry.MustRegister(drift)
+}
+
+var subSystem = "crossplane"
+
+var (
+	drift = prometheus.NewHistogramVec(prometheus.HistogramOpts{
```
negz, line 43: Do we have a precedent of using the prometheus client elsewhere? It might be nice to use Opencensus (e.g. to have a single metrics library if we wanted to add tracing in future). I'm guessing we may not get a choice in the matter, since we presumably need to extend the metrics controller-runtime is exposing (and I think it uses the Prometheus client per https://github.com/kubernetes-sigs/controller-runtime/issues/305).

pkg/reconciler/managed/reconciler.go
```diff
@@ -681,6 +721,11 @@ func NewReconciler(m manager.Manager, of resource.ManagedKind, o ...ReconcilerOp
 		ro(r)
 	}
 
+	if err := m.Add(&r.driftRecorder); err != nil {
+		r.log.Info("unable to register drift recorder with controller manager", "error", err)
```
negz, line 725: Should this be a debug log? What would a typical user do with/about this log when they saw it?

pkg/reconciler/managed/reconciler.go
```diff
@@ -892,7 +937,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		log = log.WithValues("deletion-timestamp", managed.GetDeletionTimestamp())
 
 		if observation.ResourceExists && policy.ShouldDelete() {
-			if err := external.Delete(externalCtx, managed); err != nil {
+			err := external.Delete(externalCtx, managed)
+			var retryAfterErr RetryAfterError
```
negz, line 941: Nit: This long variable name is redundant given its limited scope and the fact that its type `RetryAfterError` is written immediately after its name.

https://github.com/crossplane/crossplane/tree/master/contributing#use-descriptive-variable-names-sparingly

pkg/reconciler/managed/metrics.go
```diff
@@ -0,0 +1,100 @@
...
+// and records the time since on update as a metric. This represents an upper
+// bound for the duration the drift existed.
+type driftRecorder struct {
+	lastObservation sync.Map
+	gvk             schema.GroupVersionKind
+
+	cluster cluster.Cluster
+}
+
+var _ manager.Runnable = &driftRecorder{}
```
negz, line 61: Nit: I prefer to add these "does it satisfy the interface" checks to test files, since they are a kind of test.

pkg/reconciler/managed/metrics.go
```diff
@@ -0,0 +1,100 @@
...
+	})
+	if err != nil {
+		return errors.Wrap(err, "cannot add delete event handler to informer for drift recorder")
+	}
+	defer inf.RemoveEventHandler(registered) //nolint:errcheck // this happens on destruction. We cannot do anything anyway.
+
+	<-ctx.Done()
+
+	return nil
+}
```
negz, line 86: Is it worth adding a unit test for this?

pkg/reconciler/managed/reconciler.go
sttts: Would even promote it to an error if I could without logging interface. This signals that something is broken. It's not debug output.

pkg/reconciler/managed/metrics.go
sttts: Would argue the inverse: they are for the reader aka documentation. Testing is a side-effect.

pkg/reconciler/managed/metrics.go
sttts: what do you want to test? It's nearly completely plumbing.

pkg/reconciler/managed/reconciler.go
sttts: done

pkg/reconciler/managed/reconciler.go
sttts: Alternative: we add an error as return value as a breaking change.

pkg/reconciler/managed/metrics.go
sttts: added a unit test for the two methods.

pkg/reconciler/managed/metrics.go
negz: I'm not really suggesting anything. 🙂 I'm essentially asking whether using Opencensus instead would be possible and advisable, but I think I know the answer is "no, not possible".

pkg/reconciler/managed/reconciler.go
negz: Some rationale as to why we don't have error log level: https://dave.cheney.net/2015/11/05/lets-talk-about-logging

pkg/reconciler/managed/metrics.go
negz: They're both, and documentation like examples typically goes in tests. Not a hill I'll die on, but precedent elsewhere is to put these in test files.

pkg/reconciler/managed/reconciler.go
negz: I think my original comment still stands though - how would a user fix this? What would cause this? In the spirit of the better errors guide - what does "unable to register drift recorder" mean?

Keeping at info level is fine with me if it's meaningful and actionable to the user. Channeling the good errors guide, I don't personally think I'd know what "unable to register drift recorder" meant if I hadn't read this PR, or what I should do about it.

negz:
This was merged via https://github.com/crossplane/crossplane-runtime/pull/683.

PR #491, opened 2023-07-31
pkg/resource: consistent applicators

### LINKED ISSUE #483: pkg/resource: applicator breaks optimistic concurrency and overwrites values
### What happened?

Looking at the claim controller, updating the XR (`cp` in the code), the following race with other actors in the system (controllers, user with kubectl) exists:

1. reconciler [gets latest XR state](https://github.com/crossplane/crossplane/blob/7053ceb47dc867de7904838a1e87924079da8aba/internal/controller/apiextensions/claim/reconciler.go#L395) `cp` 
2. reconciler updates in the `cp` whatever it wants to update
3. reconciler [calls Apply](https://github.com/crossplane/crossplane/blob/7053ceb47dc867de7904838a1e87924079da8aba/internal/controller/apiextensions/claim/reconciler.go#L523) with `cp` 
   - applicator [gets current state via client](https://github.com/crossplane/crossplane-runtime/blob/4374a3b5d1ff58214b118a7cea641230bfbc785f/pkg/resource/api.go#L186)
   - applicator [sets RV](https://github.com/crossplane/crossplane-runtime/blob/4374a3b5d1ff58214b118a7cea641230bfbc785f/pkg/resource/api.go#L203C2-L203C68) `cp.RV = current.RV` 
   - applicator [calls update](https://github.com/crossplane/crossplane-runtime/blob/4374a3b5d1ff58214b118a7cea641230bfbc785f/pkg/resource/api.go#L204) with cp

Assuming between 1 and 3a some client updates the XR out of band, e.g. some other controller, or kubectl. 3c will wipe out that change, won't it?

Above is just one example for that issue. The apply func is widely used in Crossplane.

### How can we reproduce it?

This will likely show up as subtle bugs as this is some kind of time travel, or controllers fighting for their desired state of the world and racing with their informers.

Crossplane version: main

sttts:

This PR does two things:

1. the update applicator is deprecated. The override of the resource version is removed, i.e. the update applicator will correctly conflict with any other actor writing to the object when the base version of the reconcile (from the informer cache) is outdated. With that the updates follow the optimistic concurrency protocol of the API: get, modify object, update, retry on conflict.
2. the patching applicator is changed to compute patches between the base version of the reconcile (from the informer cache) and the object to be applied. Because the applicator has no access to the former, it does another call against the cached client to get the current version. If that matches the resource version of the object to be applied, we are fine and can apply the computed patch to the apiserver. If the resource version differs, this is equivalent to a conflict error because the world has moved on during reconcile. That's returned as a conflict error to the reconcile loop.

Fixes #483: 

- [x] proof PR for crossplane created: https://github.com/crossplane/crossplane/pull/4562

### DIFF
--- a/pkg/resource/api.go
+++ b/pkg/resource/api.go
@@ -127,9 +127,6 @@ func (a *APIUpdatingApplicator) Apply(ctx context.Context, o client.Object, ao .
 		}
 	}
 
-	// NOTE(hasheddan): we must set the resource version of the desired object
-	// to that of the current or the update will always fail.
-	m.SetResourceVersion(current.(metav1.Object).GetResourceVersion())
 	return errors.Wrap(a.client.Update(ctx, m), "cannot update object")
 }
 
--- a/pkg/resource/api.go
+++ b/pkg/resource/api.go
@@ -17,10 +17,14 @@ limitations under the License.
 package resource
 
 import (
+	"bytes"
 	"context"
 
+	jsonpatch "github.com/evanphx/json-patch"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/runtime/serializer/json"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -54,7 +58,7 @@ func NewAPIPatchingApplicator(c client.Client) *APIPatchingApplicator {
 // Apply changes to the supplied object. The object will be created if it does
 // not exist, or patched if it does. If the object does exist, it will only be
 // patched if the passed object has the same or an empty resource version.
-func (a *APIPatchingApplicator) Apply(ctx context.Context, obj client.Object, ao ...ApplyOption) error { //nolint:gocyclo // the logic here is crucial and deserves to stay in one method
+func (a *APIPatchingApplicator) Apply(ctx context.Context, obj client.Object, ao ...ApplyOption) error {
 	if obj.GetName() == "" && obj.GetGenerateName() != "" {
 		return a.client.Create(ctx, obj)
 	}
@@ -100,6 +104,50 @@ func groupResource(c client.Client, o client.Object) (schema.GroupResource, erro
 		return schema.GroupResource{}, errors.Wrapf(err, "cannot determine group resource of %v", gvk)
 	}
 	return m.Resource.GroupResource(), nil
+}
+
+var emptyScheme = runtime.NewScheme() // no need to recognize any types
+var jsonSerializer = json.NewSerializerWithOptions(json.DefaultMetaFactory, emptyScheme, emptyScheme, json.SerializerOptions{})
+
+// AdditiveMergePatchApplyOption returns an ApplyOption that makes
+// the Apply additive in the sense of a merge patch without null values. This is
+// the old behavior of the APIPatchingApplicator.
+//
+// This only works with a desired object of type *unstructured.Unstructured.
+//
+// Deprecated: replace with Server Side Apply.
+func AdditiveMergePatchApplyOption(_ context.Context, current, desired runtime.Object) error {
+	// set GVK uniformly to the desired object to make serializer happy
+	currentGVK, desiredGVK := current.GetObjectKind().GroupVersionKind(), desired.GetObjectKind().GroupVersionKind()
+	if !desiredGVK.Empty() && currentGVK != desiredGVK {
+		return errors.Errorf("cannot apply %v to %v", desired.GetObjectKind().GroupVersionKind(), current.GetObjectKind().GroupVersionKind())
+	}
+	desired.GetObjectKind().SetGroupVersionKind(currentGVK)
+
+	// merge `desired` additively with `current`
+	var currentBytes, desiredBytes bytes.Buffer
+	if err := jsonSerializer.Encode(current, &currentBytes); err != nil {
+		return errors.Wrapf(err, "cannot marshal current %s", HumanReadableReference(nil, current))
+	}
+	if err := jsonSerializer.Encode(desired, &desiredBytes); err != nil {
+		return errors.Wrapf(err, "cannot marshal desired %s", HumanReadableReference(nil, desired))
+	}
+	mergedBytes, err := jsonpatch.MergePatch(currentBytes.Bytes(), desiredBytes.Bytes())
+	if err != nil {
+		return errors.Wrapf(err, "cannot merge patch to %s", HumanReadableReference(nil, desired))
+	}
+
+	// write merged object back to `desired`
+	if _, _, err := jsonSerializer.Decode(mergedBytes, nil, desired); err != nil {
+		return errors.Wrapf(err, "cannot unmarshal merged patch to %s", HumanReadableReference(nil, desired))
+	}
+
+	// restore empty GVK for typed objects
+	if _, isUnstructured := desired.(runtime.Unstructured); !isUnstructured {
+		desired.GetObjectKind().SetGroupVersionKind(schema.GroupVersionKind{})
+	}
+
+	return nil
 }
 
 // An APIUpdatingApplicator applies changes to an object by either creating or
--- a/pkg/resource/api.go
+++ b/pkg/resource/api.go
@@ -18,11 +18,9 @@ package resource
 
 import (
 	"context"
-	"encoding/json"
 
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -33,6 +31,12 @@ import (
 // Error strings.
 const (
 	errUpdateObject = "cannot update object"
+
+	// taken from k8s.io/apiserver. Not crucial to match, but for uniformity it
+	// better should.
+	// TODO(sttts): import from k8s.io/apiserver/pkg/registry/generic/registry when
+	//              kube has updated otel dependencies post-1.28.
+	errOptimisticLock = "the object has been modified; please apply your changes to the latest version and try again"
 )
 
 // An APIPatchingApplicator applies changes to an object by either creating or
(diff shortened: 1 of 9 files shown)


negz:
@sttts Could you update the PR description with a little more detail, please. Between your initial issue description and my lengthy comment we touch on quite a few things, and it's not obvious to me what parts of https://github.com/crossplane/crossplane-runtime/issues/483 this is intending to solve and how.

pkg/resource/api.go
```diff
@@ -38,7 +37,8 @@ const (
 // An APIPatchingApplicator applies changes to an object by either creating or
 // patching it in a Kubernetes API server.
 type APIPatchingApplicator struct {
-	client client.Client
+	client      client.Client
+	optionalLog logging.Logger // can be nil
```
negz, line 41: There's a `logging.NopLogger` implementation that does nothing when called. Typically instead of an `if logger == nil` check we'd have `NewAPIPatchingApplicator` default to using `logging.NopLogger` such that log calls are discarded.

sttts: this is intentional as diffing is quite expensive.

sttts:
This needs a proof PR against Crossplane. I have the suspicion that it breaks the composite reconciler as it relies on the inconsistent, additive behaviour.

negz:
> This needs a proof PR against Crossplane

Agreed. I haven't had time to review closely. I like the direction but suspect it would be breaking in a few places.

pkg/resource/api.go
```diff
@@ -47,90 +47,118 @@ func NewAPIPatchingApplicator(c client.Client) *APIPatchingApplicator {
 	return &APIPatchingApplicator{client: c}
 }
 
+// WithLogger sets the logger on the APIPatchingApplicator.
+func (a *APIPatchingApplicator) WithLogger(l logging.Logger) *APIPatchingApplicator {
+	a.optionalLog = l
+	return a
+}
```
negz, line 54: This is introducing a second way to pass options (builder style) in addition to the variadic `ApplyOption` style supported by the `Apply` method.

Given that:
* As you mention `LogDiff` is expensive, so we want it to be done optionally.
* `LogDiff` is called right after where we loop over the variadic option functions.
* `LogDiff` has a very similar signature to the variadic option functions.

Could `LogDiff` itself be an option - e.g. `p.Apply(ctx, o, WithLogDiff(l logging.Logger))`?

pkg/resource/api.go
```diff
@@ -47,90 +47,118 @@ func NewAPIPatchingApplicator(c client.Client) *APIPatchingApplicator {
...
 		// TODO(negz): Apply ApplyOptions here too?
-		return errors.Wrap(a.client.Create(ctx, o), "cannot create object")
+		return errors.Wrap(a.client.Create(ctx, obj), "cannot create object")
 	}
 	if err != nil {
 		return errors.Wrap(err, "cannot get object")
 	}
 
+	// Note: this check would ideally not be necessary if the Apply signature
+	// had a current object that we could us for the diff. But we have no
```
negz, line 75:
```suggestion
	// had a current object that we could use for the diff. But we have no
```

pkg/resource/api.go
```diff
@@ -47,90 +47,118 @@ func NewAPIPatchingApplicator(c client.Client) *APIPatchingApplicator {
...
+	// current and for consistency of the patch it matters that the object we
+	// get above is the one that was originally used.
```
negz, line 77: Could we ab(use) the apply options to optionally provide the `current` object? e.g. `p.Apply(ctx, desired, WithCurrent(observed))`.


Presumably even if there was we'd still need to fall back to

negz:
> This PR does two things

@sttts +1 for deprecating the `APIUpdatingApplicator`. That makes sense to me.

With regard to the `APIPatchingApplicator` changes, my understanding (as written up in https://github.com/crossplane/crossplane-runtime/issues/483#issuecomment-1636432560) is that this applicator does not have an optimistic concurrency issue at the moment. Is that understanding wrong?

Is the fact that the `APIPatchingApplicator` now computes patches between the base version of the resource and the object to be applied addressing an optimistic concurrency issue? Or is it instead intended to allow us to support deleting object fields, since it now creates a proper merge patch?

sttts:
> applicator does not have an optimistic concurrency issue at the moment. Is that understanding wrong?

It has a consistent issue though :) It does not apply what you tell it to apply. It additively merged your changes into the version on the server.

pkg/resource/api.go
sttts: possibly. Note though that our clients are cached. So it doesn't make a big difference.

pkg/resource/api.go
sttts: I had that in the beginning and forgot why I reverted it.

pkg/resource/api.go
```diff
@@ -47,90 +47,118 @@ func NewAPIPatchingApplicator(c client.Client) *APIPatchingApplicator {
...
-	m, ok := o.(metav1.Object)
-	if !ok {
-		return errors.New("cannot access object metadata")
-	}
-
-	if m.GetName() == "" && m.GetGenerateName() != "" {
-		return errors.Wrap(a.client.Create(ctx, o), "cannot create object")
+func (a *APIPatchingApplicator) Apply(ctx context.Context, obj client.Object, ao ...ApplyOption) error { //nolint:gocyclo // the logic here is crucial and deserves to stay in one method
+	if obj.GetName() == "" && obj.GetGenerateName() != "" {
+		return errors.Wrap(a.client.Create(ctx, obj), "cannot create object")
```
pedjak, line 61: in case of error, would be nice to give some details about `obj` that could not be created?

```go
return errors.Wrapf(a.client.Create(ctx, obj), "cannot create object %s of %s", obj.GetGenerateName(), obj.GetObjectKind().GroupVersionKind())
```

pkg/resource/api.go
```diff
@@ -47,90 +47,118 @@ func NewAPIPatchingApplicator(c client.Client) *APIPatchingApplicator {
...
 	}
 
-	desired := o.DeepCopyObject()
-
-	err := a.client.Get(ctx, types.NamespacedName{Name: m.GetName(), Namespace: m.GetNamespace()}, o)
+	current := obj.DeepCopyObject().(client.Object)
+	err := a.client.Get(ctx, types.NamespacedName{Name: obj.GetName(), Namespace: obj.GetNamespace()}, current)
 	if kerrors.IsNotFound(err) {
 		// TODO(negz): Apply ApplyOptions here too?
```
pedjak, line 67: does this TODO still holds @negz ?

pkg/logging/logging.go
```diff
@@ -92,3 +93,20 @@ func (l logrLogger) Debug(msg string, keysAndValues ...any) {
 func (l logrLogger) WithValues(keysAndValues ...any) Logger {
 	return logrLogger{log: l.log.WithValues(keysAndValues...)} //nolint:logrlint // False positive - logrlint thinks there's an odd number of args.
 }
+
+// ForResource returns logging values for a resource.
+func ForResource(object client.Object) []string {
```
pedjak, line 98: it seems that this is not used anywhere in the code?

pkg/resource/api.go
```diff
@@ -47,90 +47,118 @@ func NewAPIPatchingApplicator(c client.Client) *APIPatchingApplicator {
...
+	}
+
 	for _, fn := range ao {
-		if err := fn(ctx, o, desired); err != nil {
+		if err := fn(ctx, current, obj); err != nil {
 			return err
 		}
 	}
 
+	if err := LogDiff(a.optionalLog, current, obj); err != nil {
```
pedjak, line 92: given that diff compute is expensive, could we avoid doing it twice, in case when `optionalLog` is set? It is computed by `client.MergeFromWithOptions(current, client.MergeFromWithOptimisticLock{}))` anyway, so perhaps we can create a patch wrapper around it to cache the computed value?

pkg/logging/logging.go
sttts: but should be, see you comment above about the error context.

sttts: added

pkg/resource/api.go
sttts: out of scope of the PR

pkg/resource/api.go
sttts: good idea. done

pkg/resource/api.go
sttts: Fixed

pkg/resource/api.go
sttts: Had added that context and removed it again because those error messages already contain the GR and name.

sttts:
Addressed comments.

Also extended tests.

pkg/resource/api.go
```diff
@@ -18,119 +18,166 @@ package resource
...
 // Apply changes to the supplied object. The object will be created if it does
 // not exist, or updated if it does.
-func (a *APIUpdatingApplicator) Apply(ctx context.Context, o client.Object, ao ...ApplyOption) error {
-	m, ok := o.(Object)
-	if !ok {
-		return errors.New("cannot access object metadata")
+func (a *APIUpdatingApplicator) Apply(ctx context.Context, obj client.Object, ao ...ApplyOption) error {
+	if obj.GetName() == "" && obj.GetGenerateName() != "" {
+		log := a.log.WithValues(logging.ForResource(obj))
+		log.Info("creating object")
```
pedjak, line 151: perhaps debug level?

pkg/resource/api.go
```diff
@@ -18,119 +18,166 @@ package resource
...
-	// to that of the current or the update will always fail.
-	m.SetResourceVersion(current.(metav1.Object).GetResourceVersion())
-	return errors.Wrap(a.client.Update(ctx, m), "cannot update object")
+	// log diff
+	patch := client.MergeFromWithOptions(current, client.MergeFromWithOptimisticLock{})
+	patchBytes, err := patch.Data(obj)
+	if err != nil {
+		return errors.Wrapf(err, "failed to diff %s", HumanReadableReference(a.client, obj))
+	}
+	log := a.log.WithValues(logging.ForResource(obj))
```
pedjak, line 177: could we declar this at the top of the function and then remove line 150?

pkg/resource/api.go
```diff
@@ -18,119 +18,166 @@ package resource
...
+	log.WithValues("diff", string(patchBytes)).Info("updating object")
+
+	return a.client.Update(ctx, obj)
```
pedjak, line 180: just wonder in case of error, would be able to see something in logs and correlate it with the previous log line? Also, do we really want to emit log line on INFO?

pkg/resource/api.go
```diff
@@ -20,117 +20,199 @@ import (
...
-func (p *patch) Type() types.PatchType                { return types.MergePatchType }
-func (p *patch) Data(_ client.Object) ([]byte, error) { return json.Marshal(p.from) }
+// AdditiveMergePatchApplyOption returns an ApplyOption that makes
+// the Apply additive in the sense of a merge patch without null values. This is
+// the old behavior of the APIPatchingApplicator.
+//
+// This only works with a desired object of type *unstructured.Unstructured.
+//
+// Deprecated: replace with Server Side Apply.
+func AdditiveMergePatchApplyOption(ctx context.Context, current, desired runtime.Object) error {
```
sttts, line 135: @pedjak @negz this should get us over the line in Crossplane until we have SSA implemented.

pkg/resource/api.go
sttts: No, this is intentional. This is progress output, not debug.

pkg/resource/api.go
sttts: done

pkg/resource/api.go
sttts: the clients produce errors with context. No need to add another one.

Error processing and error logging I leave to the caller. controller-runtime will also log if the error is passed up. If the error is expected, the caller will do the thing and not log. We don't know here.

sttts: But, I would like to see this in action in a proof PR in Crossplane.

turkenh:
> > Or is it instead intended to allow us to support deleting object fields, since it now creates a proper merge patch?
> 
> Yes, it is.
> 
> But this of course means that it wipes everything not in the passed object. E.g. if you override the spec in the composition controller code to be equal to that in the `Composition`, it will wipe everything that a provider adds to the spec. To work around that we have to switch the compisition controller to server side apply. But that means that we will get rid of the patching applicator in that controller and talk directly via SSA to the server.

The logic in the composite controller (also in the claim controller) heavily relies on additive behavior currently. Hence, it wouldn't be possible to switch to the patching applicator introduced in this PR without a good amount of refactoring/testing, which could be a similar amount of effort as switching to Server Side Apply. IIUC, this will work quite similar to the extract flow mentioned [here](https://kubernetes.io/blog/2022/10/20/advanced-server-side-apply/#get-modify-put-patch-controllers).
So, I am wondering if it is worth the effort to switch to this or if should we simply invest in Server Side Apply instead 🤔

pkg/resource/api.go
turkenh: So, would this mean we would have one log line per created object during their creation? Which I believe would still be too verbose for default logging and not inline with the existing logging behavior in XP.

Also, there is another `client.Create` line below, why don't we have a similar log line below? Are they different cases?

sttts: > Also, there is another client.Create line below, why don't we have a similar log line below? Are they different cases?

fixed.

> So, would this mean we would have one log line per created object during their creation? 

yes

> Which I believe would still be too verbose for default logging and not inline with the existing logging behavior in XP

It's intentionally not inline with the existing logging behaviour. The current behaviour is not useful.

Counter question: why is it ok to log the same in the apiserver, but not where it matters for understanding what a component does?

How many creates or updates do we expect in a reasonably big XP installation on average? My claim: the system is in steady state 99% of the time without any update to the apiserver.

pkg/resource/api.go
```diff
@@ -20,117 +20,221 @@ import (
...
-func (p *patch) Data(_ client.Object) ([]byte, error) { return json.Marshal(p.from) }
+// AdditiveMergePatchApplyOption returns an ApplyOption that makes
+// the Apply additive in the sense of a merge patch without null values. This is
+// the old behavior of the APIPatchingApplicator.
+//
+// This only works with a desired object of type *unstructured.Unstructured.
+//
+// Deprecated: replace with Server Side Apply.
+func AdditiveMergePatchApplyOption(_ context.Context, current, desired runtime.Object) error {
+	u, ok := desired.(*unstructured.Unstructured)
```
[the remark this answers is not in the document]
sttts: ErrTooManyWrappers. Fixing.

sttts:
> The logic in the composite controller (also in the claim controller) heavily relies on additive behavior currently. Hence, it wouldn't be possible to switch to the patching applicator introduced in this PR without a good amount of refactoring/testing, which could be a similar amount of effort as switching to Server Side Apply. IIUC, this will work quite similar to the extract flow mentioned [here](https://kubernetes.io/blog/2022/10/20/advanced-server-side-apply/#get-modify-put-patch-controllers).
> So, I am wondering if it is worth the effort to switch to this or if should we simply invest in Server Side Apply instead 🤔

Hence, `AdditiveMergePatchApplyOption` to make this old behaviour explicit. We can fix the compositor with SSA in a follow-up.

pkg/resource/api.go
```diff
@@ -20,117 +20,221 @@ import (
...
-		if err := fn(ctx, current, m); err != nil {
-			return err
+		if err := fn(ctx, current, obj); err != nil {
+			return errors.Wrapf(err, "apply option failed for %s", HumanReadableReference(a.client, obj))
 		}
 	}
 
-	// NOTE(hasheddan): we must set the resource version of the desired object
-	// to that of the current or the update will always fail.
-	m.SetResourceVersion(current.(metav1.Object).GetResourceVersion())
```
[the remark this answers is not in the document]
sttts: the object passed to the applicator here is based on a `Get`, i.e. has a resource version. Any other use of the applicator is invalid.

pkg/resource/api.go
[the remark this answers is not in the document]
sttts: done

pkg/resource/resource.go
```diff
@@ -190,13 +191,24 @@ func IsAPIErrorWrapped(err error) bool {
...
 	return c.Status == corev1.ConditionTrue
 }
 
-// An Applicator applies changes to an object.
+// An Applicator applies changes to an object. The passed object is expected to
+// be complete, i.e. not partial.
 type Applicator interface {
+	// Apply updates the given object to exactly the given state. It conflicts
+	// if the resource version stored on the apiserver does not match anymore
+	// the one in the given object.
```
sttts, line 211: @turkenh this comment and the one above are crucial. Any other behaviour must go into an `ApplyOption`.

pkg/resource/api.go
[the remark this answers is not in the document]
turkenh: We pass `current` (which is a deepcopy of obj) to the `Get`. So, the object passed to Update does not have a resource version.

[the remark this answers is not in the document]
turkenh: Ah ok, you mean "should be based on a `Get`" ?

pedjak:
> The logic in the composite controller (also in the claim controller) heavily relies on additive behavior currently. Hence, it wouldn't be possible to switch to the patching applicator introduced in this PR without a good amount of refactoring/testing, which could be a similar amount of effort as switching to Server Side Apply.

Independently from the strategy, we need to create crossplane e2e tests, covering at least major usecases advocated to users - only then we can switch with confidence to an improved applicator/server-side apply.

pkg/resource/api.go
turkenh: > It's intentionally not inline with the existing logging behaviour. The current behaviour is not useful.

I believe this deserves some upfront discussion and alignment as it contradicts the current [observability guide](https://github.com/crossplane/crossplane/blob/master/contributing/guide-observability.md#in-practice), for example:

> _Almost nothing is worth logging at info level. When deciding which logging level to use, consider a production deployment of Crossplane reconciling tens or hundreds of managed resources. If in doubt, pick debug. You can easily increase the log level later if it proves warranted._

And I personally would agree with the current approach we're following.

sttts: Have moved the logging parts into https://github.com/crossplane/crossplane-runtime/pull/532. Please continue discussion there. And yes, we need a bigger discussion around what to log.

> I believe events and conditions are better fit for emitting "per instance" information.

I disagree. Neither events nor conditions are for CRUD or reconcile logging. Logs are lower level, component level, not API level.

lsviben:
Hey @sttts , is this PR still relevant?

sttts:
The applicators are semantically as broken as before. So yes, it's relevant.

negz:
Per my update to the issue this attempts to fix - I think I'd rather just work toward removing `ClientApplicator`.

PR #492, opened 2023-07-31
Ignore k8s secret not found when ESS enabled

ezgidemirel:

In this change, we're ignoring k8s secret not found errors when the external secret store feature is enabled. In Vault implementation, we're not returning any error if the secret doesn't exist. To have the same behavior and allow secret creation at later stages, we start not to return any error in the k8s store as well. This is the root cause of this [bug](https://github.com/crossplane/crossplane/issues/4191). 

Fixes #

I have:

- consumed the change in crossplane
- enabled both the external store and xfn features 
- created a claim which uses Vault as a store
- created a claim which uses k8s as a store
- disabled xfn feature
- created a claim which uses Vault as a store
- created a claim which uses k8s as a store

### DIFF
--- a/pkg/connection/store/kubernetes/store.go
+++ b/pkg/connection/store/kubernetes/store.go
@@ -93,7 +93,7 @@ func buildClient(ctx context.Context, local client.Client, cfg v1.SecretStoreCon
 // ReadKeyValues reads and returns key value pairs for a given Kubernetes Secret.
 func (ss *SecretStore) ReadKeyValues(ctx context.Context, n store.ScopedName, s *store.Secret) error {
 	ks := &corev1.Secret{}
-	if err := ss.client.Get(ctx, types.NamespacedName{Name: n.Name, Namespace: ss.namespaceForSecret(n)}, ks); err != nil {
+	if err := ss.client.Get(ctx, types.NamespacedName{Name: n.Name, Namespace: ss.namespaceForSecret(n)}, ks); resource.IgnoreNotFound(err) != nil {
 		return errors.Wrap(err, errGetSecret)
 	}
 	s.Data = ks.Data
--- a/pkg/connection/store/kubernetes/store_test.go
+++ b/pkg/connection/store/kubernetes/store_test.go
@@ -102,20 +102,37 @@ func TestSecretStoreReadKeyValues(t *testing.T) {
 			args: args{
 				client: resource.ClientApplicator{
 					Client: &test.MockClient{
-						MockGet: test.NewMockGetFn(nil, func(obj client.Object) error {
+						MockGet: func(ctx context.Context, key client.ObjectKey, obj client.Object) error {
+							if key.Name != fakeSecretName || key.Namespace != fakeSecretNamespace {
+								return errors.New("unexpected secret name or namespace to get the secret")
+							}
 							*obj.(*corev1.Secret) = corev1.Secret{
 								Data: fakeKV(),
 							}
 							return nil
-						}),
+						},
 					},
 				},
 				n: store.ScopedName{
-					Name: fakeSecretName,
+					Name:  fakeSecretName,
+					Scope: fakeSecretNamespace,
 				},
 			},
 			want: want{
 				result: store.KeyValues(fakeKV()),
+			},
+		},
+		"SecretNotFound": {
+			reason: "Should return nil as an error if secret is not found",
+			args: args{
+				client: resource.ClientApplicator{
+					Client: &test.MockClient{
+						MockGet: test.NewMockGetFn(kerrors.NewNotFound(schema.GroupResource{}, "")),
+					},
+				},
+			},
+			want: want{
+				err: nil,
 			},
 		},
 	}
@@ -126,6 +143,7 @@ func TestSecretStoreReadKeyValues(t *testing.T) {
 			}
 
 			s := &store.Secret{}
+			s.ScopedName = tc.args.n
 			err := ss.ReadKeyValues(context.Background(), tc.args.n, s)
 			if diff := cmp.Diff(tc.want.err, err, test.EquateErrors()); diff != "" {
 				t.Errorf("\n%s\nss.ReadKeyValues(...): -want error, +got error:\n%s", tc.reason, diff)


pkg/connection/store/kubernetes/store.go
```diff
@@ -93,7 +93,7 @@ func buildClient(ctx context.Context, local client.Client, cfg v1.SecretStoreCon
 // ReadKeyValues reads and returns key value pairs for a given Kubernetes Secret.
 func (ss *SecretStore) ReadKeyValues(ctx context.Context, n store.ScopedName, s *store.Secret) error {
 	ks := &corev1.Secret{}
-	if err := ss.client.Get(ctx, types.NamespacedName{Name: n.Name, Namespace: ss.namespaceForSecret(n)}, ks); err != nil {
+	if err := ss.client.Get(ctx, types.NamespacedName{Name: n.Name, Namespace: ss.namespaceForSecret(n)}, ks); resource.IgnoreNotFound(err) != nil {
```
pedjak, line 96: if not found, then what would be the value of `ks.Data`?

could we add a unit test for the new behavior?

If ESS is not enabled, then should we return the error?

ezgidemirel: > > It will be an empty map.
> 
> Can we update the functions docs to reflect that?

It shouldn't be necessary. In real world, we shouldn't end up with an empty secret. This change is done to proceed with an intermediate stage where the claim/composite secret is not created yet but it's fetched by the PTF [composer](https://github.com/crossplane/crossplane/blob/2df2202d4ddc846bfdcf7f1cce7bc82c55d28010/internal/controller/apiextensions/composite/composition_ptf.go#L286) to observe the state. Later in the reconciliation loop, the secret will be created as usual and we don't need to handle not found errors.

pkg/connection/store/kubernetes/store_test.go
```diff
@@ -118,6 +118,22 @@ func TestSecretStoreReadKeyValues(t *testing.T) {
 				result: store.KeyValues(fakeKV()),
 			},
 		},
+		"SecretNotFound": {
+			reason: "Should return an empty secret and without any error if the secret is not found",
+			args: args{
+				client: resource.ClientApplicator{
+					Client: &test.MockClient{
+						MockGet: test.NewMockGetFn(kerrors.NewNotFound(schema.GroupResource{}, "")),
```
pedjak, line 126: We should return error only if we try to retrieve the secret with the name equal to `fakeSecretName`. Otherwise, if we do not perform this check, this tests is going to pass even if `ReadKeyValues` does not properly use the passed `ScopedName`

pkg/connection/store/kubernetes/store_test.go
```diff
@@ -118,6 +118,22 @@ func TestSecretStoreReadKeyValues(t *testing.T) {
...
+					},
+				},
+				n: store.ScopedName{
+					Name: fakeSecretName,
+				},
+			},
+			want: want{
+				result: nil,
```
pedjak, line 134: if the test name says that `return an empty secret`, should then the result be an empty map? If it should remain `nil`, then perhaps we should rephrase the reason above to reflect that?

pedjak (changes requested):

pkg/connection/store/kubernetes/store_test.go
ezgidemirel: What we `want` is not the secret, but the error. Therefore, the test is correct

ezgidemirel: > What we `want` is not the secret, but the error. Therefore, the test is correct

Nevermind, that I thought is not the one I wrote. I'll update it.

pedjak: Not sure that I follow: in this test `ReadKeyValues` is not going to return the error, and then we compare secrets data with `want.result`. Do we consider an empty secret the one with `Data` field being set to `nil` or to an empty map?

ezgidemirel: The zero value of a map is nil. I observed different behaviors like while printing it, I get `map[]` but in the `cmp.Diff` method, it considers it as a `nil`.

pkg/connection/store/kubernetes/store_test.go
phisco: yeah, that's the issue with mocking, the mock is never going to be 100% precise with the real behaviour. I don't think it's strictly required here, but could be a separate test case maybe.

pkg/connection/store/kubernetes/store.go
phisco: or is the caller relying on the fact that Metadata is set even on not found?

ezgidemirel: I think if we cannot find the secret on the store, we should overwrite the given secret fields with empty/nil values.

pkg/connection/store/kubernetes/store_test.go
pedjak: mocks should react for only specific inputs. We are adding here the new test, so why not making it more tight here?

phisco: +1 for adding that to the SuccessfulRead test.

phisco: Actually, probably the right thing to do would be to add a test case getting a specific secret and returning an errBoom error only for the specified name. That would ensure we are actually calling it with the right name.

pedjak: > The NewMockGetFn function does not use the object key. Which means we cannot check the scope name with that.

yeah, but we can do something like:

```go
Client: &test.MockClient{
  MockGet: func(ctx context.Context, key client.ObjectKey, obj client.Object) error {
	if key.Name == fakeSecretName {
		return kerrors.NewNotFound(schema.GroupResource{}, "")
	}
	return errors.Errorf("Should return not found error actually for %s", key)
},
},
```

ezgidemirel: I know we can. But my point was, `NewMockGetFn` function is used for dozens of places without actually checking the name or namespace and I think it's OK. IMHO, if we're mocking an interface, we don't need to worry about the mocker itself. The critical part is checking the behavior after getting the result from the mock. 

Anyways, I updated the `SuccessfulRead` test.

pedjak: > I know we can. But my point was, NewMockGetFn function is used for dozens of places without actually checking the name or namespace and I think it's OK. IMHO, if we're mocking an interface, we don't need to worry about the mocker itself. The critical part is checking the behavior after getting the result from the mock.

We are not mocking just interface, we are actually mocking its invocations, and we in essence should define a number of invocations as set of (input params, results).

The code under tests should invoke an interface with proper input params, and should react properly on the obtained results. If we do not check the input params, then we verify just partially the code behavior.

pedjak (approved):

phisco (approved):

PR #504, opened 2023-08-08
pkg.parser: add the ability to parse manifests with comments

muvaf:

Adds the ability to skip the files with no content, which was determined with having only white spaces but files where all lines are commented out should be skipped as well.

I have:

Manually and with new unit tests added.

### DIFF
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -22,7 +22,6 @@ import (
 	"context"
 	"io"
 	"strings"
-	"unicode"
 
 	"github.com/spf13/afero"
 	corev1 "k8s.io/api/core/v1"
@@ -91,7 +90,7 @@ func New(meta, obj ObjectCreaterTyper) *PackageParser {
 // decode objects recognized by the meta scheme, then attempts to decode objects
 // recognized by the object scheme. Objects not recognized by either scheme
 // return an error rather than being skipped.
-func (p *PackageParser) Parse(_ context.Context, reader io.ReadCloser) (*Package, error) { //nolint:gocyclo // Only at 11.
+func (p *PackageParser) Parse(_ context.Context, reader io.ReadCloser) (*Package, error) {
 	pkg := NewPackage()
 	if reader == nil {
 		return pkg, nil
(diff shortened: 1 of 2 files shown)


pkg/parser/parser.go
```diff
@@ -133,17 +130,26 @@ func (p *PackageParser) Parse(_ context.Context, reader io.ReadCloser) (*Package
...
+		trimmed := strings.TrimSpace(line)
+		if trimmed == "" || strings.HasPrefix(trimmed, "#") {
+			continue
+		}
+		lines = append(lines, line)
+	}
+	// If all left are just separators and/or terminators, then there is no
+	// content.
+	for _, c := range lines {
+		trimmed := strings.TrimSpace(c)
```
turkenh, line 147: Do we need this TrimSpace here, given we already trimmed all lines above?

pkg/parser/parser.go
```diff
@@ -133,17 +130,26 @@ func (p *PackageParser) Parse(_ context.Context, reader io.ReadCloser) (*Package
...
+		if trimmed != "---" && trimmed != "..." {
```
turkenh, line 148: nit: would it be more readable if we perform this check in the above loop and return cleaned yaml `if len(lines) > 0` ?

pkg/parser/parser.go
muvaf: In the first loop, we trim to check but don't add it to the final lines array because that'd shift all lines in the YAML to left, hence break it. So `lines` always has the original content. Maybe I should rename the function so it doesn't imply that - the only thing it does is to return empty when there is no content.

muvaf: In the first loop, the empty and commented lines are excluded but the ones that are included are untouched so we still need to trim to for an exact match.

pkg/parser/parser.go
muvaf: The separators are not removed so there might be multiple lines with just separators but you're right we don't need the second loop. I added that emptiness check to the loop above.

turkenh (approved):

PR #512, opened 2023-08-11
Cleanup deprecated code that are no longer used

turkenh:

This PR cleans up deprecated code no longer used, focusing everything under `pkg` directory.
As proposed in [this comment](https://github.com/crossplane/crossplane/issues/4060#issuecomment-1668285052), we evaluate having identical release branches for Crossplane Runtime as Crossplane. This would mean releasing `>v1` of runtime; we wanted to make this cleanup before that.

I have:

Apart from `make reviewable` on this repo, I also tested it against the following and verified that cleaned-up code is not used by any of them:

- https://github.com/crossplane/crossplane (Core Crossplane)
- https://github.com/upbound/provider-aws (A provider)
- https://github.com/upbound/upjet (A non-provider dependency)


negz (approved):
Thanks @turkenh!

PR #526, opened 2023-08-23
reconciler/managed: avoid temporary data loss to managed on annotation update

sttts:

When storing critical annotations, the retry loop always dropped changes to the `managed` resource object, i.e.
part of the reconcile loop's effects on `managed` were just lost, always. This PR changes this to try one update
of the resource without reset (aka client get call), and only after that fall back into best-effort code path. In effect, this
might save a number of (expensive) reconciles.
I have:

### DIFF
--- a/pkg/reconciler/managed/api.go
+++ b/pkg/reconciler/managed/api.go
@@ -22,6 +22,7 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 	corev1 "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/retry"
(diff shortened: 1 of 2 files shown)


pkg/reconciler/managed/api.go
```diff
@@ -172,12 +172,18 @@ func NewRetryingCriticalAnnotationUpdater(c client.Client) *RetryingCriticalAnno
 // reset to their current state according to the API server.
```
negz, line 172: Perhaps change this comment from "pending changes _are_ reset" to "pending changes _may be_ reset".

negz (approved):
LGTM, nice optimization.

negz:
Do you think it's worth "E2E" testing this? It would have to be manual at the moment, unfortunately. (i.e. Build a provider using the updated reconciler.)

sttts:
> Do you think it's worth "E2E" testing this? 

That would be a tricky e2e. Certainly doable, but on the borderline of a test belonging into e2e.

sttts:
Actually, we should only retry with a Get on conflict. Let me change that before merging. On any other error we can just keep the old object and retry.

pkg/reconciler/managed/api.go
sttts: done

sttts:
Have extended the unit tests.

pkg/reconciler/managed/api_test.go
```diff
@@ -364,26 +366,37 @@ func TestRetryingCriticalAnnotationUpdater(t *testing.T) {
...
 	cases := map[string]struct {
-		reason string
-		c      client.Client
-		args   args
-		want   error
+		reason        string
+		c             *test.MockClient
+		args          args
+		want          error
+		wantGetCalled bool
```
negz, line 379: Nit: Typically once we want more than one thing we nest them into a `want` struct (similar to `args`).

pkg/reconciler/managed/api_test.go
```diff
@@ -411,6 +424,10 @@ func TestRetryingCriticalAnnotationUpdater(t *testing.T) {
 			if diff := cmp.Diff(tc.want, got, test.EquateErrors()); diff != "" {
 				t.Errorf("\n%s\nu.UpdateCriticalAnnotations(...): -want, +got:\n%s", tc.reason, diff)
 			}
+			getCalled := tc.args.o.GetAnnotations()["getcalled"] == "true"
+			if getCalled != tc.wantGetCalled {
+				t.Errorf("\n%s\nu.UpdateCriticalAnnotations(...) calling get: -want %v, +got %v", tc.reason, tc.wantGetCalled, getCalled)
+			}
```
negz, line 430:
```go
			// tc.want.o replaces tc.wantGetCalled
			if diff := cmp.Diff(tc.want.o, tc.args.o,); diff != "" {
				t.Errorf("\n%s\nu.UpdateCriticalAnnotations(...): -want, +got:\n%s", tc.reason, diff)
			}
```

Nit: Typically in this case where we're testing for a mutation of the supplied object we'd use `cmp.Diff` on the whole object as above. A little overkill here, but more extensible if we ever want to test more things about the object we're mutating.

pkg/reconciler/managed/api_test.go
sttts: fixed

negz (approved):
Nice work, thank you!

PR #535, opened 2023-09-06
controller: add TriggeredBy custom watches to engine

sttts:

Added `controller.TriggeredBy` to create controller engine watches based on a different informer cache (the one from outside).

This is helpful e.g. to [trigger the compositor reconciling XRs through changes on `CompositionRevision`](https://github.com/crossplane/crossplane/pull/4582). The compositor has its own informer cache because it is dynamically started for each XR kind, done using the controller engine. This kind of controllers though today can only watch objects in the same cache. This PR adds a way to inject watches from another cache, e.g. the top-level cache watching the `CompositionRevision` kind.

I have:

This has been tested as part of https://github.com/crossplane/crossplane/pull/4582, manually for the demo in that PR, plus I am working on tests for it.

### DIFF
--- a/pkg/controller/engine.go
+++ b/pkg/controller/engine.go
@@ -149,7 +149,10 @@ func (e *Engine) done(name string, err error) {
 
 // Watch an object.
 type Watch struct {
-	kind       client.Object
+	// one of the two:
+	kind         client.Object
+	customSource source.Source
+
 	handler    handler.EventHandler
 	predicates []predicate.Predicate
 }
@@ -158,6 +161,14 @@ type Watch struct {
 // by the supplied EventHandler, and may be filtered by the supplied predicates.
 func For(kind client.Object, h handler.EventHandler, p ...predicate.Predicate) Watch {
 	return Watch{kind: kind, handler: h, predicates: p}
+}
+
+// TriggeredBy returns a custom watch for secondary resources triggering the
+// controller. source.Kind can be used to create a source for a secondary cache.
+// Events will be handled by the supplied EventHandler, and may be filtered by
+// the supplied predicates.
+func TriggeredBy(source source.Source, h handler.EventHandler, p ...predicate.Predicate) Watch {
+	return Watch{customSource: source, handler: h, predicates: p}
 }
 
 // Start the named controller. Each controller is started with its own cache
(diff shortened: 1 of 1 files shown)


negz:
@sttts This LGTM, but can you update the PR description with some context as to why we want this. Any context on how it's been tested would also help.

sttts:
Added context.

pkg/controller/engine.go
```diff
@@ -160,6 +161,13 @@ func For(kind client.Object, h handler.EventHandler, p ...predicate.Predicate) W
 	return Watch{kind: kind, handler: h, predicates: p}
 }
 
+// TriggeredBy returns a custom watch for secondary resources triggering the
+// controller. Events will be handled by the supplied EventHandler, and may be
+// filtered by the supplied predicates.
+func TriggeredBy(cache cache.Cache, kind client.Object, h handler.EventHandler, p ...predicate.Predicate) Watch {
```
ytsarev, line 167: I assume it was tested e2e by https://github.com/crossplane/crossplane/pull/4582/files#diff-52e214a3e1aa3741fcbc3dd3faca4e6ae54e98b177690300591a4d147294a70dR416 ? 

Do we want to extend unit test at https://github.com/crossplane/crossplane-runtime/blob/master/pkg/controller/engine_test.go ?

sttts: The engine is basically untested today. It only checks some errors.

ytsarev: ok, so maybe we should separately track extending unit test coverage? I'll leave it up to you and not hold this one :)

ytsarev (approved):
LGTM

pkg/controller/engine.go
sttts: I tried now for 20min. You mock yourself to death with cache, informer, controller. At the end, you are not sure anymore what you are testing. There is a reason this is only lightly tested 🤷

ytsarev: @sttts got it, thanks a lot for checking! 👍

negz: Yeah, this whole package is kinda just syntactic sugar. 😬

negz (approved):

negz:
Thanks @sttts! The use case makes sense to me now.

sttts:
What is missing to merge this?

pkg/controller/engine.go
```diff
@@ -160,6 +163,14 @@ func For(kind client.Object, h handler.EventHandler, p ...predicate.Predicate) W
 	return Watch{kind: kind, handler: h, predicates: p}
 }
 
+// TriggeredBy returns a custom watch for secondary resources triggering the
+// controller. source.Kind can be used to create a source for a secondary cache.
+// Events will be handled by the supplied EventHandler, and may be filtered by
+// the supplied predicates.
+func TriggeredBy(source source.Source, h handler.EventHandler, p ...predicate.Predicate) Watch {
```
sttts, line 170: @negz see that I slightly changed the signature. It's more flexible now.

Before: `TriggeredBy(someCache, kind, h, p)`
Now: `TriggeredBy(source.Kind(someCache, kind), h, p)`

PR #541, opened 2023-09-11
Enable Namespace as an optional field in TypedReference

### LINKED ISSUE #251: Support Namespaced Managed Resources

diviner524:

Added support for namespace scoped objects in TypedReference, so we can introduce general namespace scoped resource in the resources list of composition. This change is backward compatible with existing implementation and does not require any updates from existing providers.

This also requires some changes in the main Crossplane repo after the runtime is updated. I have created a corresponding draft PR to demonstrate the idea:

https://github.com/crossplane/crossplane/pull/4606

Fixes #251 

I have:

Some manually testing with namespace scoped objects like ConfigMap.

For example, a Composition like below will help to update a namespace scoped ConfigMap with a redisHost URL from the status of certain CompositeResources. This will then enable workloads (like Deployments/StatefulSets) under the same namespace to consume the data from ConfigMap.

```
kind: Composition
...
spec:
...
  resources:
  - name: configmap
    base:
      apiVersion: v1
      kind: ConfigMap
      metadata:
        namespace: "some-namespace"
    patches:
    - type: FromCompositeFieldPath
      fromFieldPath: status.redisHost
      toFieldPath: data.redisHost
```

### DIFF
--- a/apis/common/v1/resource.go
+++ b/apis/common/v1/resource.go
@@ -122,9 +122,9 @@ type Reference struct {
 	Policy *Policy `json:"policy,omitempty"`
 }
 
-// A TypedReference refers to an object by Name, Kind, and APIVersion. It is
-// commonly used to reference cluster-scoped objects or objects where the
-// namespace is already known.
+// A TypedReference refers to an object by Name, Namespace, Kind, and
+// APIVersion. It is commonly used to reference cluster-scoped objects
+// and can also be used for namespace-scoped objects.
 type TypedReference struct {
 	// APIVersion of the referenced object.
 	APIVersion string `json:"apiVersion"`
@@ -134,6 +134,10 @@ type TypedReference struct {
 
 	// Name of the referenced object.
 	Name string `json:"name"`
+
+	// Namespace of the referenced object.
+	// +optional
+	Namespace string `json:"namespace,omitempty"`
 
 	// UID of the referenced object.
 	// +optional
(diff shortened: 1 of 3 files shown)


negz:
I'm not sure this is something we want to support.

Today, Crossplane is designed to allow you to compose Crossplane managed resources (MRs). This is an intentional constraint. The consistent API shape and behaviors of these MRs makes Crossplane easier to reason about, both for end-users and Crossplane maintainers.

I appreciate that this is just a small change. What I am concerned about is the "slippery slope" - that folks will begin to use Crossplane as a composition layer for arbitrary Kubernetes resources and we as maintainers will end up on the hook for trying to make that a good experience.

jbw976:
Here is some historical related discussion and thinking around this topic that is helpful for context and understanding more nuances of the impact of this type of change: https://github.com/crossplane/crossplane-runtime/pull/541

negz:
I think the spirit of this PR is / will be addressed by Crossplane v2 (see [proposal](https://github.com/crossplane/crossplane/pull/6255)) which'll allow composing namespaced resources.

PR #548, opened 2023-09-15
fix(package parser): only drop fully commented files

phisco:

Fixes https://github.com/crossplane/crossplane/issues/4631

I have:

Added unit tests and updated existing tests.

### DIFF
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -130,19 +130,16 @@ func (p *PackageParser) Parse(_ context.Context, reader io.ReadCloser) (*Package
 	return pkg, nil
 }
 
-// cleanYAML cleans up YAML by removing empty and commented out lines which
+// cleanYAML cleans up YAML only drops fully commented inputs which would
 // cause issues with decoding.
 func cleanYAML(y []byte) []byte {
 	lines := []string{}
 	empty := true
 	for _, line := range strings.Split(string(y), "\n") {
 		trimmed := strings.TrimSpace(line)
-		if trimmed == "" || strings.HasPrefix(trimmed, "#") {
-			continue
-		}
 		// We don't want to return an empty document with only separators that
 		// have nothing in-between.
-		if empty && trimmed != "---" && trimmed != "..." {
+		if empty && trimmed != "" && trimmed != "---" && trimmed != "..." && !strings.HasPrefix(trimmed, "#") {
 			empty = false
 		}
 		lines = append(lines, line)
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -22,6 +22,7 @@ import (
 	"context"
 	"io"
 	"strings"
+	"unicode"
 
 	"github.com/spf13/afero"
 	corev1 "k8s.io/api/core/v1"
@@ -107,8 +108,7 @@ func (p *PackageParser) Parse(_ context.Context, reader io.ReadCloser) (*Package
 		if errors.Is(err, io.EOF) {
 			break
 		}
-		content = cleanYAML(content)
-		if len(content) == 0 {
+		if isEmptyYAML(content) {
 			continue
 		}
 		m, _, err := dm.Decode(content, nil, nil)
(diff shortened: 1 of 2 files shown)


pkg/parser/parser.go
```diff
@@ -130,19 +130,16 @@ func (p *PackageParser) Parse(_ context.Context, reader io.ReadCloser) (*Package
 	return pkg, nil
 }
 
-// cleanYAML cleans up YAML by removing empty and commented out lines which
+// cleanYAML cleans up YAML only drops fully commented inputs which would
```
jbw976, line 133: This is reading a little off now - is it correct to say the following? I'm trying to be more complete about what this function is doing and returning - feel free to disregard the completely aspect if its wrong or you don't think its needed, but the sentence as is is missing some flow 🤓 

```suggestion
// cleanYAML cleans up the input YAML by removing all leading and trailing white
// space from each line. If the input yaml is completely commented, then nil
// will be returned to avoid issues with decoding.
```

pkg/parser/parser_test.go
```diff
@@ -210,3 +215,84 @@ func TestParser(t *testing.T) {
...
+		reason string
+		args   args
+		want   want
+	}{
+		"Empty": {
+			reason: "Should return nil on empty input",
+			args:   args{in: []byte("")},
+			want:   want{out: nil},
+		},
+		"CommentedOut": {
```
jbw976, line 236: good test cases! one thing that is nagging my mind here is if there's a scenario @muvaf was handling with the original function that really needed all commented lines removed.  It seems like the conclusion here is that that was too aggressive, and its OK to leave in commented lines if the full block isn't completely commented.

You feel good about that too? There is not a scenario where commented lines should actually be removed? 🤔

jbw976 (approved):
Thanks for a very quick turnaroud on this @phisco!!! 💪  - a couple small comments for your consideration, but I've already added an approval ✅

pkg/parser/parser_test.go
phisco: From the description of the original PR my understanding was that only fully commented files were an issue, so being less aggressive should be fine. The current behavior in released versions of Crossplane is to not drop any line at all, so I guess there should be no issue only dropping fully commented files.

pkg/parser/parser.go
```diff
@@ -130,19 +130,16 @@ func (p *PackageParser) Parse(_ context.Context, reader io.ReadCloser) (*Package
...
 	empty := true
 	for _, line := range strings.Split(string(y), "\n") {
 		trimmed := strings.TrimSpace(line)
-		if trimmed == "" || strings.HasPrefix(trimmed, "#") {
-			continue
-		}
 		// We don't want to return an empty document with only separators that
 		// have nothing in-between.
-		if empty && trimmed != "---" && trimmed != "..." {
+		if empty && trimmed != "" && trimmed != "---" && trimmed != "..." && !strings.HasPrefix(trimmed, "#") {
```
sttts, line 142: why do we do anything like that on the text level? Can't we just decode and see that a null object (no keys) comes out? Would be much safer than any complicated risky logic here.

pkg/parser/parser_test.go
muvaf: @phisco is right, my motivation was to have the parser skip fully-commented documents so we don't have errors in the upstream marshal/unmarshal functions. It was essentially extending the definition of empty document from having no characters at all to having all lines commented out. Not sure why I ended up removing any lines, maybe while [trying to reduce](https://github.com/crossplane/crossplane-runtime/pull/504#discussion_r1287250192) the number of loops?

I think this function can just be a check like `isEmpty` and not make any changes to the document - the problem isn't about having comments but more about having the whole document not have any useful data hence failing the functions trying to cast it onto a struct.

phisco: Thanks @muvaf for the feedback, I'll rename the function to IsEmpty and make sure it doesn't have any side effect 👍

pkg/parser/parser.go
muvaf: We use a high level `Decode` function from upstream which does not have a specific error type for empty documents as far as I remember (@phisco might know better now) and checking only for `null` would not be specific enough due to number of reasons it can be `null`.

turkenh (approved):
Thanks @phisco 🙌

PR #549, opened 2023-09-18
GMP to beta

lsviben:

As part of [promoting GMP to beta](https://github.com/crossplane/crossplane/issues/4126) change the flag into a BETA flag

I have:

### DIFF
--- a/apis/common/v1/resource.go
+++ b/apis/common/v1/resource.go
@@ -197,9 +197,8 @@ type ResourceSpec struct {
 	// +kubebuilder:default={"name": "default"}
 	ProviderConfigReference *Reference `json:"providerConfigRef,omitempty"`
 
-	// THIS IS AN ALPHA FIELD. Do not use it in production. It is not honored
-	// unless the relevant Crossplane feature flag is enabled, and may be
-	// changed or removed without notice.
+	// THIS IS A BETA FIELD. It is on by default but can be opted out
+	// through a Crossplane feature flag.
 	// ManagementPolicies specify the array of actions Crossplane is allowed to
 	// take on the managed and external resources.
 	// This field is planned to replace the DeletionPolicy field in a future
--- a/pkg/reconciler/managed/reconciler.go
+++ b/pkg/reconciler/managed/reconciler.go
@@ -660,7 +660,7 @@ func WithRecorder(er event.Recorder) ReconcilerOption {
 // WithManagementPolicies enables support for management policies.
 func WithManagementPolicies() ReconcilerOption {
 	return func(r *Reconciler) {
-		r.features.Enable(feature.EnableAlphaManagementPolicies)
+		r.features.Enable(feature.EnableBetaManagementPolicies)
 	}
 }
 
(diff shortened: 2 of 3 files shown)


apis/common/v1/resource.go
```diff
@@ -197,9 +197,8 @@ type ResourceSpec struct {
 	// +kubebuilder:default={"name": "default"}
 	ProviderConfigReference *Reference `json:"providerConfigRef,omitempty"`
 
-	// THIS IS AN ALPHA FIELD. Do not use it in production. It is not honored
-	// unless the relevant Crossplane feature flag is enabled, and may be
-	// changed or removed without notice.
+	// THIS IS AN BETA FIELD. It is on by default but can be opted out
```
turkenh, line 200:
```suggestion
	// THIS IS A BETA FIELD. It is on by default but can be opted out
```

pkg/feature/feature.go
```diff
@@ -40,6 +40,13 @@ func (fs *Flags) Enable(f Flag) {
 	fs.m.Unlock()
 }
 
+// Disable a feature flag.
+func (fs *Flags) Disable(f Flag) {
```
turkenh, line 44: We won't need a disable method at this level. If we don't enable it will be disabled and we can do the defaulting at the flag level. See Composition functions handling beta: https://github.com/crossplane/crossplane/blob/75e39029366de2946c9c54f318f7822a6cefcc40/cmd/crossplane/core/core.go#L113

pkg/reconciler/managed/reconciler.go
```diff
@@ -657,10 +657,10 @@ func WithRecorder(er event.Recorder) ReconcilerOption {
 	}
 }
 
-// WithManagementPolicies enables support for management policies.
-func WithManagementPolicies() ReconcilerOption {
+// WithoutManagementPolicies disables support for management policies beta feature.
+func WithoutManagementPolicies() ReconcilerOption {
```
turkenh, line 661: See the above comment. I don't think we would need any change other than making the value of [this flag](https://github.com/upbound/provider-aws/blob/6bbeee5ef06a4c83eda9f601678b2f425ba082c7/hack/main.go.tmpl#L50) `true` by default.

pkg/feature/feature.go
lsviben: I also checked this at the beginning, but in the end I decided to do it this way because controller-runtime is a library. So if we want to have the flag set to true, we need to  [activate it by default](https://github.com/upbound/provider-aws/blob/6bbeee5ef06a4c83eda9f601678b2f425ba082c7/hack/main.go.tmpl#L50) in every provider that currently exists, and every future provider. While for composition functions we just needed to do it in one place.

Thats why I thought its better to have it in the library by default, and then let providers opt-out. But now when I look at it again, it can be confusing that a feature flag is set in the reconciler. Ill switch to what you suggest and just change the Alpha -> Beta thing here

turkenh: I see. 

Considering that the flag already exists in the existing providers controlling the behavior (and we will anyway need a change on every provider), I suggest continuing to use it and also being consistent with Core Crossplane. Updating our template repositories accordingly should be good enough for the new ones.

turkenh: Once we graduate to v1, where disabling the feature wouldn't be possible, we can do a clean-up here and always on by ignoring whatever is passed from the providers.

turkenh (approved):

PR #550, opened 2023-09-18
resource: use Patch in APIFinalizer

sttts:

Calling `Update` on a non-unstructured object will potentially lead to data-loss because it is not guaranteed that the object is a complete representation of all the fields stored in the apiserver, e.g. because of version skew between client and Crossplane.

I have:

- [ ] is that type cast safe?
- [ ] do we have enough test coverage to confirm the safety of this change?
- [ ] maybe run proof PR in crossplane

Via Crossplane test suite, aka unit tests and e2e tests.


sttts:
Actually, this can lose changes that were made before calling the finalizer code, i.e. there is no way around comparing to a fresh object :-/

PR #552, opened 2023-09-20
SetClaimReference must pass to underlying object just claimRef allowed fields

pedjak:

`spec.claimRef` schema is just subset of `corev1.ObjectReference`, and hence `SetClaimReference` might get a reference that have more fields set, e.g. UID. The fields that do not exist in claimRef schema must be not set on the underlying object, otherwise K8s API server complains about non-existing field when client sends a patch request.
    
* Introduced `claim.Reference` type that corresponds to `spec.claimRef` schema.
* `composite.SetClaimReference` signature changed to accept an instance of `claim.Reference`
* appropriate tests updated/new tests added

Discovered while working on https://github.com/crossplane/crossplane/issues/4047

I have:

Adjusted unit test to catch regression.

### DIFF
--- a/pkg/resource/unstructured/composite/composite.go
+++ b/pkg/resource/unstructured/composite/composite.go
@@ -25,6 +25,7 @@ import (
 
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 	"github.com/crossplane/crossplane-runtime/pkg/fieldpath"
+	"github.com/crossplane/crossplane-runtime/pkg/resource/unstructured/claim"
 )
 
 // An Option modifies an unstructured composite resource.
@@ -137,8 +138,8 @@ func (c *Unstructured) GetCompositionUpdatePolicy() *xpv1.UpdatePolicy {
 }
 
 // GetClaimReference of this Composite resource.
-func (c *Unstructured) GetClaimReference() *corev1.ObjectReference {
-	out := &corev1.ObjectReference{}
+func (c *Unstructured) GetClaimReference() *claim.Reference {
+	out := &claim.Reference{}
 	if err := fieldpath.Pave(c.Object).GetValueInto("spec.claimRef", out); err != nil {
 		return nil
 	}
@@ -146,7 +147,7 @@ func (c *Unstructured) GetClaimReference() *corev1.ObjectReference {
 }
 
 // SetClaimReference of this Composite resource.
-func (c *Unstructured) SetClaimReference(ref *corev1.ObjectReference) {
+func (c *Unstructured) SetClaimReference(ref *claim.Reference) {
 	_ = fieldpath.Pave(c.Object).SetValue("spec.claimRef", ref)
 }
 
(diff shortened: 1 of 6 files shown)


negz:
@pedjak Do you think it would be possible to switch to using a type without the extraneous fields? That seems ideal to me, since it would also update our claim OpenAPI schemas to better communicate that we don't actually pay any attention to all the extraneous fields (e.g. UID). In some places we've started defining our own types instead of ObjectReference, per https://github.com/crossplane/crossplane-runtime/issues/49.

pkg/resource/unstructured/composite/composite.go
```diff
@@ -147,7 +147,17 @@ func (c *Unstructured) GetClaimReference() *corev1.ObjectReference {
...
+	// the passed ref might have more fields set, e.g. UID
+	// and they need to be removed, otherwise K8s API server complains about non-existing field
+	// when client sends patch request
+	claimRef := &corev1.ObjectReference{
+		APIVersion: ref.APIVersion,
+		Kind:       ref.Kind,
+		Namespace:  ref.Namespace,
+		Name:       ref.Name,
+	}
+	_ = fieldpath.Pave(c.Object).SetValue("spec.claimRef", claimRef)
```
sttts, line 160: lgtm.

Just curious, why do we use a type if it is not supported?

pedjak:
> Do you think it would be possible to switch to using a type without the extraneous fields? That seems ideal to me, since it would also update our claim OpenAPI schemas to better communicate that we don't actually pay any attention to all the extraneous fields (e.g. UID). In some places we've started defining our own types instead of ObjectReference, per 

Not sure if I got it right: would you like to change the signature of `SetClaimReference`? That would be a breaking API change, not sure what would be the blast radius? 

IMHO, it is quite alright to pass `ObjectReference` and the method take out of it what is needed to craft claim reference valid according to the schema.

pkg/resource/unstructured/composite/composite.go
pedjak: I also wonder if this method needs to live in crossplane-runtime at all, because it is only consumed in crossplane?

sttts: That's a different question. What @negz and I mean is that `corev1.ObjectReference` is the wrong type. We should have a `ClaimReference` type instead that only has the fields that are allowed to be set.

sttts: can you change this, i.e. introduce that type?

sttts:
> That would be a breaking API change, not sure what would be the blast radius?

Blast radius is minimal:

1. https://grep.app/search?q=SetClaimReference
2. the change is safe, i.e. compilation breaks. So ppl will notice.

pkg/resource/unstructured/composite/composite.go
negz: Does anything actually write (or read) the UID field in a claimRef today? If no, I feel fine making the change since while technically breaking at the API level it would functionally be a no-op.

pedjak:
> Blast radius is minimal:

@sttts @negz I have updated PR, ptal.

turkenh (approved):

PR #553, opened 2023-09-20
introduce Lintable interace 

lsviben:

Introduces a Lintable interface as a common API for lintable packages.
Resolves one of the [TODO's](https://github.com/upbound/up/blob/main/internal/xpkg/parser/linter/linter.go#L33-L40) in `up xpkg` that we are planning to move to Crossplane, so 
that we avoid code duplication for the migration.

I have:

Just ran the unit tests.

### DIFF
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -33,6 +33,12 @@ import (
 	"github.com/crossplane/crossplane-runtime/pkg/errors"
 )
 
+// Lintable defines the common API for lintable packages.
+type Lintable interface {
+	GetMeta() []runtime.Object
+	GetObjects() []runtime.Object
+}
+
 // AnnotatedReadCloser is a wrapper around io.ReadCloser that allows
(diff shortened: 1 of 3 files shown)


phisco (approved):

pkg/parser/parser.go
```diff
@@ -33,6 +33,12 @@ import (
 	"github.com/crossplane/crossplane-runtime/pkg/errors"
 )
 
+// Lintable defines the common API for lintable packages.
+type Lintable interface {
+	GetMeta() []runtime.Object
```
pedjak, line 38: what is the semantic difference between these two methods, since they return the same types? Can you please add some doc?

pedjak (changes requested):

pkg/parser/parser.go
lsviben: Added some comments, hope it makes it more clear!

pedjak (approved):

PR #556, opened 2023-09-21
errors: add SilentlyRequeueOnConflicts

sttts:

Example:
```golang
func Reconcile(...) (reconcile.Result, error) {
    ...

    return errors.SilentlyRequeueOnConflict(reconcile.Result{}, r.hcClient.Patch(ctx, obj, client.MergeFrom(orig)))
}
```

If the returned error is a (possibly wrapped) Kube conflict error, it is dropped (replaced with `nil`) and the result is replaced with an immediate requeue.

This way, the conflict errors, which are part of the Kube's optimistic concurrency protocol and not actual errors, do not show up as ugly backtrace in the logs.

Alternative way to use:
```golang
errors.WithSilentRequeueOnConflict(NewReconciler())
```

I have:

Unit tested.

### DIFF
--- a/pkg/errors/reconcile_test.go
+++ b/pkg/errors/reconcile_test.go
@@ -1,5 +1,18 @@
-// Copyright 2023 Upbound Inc.
-// All rights reserved
+/*
+Copyright 2023 The Crossplane Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


pkg/errors/reconcile_test.go
```diff
@@ -0,0 +1,86 @@
+// Copyright 2023 Upbound Inc.
+// All rights reserved
```
lsviben, line 2: Should be the Crossplane copyright stuff here.

lsviben (approved):
LGTM, thanks @sttts. Added just a comment about license boilerplate that should be fixed before merging

pkg/errors/reconcile_test.go
sttts: indeed

sttts: fixed

negz (approved):

PR #618, opened 2023-12-04
Treat all maps & slices of zero-length equal whether they are nil or not in managed.APISimpleReferenceResolver

### LINKED ISSUE #617: Treat all maps & slices of zero-length equal whether they are nil or not in `managed.APISimpleReferenceResolver`
### What happened?

[Here](https://github.com/crossplane/crossplane-runtime/blob/fa372205d05d18104f3e67a6837ad449d4f19bce/pkg/reconciler/managed/api.go#L149), the `managed.APISimpleReferenceResolver` does not consider nil maps & slices as being equal to empty maps or slices. Because in the context of reference resolution, the target referencer fields being set (e.g., `Ref` fields, if a label selector is specified, and the reference target fields) are semantically equal if they are nil or if they are empty, we had better treat them as equals when doing a comparison. This will prevent unnecessary apply operations when the comparison finds a nil map or slice not equal to an empty one. Using the `cmpopts.EquateEmpty()` should improve this.

### How can we reproduce it?

This behavior can be invoked, for example, by specifying a label selector for a cross-resource reference field when the resolution is not done yet (e.g., the referenced MRs do not exist or not ready yet).

Crossplane version: 1.14.3

ulucinar:

Fixes #617 

This PR passes the `cmpopts.EquateEmpty()` option when the `managed.APISimpleReferenceResolver` is checking whether any reference fields (`Ref` fields if a label selector is specified and the reference target fields) have been set by the downstream`ResolveReferences` call, so that all maps & slices of zero-length will be considered as equal whether they are nil or empty.

I have:

Tested with a custom build of `upbound/provider-aws`.

### DIFF
--- a/pkg/reconciler/managed/api.go
+++ b/pkg/reconciler/managed/api.go
@@ -146,7 +146,7 @@ func (a *APISimpleReferenceResolver) ResolveReferences(ctx context.Context, mg r
 		return errors.Wrap(err, errResolveReferences)
 	}
 
-	if cmp.Equal(existing, mg) {
+	if cmp.Equal(existing, mg, cmpopts.EquateEmpty()) {
 		// The resource didn't change during reference resolution.
 		return nil
 	}


turkenh (approved):
Thank you @ulucinar 🙌

PR #627, opened 2023-12-12
Add a nil case to the expandWildcards function

sergenyalcin:

This PR adds a nil case to the `expandWildcards` function to handle the case that the value of the segment is nil.

The issue was observed while testing the GCP `certificatemanager.Certificate` resource. While trying to get the connection details of this resource, we try to read the `self_managed[*].certificate_pem` attribute of this resource. If the `self_managed` object is nil, then we observe an error: 

`cannot get connection details: cannot get connection
        details: cannot expand wildcards: cannot expand wildcards for segments: "self_managed[*].certificate_pem":
        "self_managed": unexpected wildcard usage'`
        
If the value of this field is nil, then we do not want to return an error because there is nothing to read in the sensitive attribute. In this case, we want to handle this situation in Upjet as a specific case other than the different cases. So, by returning a specific error from the `nil` case, we can handle this in Upjet by ignoring this error type.

I observed a specific error type, `errNotFound` and used this for the `nil` case.

I have:

This was tested in the Upjet-based GCP provider. Resource: `certificatemanager.Certificate`

### DIFF
--- a/pkg/fieldpath/paved.go
+++ b/pkg/fieldpath/paved.go
@@ -205,7 +205,7 @@ func expandWildcards(data any, segments Segments) ([]Segments, error) { //nolint
 					res = append(res, r...)
 				}
 			case nil:
-				return nil, errNotFound{errors.Errorf("%s: value of the field is nil", segments[:i])}
+				return nil, errNotFound{errors.Errorf("wildcard field %q is not found in the path", segments[:i])}
 			default:
 				return nil, errors.Errorf("%q: unexpected wildcard usage", segments[:i])
 			}
--- a/pkg/fieldpath/paved.go
+++ b/pkg/fieldpath/paved.go
@@ -204,6 +204,8 @@ func expandWildcards(data any, segments Segments) ([]Segments, error) { //nolint
 					}
 					res = append(res, r...)
 				}
+			case nil:
+				return nil, errNotFound{errors.Errorf("%s: value of the field is nil", segments[:i])}
 			default:
 				return nil, errors.Errorf("%q: unexpected wildcard usage", segments[:i])
 			}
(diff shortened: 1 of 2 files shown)


pkg/fieldpath/paved.go
```diff
@@ -204,6 +204,8 @@ func expandWildcards(data any, segments Segments) ([]Segments, error) { //nolint
 					}
 					res = append(res, r...)
 				}
+			case nil:
+				return nil, errNotFound{errors.Errorf("%s: value of the field is nil", segments[:i])}
```
ulucinar, line 208: nit:
```suggestion
				return nil, errNotFound{errors.Errorf("value of the field %q is nil", segments[:i])}
```

ulucinar: Here, we differentiate between the cases where an unexpected type is encountered versus the specified path is not found and we hit the new added `case nil` here.

pkg/fieldpath/paved.go
```diff
@@ -204,6 +204,8 @@ func expandWildcards(data any, segments Segments) ([]Segments, error) { //nolint
 					}
 					res = append(res, r...)
 				}
+			case nil:
```
ulucinar, line 207: Let's add a new test case here similar to the following:
```go
			path:   "spec.containers[*].a",
			data:   []byte(`{"spec":{"containers":null}}`),
```

sergenyalcin: Added. Thanks @ulucinar

ulucinar (approved):
Thanks @sergenyalcin, lgtm.

turkenh (approved):
Thanks @sergenyalcin 🙌 Returning `errNotFound` for that case makes sense to me.

PR #633, opened 2023-12-27
Add ability to expose resource reconciliation progress

pedjak:

* Optional `status.observedGeneration` fields has been added to claim/composite/composed status, showing the latest metadata.generation which resulted in either a ready state, or stalled due to error it can not recover from without human intervention.
* Optional `status.conditions[x].observedGeneration` represents the .metadata.generation that the condition was set based upon.

Conditions are now aligned with [SIG Architecture API convention](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties).

Fixes https://github.com/crossplane/crossplane/issues/4695

Part of https://github.com/crossplane/crossplane/issues/4655

I have:


apis/common/v1/condition.go
```diff
@@ -107,6 +107,21 @@ type ConditionedStatus struct {
...
+
+	// ObservedLabels are the most recent resource metadata.labels
+	// observed by the reconciler
+	// +optional
+	ObserverLabels map[string]string `json:"observedLabels,omitempty"`
+
+	// ObservedLabels are the most recent resource metadata.annotations
+	// observed by the reconciler
+	// +optional
+	ObserverAnnotations map[string]string `json:"observedAnnotations,omitempty"`
```
negz, line 124:
```suggestion
	// ObservedLabels are the most recent resource metadata.labels
	// observed by the reconciler
	// +optional
	ObservedLabels map[string]string `json:"observedLabels,omitempty"`

	// ObservedLabels are the most recent resource metadata.annotations
	// observed by the reconciler
	// +optional
	ObservedAnnotations map[string]string `json:"observedAnnotations,omitempty"`
```

Can you please create a new struct (something like `ObservedStatus`) for these fields, instead of adding them to `ConditionedStatus`. The idea of `ConditionedStatus` is that you can embed it in a status struct to support status conditions. The new struct should be in a different file (not condition.go).

Is there precedent for other tools having observedLabels and observedAnnotations as well as observedGeneration? I'm supportive of observedGeneration but not convinced that it's worth two additional fields for labels and annotations.

pkg/resource/interfaces.go
```diff
@@ -176,6 +176,18 @@ type ConnectionDetailsPublishedTimer interface {
...
+type ReconciliationObserver interface {
+	SetObservedGeneration(generation int64)
+	GetObservedGeneration() *int64
+
+	SetObservedLabels(labels map[string]string)
+	GetObservedLabels() map[string]string
+
+	SetObservedAnnotations(annotations map[string]string)
+	GetObservedAnnotations() map[string]string
+}
```
negz, line 189: Do you expect that managed resources would also satisfy this interface, eventually?

apis/common/v1/condition.go
pedjak: @negz great that you spot the typo!

> The new struct should be in a different file (not condition.go).

Done, the struct lives now in `apis/common/v1/observation.go`

> Is there precedent for other tools having observedLabels and observedAnnotations as well as observedGeneration? I'm supportive of observedGeneration but not convinced that it's worth two additional fields for labels and annotations.

Value in `metadata.generation` get increased only when resource `spec` changes. Given that our controllers react also on label/annotation changes and often use/propagate them during the reconciliation, writing down just the observed generation might not be enough, i.e. an observer would not be able to detect if/when a label/annotation change got processed.

pedjak: > Do we know of concrete situations (ideally issues) that recording observed labels and annotations would have helped with? If not, I would suggest we omit them until there's evidence they're needed.

The PR adds an ability to record the observed labels/annotations and the both fields are optional. IMHO a controller should set them if they are used anyhow to influence the reconcilation logic.

In a Crossplane claim, its annotations and labels are copied down to its counterpart composite as a part of reconciliation.

If we do not record the observed claim annotations/lables, then an observer is not able to detect if a reconciliation has happened:
* A claim got created and became ready
* The claim annotation got added/removed/updated
* Given that `metadata.generation` did not change, `status.observedGeneration` does not change either after the reconcile

I have found an old issue about asking to increase `metadata.generation` on annotation/label changes:
https://github.com/kubernetes/kubernetes/issues/67428 and very interesting followup discussion.

@sttts what are your thought on this? Should we stop propagating annotations/labels from claim to composite? Should it be acceptable if claim reconciliation on pure annotation/label changes is not detectable by an observer?

negz: > If we do not record the observed claim annotations/lables, then an observer is not able to detect if a reconciliation has happened.

I get that, but I'm challenging whether this has ever been a real problem for someone. Should we wait to see if this is hurting people before we fix it?

negz: > Should we stop propagating annotations/labels from claim to composite? 

I don't think we can - this would be a breaking behaviour change.

sttts:
Which labels and annotations do we have in mind here? In the Kubernetes object model neither should really matter, the spec does.

How can we make this forward? If the observedGeneration is a no-brainer, can we move that forward first?

apis/common/v1/condition.go
sttts: > The PR adds an ability to record the observed labels/annotations and the both fields are optional. IMHO a controller should set them if they are used anyhow to influence the reconcilation logic.
> In a Crossplane claim, its annotations and labels are copied down to its counterpart composite as a part of reconciliation.
> If we do not record the observed claim annotations/lables, then an observer is not able to detect if a reconciliation has happened

I tend to agree with these statements.

That having said, IMO passing down labels/annotations from claim to xrd and similar is wrong, especially for annotations. But as they are passed down today, it sounds logical that the controller records them as part of the observed state.

negz:
> If the observedGeneration is a no-brainer, can we move that forward first?

Yes, I think we should do that.

The labels/annotations seems trickier, so I want to hold on that for now:

* There's a school of thought that we shouldn't propagate them at all.
* `observedGeneration` is an established pattern in the Kobe world, but I don't think `observedLabels` etc is.
* `observedGeneration` copies only the `generation` int, labels etc we'd need to copy the whole object.

apis/common/v1/resource.go
```diff
@@ -226,6 +226,7 @@ type ResourceSpec struct {
 // ResourceStatus represents the observed state of a managed resource.
 type ResourceStatus struct {
 	ConditionedStatus `json:",inline"`
+	ObservedStatus    `json:",inline"`
```
negz, line 229: Should XRs and claims have `observedGeneration` too? I think that would be ideal.

Unfortunately they work a little differently, we'd need to update `internal/xcrd` in c/c.

negz: Edit: Nevermind, this PR already indicates that that's the goal.

pedjak:
> Yes, I think we should do that.

@negz @sttts I have updated PR, aligned also with [SIG Architecture API convetion](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties).

This opens up the possibility to achieve compatibility with [kstatus specification](https://github.com/kubernetes-sigs/cli-utils/tree/master/pkg/kstatus#statuses) at some point.

sttts:
@negz can you take another look?

pkg/resource/unstructured/claim/claim_test.go
```diff
@@ -368,3 +368,24 @@ func TestConnectionDetailsLastPublishedTime(t *testing.T) {
...
+	}
+}
+
+func TestObservedGenerationNotFound(t *testing.T) {
+	u := &Unstructured{unstructured.Unstructured{Object: map[string]any{}}}
+
+	if g := u.GetObservedGeneration(); g != 0 {
+		t.Errorf("u.GetObservedGeneration(): expected nil, but got %v", g)
+	}
+}
```
negz, line 391: These tests are written in a different style from every other test in the file. Despite the fact that they're mostly tables with a single case, I'd prefer to match the existing practices.

apis/common/v1/condition.go
```diff
@@ -141,6 +157,9 @@ func (s *ConditionedStatus) SetConditions(c ...Condition) {
 
 			if existing.Equal(new) {
```
negz, line 158: GitHub won't let me add a comment to the `existing.Equal` docstring, but it should explicitly call out that it doesn't consider `ObservedGeneration` when assessing equality. It already calls out that it ignores LastTransitionTime.

pkg/resource/unstructured/claim/claim.go
```diff
@@ -223,7 +223,7 @@ func (c *Unstructured) SetPublishConnectionDetailsTo(ref *xpv1.PublishConnection
 
 // GetCondition of this composite resource claim.
 func (c *Unstructured) GetCondition(ct xpv1.ConditionType) xpv1.Condition {
-	conditioned := xpv1.ConditionedStatus{}
+	conditioned := xpv1.ResourceStatus{}
```
negz, line 226: Why change this to ResourceStatus?

pkg/resource/unstructured/claim/claim.go
```diff
@@ -233,7 +233,7 @@ func (c *Unstructured) GetCondition(ct xpv1.ConditionType) xpv1.Condition {
 
 // SetConditions of this composite resource claim.
 func (c *Unstructured) SetConditions(conditions ...xpv1.Condition) {
-	conditioned := xpv1.ConditionedStatus{}
+	conditioned := xpv1.ResourceStatus{}
```
negz, line 236: Why change this to ResourceStatus?

pkg/resource/unstructured/composed/composed.go
```diff
@@ -72,7 +72,7 @@ func (cr *Unstructured) GetUnstructured() *unstructured.Unstructured {
 
 // GetCondition of this Composed resource.
 func (cr *Unstructured) GetCondition(ct xpv1.ConditionType) xpv1.Condition {
-	conditioned := xpv1.ConditionedStatus{}
+	conditioned := xpv1.ResourceStatus{}
```
negz, line 75: Why change this to ResourceStatus?

pedjak:
I have addressed your comments @negz ptal.

negz (approved):
Thank you!

PR #641, opened 2024-01-09
controller/engine: sync informers on controller start

sttts:

Fixes crossplane/crossplane#5151.

Before this, there was no wait for informers when starting the XR controller. This PR is supposed to ensure that all informers are in sync before XRs are reconciled.

I have:

Mostly by reading the code carefully. Waiting for feedback from https://github.com/crossplane/crossplane/issues/5151.

### DIFF
--- a/pkg/controller/engine.go
+++ b/pkg/controller/engine.go
@@ -271,6 +271,10 @@ func (c *namedController) Start(ctx context.Context) error {
 	}()
 	go func() {
 		<-c.e.mgr.Elected()
+		if synced := c.ca.WaitForCacheSync(ctx); !synced {
+			c.e.done(c.name, errors.New(errCrashCache))
+			return
+		}
 		c.e.done(c.name, errors.Wrap(c.ctrl.Start(ctx), errCrashController))
 	}()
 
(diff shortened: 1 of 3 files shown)


pkg/controller/engine.go
```diff
@@ -271,6 +271,10 @@ func (c *namedController) Start(ctx context.Context) error {
 	}()
 	go func() {
 		<-c.e.mgr.Elected()
+		if synced := c.ca.WaitForCacheSync(ctx); !synced {
+			c.e.done(c.name, errors.New(errCrashCache))
+			return
+		}
```
negz, line 277: I think this is already happening inside of controller-runtime for the non-realtime case. I find their code hard to follow due to all the returning of interfaces, but AFAICT:

1. Our `e.Create` calls `ctrl.Watch`, with `source.Kind(ca, ...)` as an argument. This adds the source to the controller's internal `startWatches` slice of watches to start when `ctrl.Start` is called.
2. When `ctrl.Start` is called it iterates over `startWatches`, calling each source's `Start` method.
3. The source should be an `internal.Kind`, and its `Start` method calls its cache's `WaitForCacheSync` method.

That's not guaranteed to be the case when using `TriggeredBy` though.

Assuming it doesn't hurt to call `WaitForCacheSync` multiple times I think it's a good idea to add this regardless.

negz (approved):

PR #651, opened 2024-01-27
reconciler/managed: make more resilient to error conditions

sttts:

Two cases where the manager reconciler bricked an object forever:

1. when `external.Create` returned a conflict error. This was an oversight in https://github.com/crossplane/crossplane-runtime/pull/540 when introducing less noisy conflict handling.
2. on non-API errors when the critical failed/succeeded annotations were applied. Think of network issues. These are not API errors and hence they stopped the retry loop.

There is still a risk that an object becomes bricked, e.g.:
- when the ctx is cancelled after some timeout (30s or so per reconcile?). I.e. if the error persists for long enough.
- when the controller crashes before apply the critical annotations.

I have:

Careful code reading and review. This is hard to test other than under real-life condition under heavy load.

### DIFF
--- a/pkg/reconciler/managed/api.go
+++ b/pkg/reconciler/managed/api.go
@@ -212,7 +212,9 @@ func NewRetryingCriticalAnnotationUpdater(c client.Client) *RetryingCriticalAnno
 // case of a conflict error.
 func (u *RetryingCriticalAnnotationUpdater) UpdateCriticalAnnotations(ctx context.Context, o client.Object) error {
 	a := o.GetAnnotations()
-	err := retry.OnError(retry.DefaultRetry, resource.IsAPIError, func() error {
+	err := retry.OnError(retry.DefaultRetry, func(err error) bool {
+		return !errors.Is(err, context.Canceled)
+	}, func() error {
 		err := u.client.Update(ctx, o)
 		if kerrors.IsConflict(err) {
 			if getErr := u.client.Get(ctx, client.ObjectKeyFromObject(o), o); getErr != nil {
--- a/pkg/reconciler/managed/reconciler.go
+++ b/pkg/reconciler/managed/reconciler.go
@@ -1051,10 +1051,9 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (resu
 			// issue we'll be requeued implicitly when we update our status with
 			// the new error condition. If not, we requeue explicitly, which will trigger backoff.
 			log.Debug("Cannot create external resource", "error", err)
-			if kerrors.IsConflict(err) {
-				return reconcile.Result{Requeue: true}, nil
+			if !kerrors.IsConflict(err) {
+				record.Event(managed, event.Warning(reasonCannotCreate, err))
 			}
-			record.Event(managed, event.Warning(reasonCannotCreate, err))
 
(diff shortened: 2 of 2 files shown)


pkg/reconciler/managed/api.go
```diff
@@ -212,7 +212,9 @@ func NewRetryingCriticalAnnotationUpdater(c client.Client) *RetryingCriticalAnno
 // case of a conflict error.
 func (u *RetryingCriticalAnnotationUpdater) UpdateCriticalAnnotations(ctx context.Context, o client.Object) error {
 	a := o.GetAnnotations()
-	err := retry.OnError(retry.DefaultRetry, resource.IsAPIError, func() error {
+	err := retry.OnError(retry.DefaultRetry, func(err error) bool {
+		return !errors.Is(err, context.Canceled)
```
luxas, line 216: This seems common enough that it could be encapsulated somewhere; to our own RetryOnError func?
Or when would we want to retry although the context was cancelled?

luxas (commented):
Change seems sensible.

However, when I read the code, I noticed that we don't propagate the possible error of UpdateCriticalAnnotations to the managed resource's status, we only emit an event

```go
			// We handle annotations specially here because it's
			// critical that they are persisted to the API server.
			// If we don't add the external-create-failed annotation
			// the reconciler will refuse to proceed, because it
			// won't know whether or not it created an external
			// resource.
			meta.SetExternalCreateFailed(managed, time.Now())
			if err := r.managed.UpdateCriticalAnnotations(ctx, managed); err != nil {
				log.Debug(errUpdateManagedAnnotations, "error", err)
				record.Event(managed, event.Warning(reasonCannotUpdateManaged, errors.Wrap(err, errUpdateManagedAnnotations)))

				// We only log and emit an event here rather
				// than setting a status condition and returning
				// early because presumably it's more useful to
				// set our status condition to the reason the
				// create failed.
			}

			managed.SetConditions(xpv1.Creating(), xpv1.ReconcileError(errors.Wrap(err, errReconcileCreate)))
```

the `if err := r.managed.UpdateCriticalAnnotations(ctx, managed); err != nil {` creates a new err variable, and does not overwrite the create error (https://go.dev/play/p/DXtx63QBA3v), so `xpv1.ReconcileError(errors.Wrap(err, errReconcileCreate))` will contain only the info of that the create failed. maybe that is what we want, but I was wondering that if these annotations are so critical to add, in order to avoid dead objects, that maybe is worth propagating to status as well?

luxas:
> There is still a risk that an object becomes bricked, e.g.:
>   when the ctx is cancelled after some timeout (30s or so per reconcile?). I.e. if the error persists for long enough.
>   when the controller crashes before apply the critical annotations.

Isn't there any solution to retry without the annotations? What does it take to do a reverse-lookup of the external resource to see if it indeed was created before the controller crashed, and then delete or update it on a new reconcile?

negz:
> What does it take to do a reverse-lookup of the external resource to see if it indeed was created before the controller crashed, and then delete or update it on a new reconcile?

@luxas as far as I know, there's no way to do this. Especially no way that will work for every external system we want to reconcile. I opened https://github.com/crossplane/docs/pull/688 to (finally) document what these annotations are doing and why.

PR #666, opened 2024-02-06
Add a nil case to the `getValueFromInterface` function

sergenyalcin:

This PR adds a nil case to the `getValueFromInterface` function to handle the case that the value of the segment is nil. This [issue](https://github.com/upbound/provider-aws/pull/1118#issuecomment-1922017676) was observed while testing the conversion functions in provider-aws.

This PR addresses the same issue as [here](https://github.com/crossplane/crossplane-runtime/pull/627), but for a different function.

I have:

Tested locally for the observed case.

### DIFF
--- a/pkg/fieldpath/paved.go
+++ b/pkg/fieldpath/paved.go
@@ -112,7 +112,9 @@ func (p *Paved) getValue(s Segments) (any, error) {
 	return getValueFromInterface(p.object, s)
 }
 
-func getValueFromInterface(it any, s Segments) (any, error) {
+func getValueFromInterface(it any, s Segments) (any, error) { //nolint:gocyclo // See note below.
+	// Although the complexity of the function may seem high, in fact the same
+	// operation is performed in different cases.
 	for i, current := range s {
 		final := i == len(s)-1
 		switch current.Type {
@@ -129,18 +131,21 @@ func getValueFromInterface(it any, s Segments) (any, error) {
 			}
 			it = array[current.Index]
 		case SegmentField:
-			object, ok := it.(map[string]any)
-			if !ok {
+			switch object := it.(type) {
+			case map[string]any:
+				v, ok := object[current.Field]
+				if !ok {
+					return nil, errNotFound{errors.Errorf("%s: no such field", s[:i+1])}
+				}
+				if final {
+					return v, nil
+				}
+				it = object[current.Field]
+			case nil:
+				return nil, errNotFound{errors.Errorf("field %q is not found in the path", s[:i])}
+			default:
 				return nil, errors.Errorf("%s: not an object", s[:i])
 			}
-			v, ok := object[current.Field]
-			if !ok {
-				return nil, errNotFound{errors.Errorf("%s: no such field", s[:i+1])}
-			}
-			if final {
-				return v, nil
-			}
-			it = object[current.Field]
 		}
 	}
 
--- a/pkg/fieldpath/paved.go
+++ b/pkg/fieldpath/paved.go
@@ -142,7 +142,7 @@ func getValueFromInterface(it any, s Segments) (any, error) { //nolint:gocyclo /
 				}
 				it = object[current.Field]
 			case nil:
-				return nil, errNotFound{errors.Errorf("path %q is not found in the paved object", s[:i])}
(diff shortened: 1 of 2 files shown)


pkg/fieldpath/paved.go
```diff
@@ -129,18 +131,21 @@ func getValueFromInterface(it any, s Segments) (any, error) {
...
+				v, ok := object[current.Field]
+				if !ok {
+					return nil, errNotFound{errors.Errorf("%s: no such field", s[:i+1])}
+				}
+				if final {
+					return v, nil
+				}
+				it = object[current.Field]
+			case nil:
+				return nil, errNotFound{errors.Errorf("field %q is not found in the path", s[:i])}
```
ulucinar, line 145: nit:
```suggestion
				return nil, errNotFound{errors.Errorf("path %q is not found in the paved object", s[:i])}
```

pkg/fieldpath/paved_test.go
```diff
@@ -161,6 +161,14 @@ func TestGetValue(t *testing.T) {
 				err: errors.Wrap(errors.New("unexpected ']' at position 5"), "cannot parse path \"spec[]\""),
 			},
 		},
+		"NilValue": {
+			reason: "Requesting for an object that has nil value",
+			path:   "spec.containers[*].name",
+			data:   []byte(`{"spec":{"containers": null}}`),
+			want: want{
+				err: errNotFound{errors.Errorf("field %q is not found in the path", "spec.containers")},
```
ulucinar, line 169: nit:
```suggestion
				err: errNotFound{errors.Errorf("path %q is not found in the paved object", "spec.containers")},
```

pkg/fieldpath/paved_test.go
```diff
@@ -161,6 +161,14 @@ func TestGetValue(t *testing.T) {
 				err: errors.Wrap(errors.New("unexpected ']' at position 5"), "cannot parse path \"spec[]\""),
 			},
 		},
+		"NilValue": {
+			reason: "Requesting for an object that has nil value",
```
ulucinar, line 165: nit: I think we had better tell the nil value is not a leaf but rather an intermediate component in the requested path (a parent):
```suggestion
		"NilParent": {
			reason: "Request for a path with a nil parent value",
```

ulucinar (approved):
Thanks @sergenyalcin, lgtm.

pkg/fieldpath/paved.go
sergenyalcin: Done!

pkg/fieldpath/paved_test.go
sergenyalcin: Done!

pkg/fieldpath/paved_test.go
sergenyalcin: Done!

pkg/fieldpath/paved.go
```diff
@@ -129,18 +131,21 @@ func getValueFromInterface(it any, s Segments) (any, error) {
...
+				v, ok := object[current.Field]
+				if !ok {
+					return nil, errNotFound{errors.Errorf("%s: no such field", s[:i+1])}
+				}
+				if final {
+					return v, nil
+				}
+				it = object[current.Field]
+			case nil:
+				return nil, errNotFound{errors.Errorf("path %q is not found in the paved object", s[:i])}
```
phisco, line 145: Nit: Would it make sense to have it more consistent with the other error above? Something like `%s: expected map, got nil`

turkenh (approved):
LGTM, thanks!

pkg/fieldpath/paved.go
sergenyalcin: Done!

haarchri:
any reason why this PR is not with label for backport to v1.15 release ? - looks like we started using 1.16 rc for providers https://github.com/upbound/provider-aws/blob/main/go.mod#L13 after https://github.com/upbound/provider-aws/pull/1159

lsviben:
> any reason why this PR is not with label for backport to v1.15 release ? - looks like we started using 1.16 rc for providers https://github.com/upbound/provider-aws/blob/main/go.mod#L13 after https://github.com/upbound/provider-aws/pull/1159

Lets backport it

PR #672, opened 2024-02-22
controller/engine: add queueing index method to NamedController

sttts:

Using `GetCache()`'s `IndexField` method will use the passed context for the life-cycle of informers started at that moment. This is usually not what is desired, but the life-cycle of the cache should match the life-cycle of the controller, and this applies to all informers of course.

This PR adds a `IndexField` method to queue up index creation until the controller has been started.

Note that the `GetCache()` method is still useful to e.g. wire it up into other components that want to read from the cache later.

I have:

Tested in Crossplane fixing realtime composition issues.

### DIFF
--- a/pkg/controller/engine.go
+++ b/pkg/controller/engine.go
@@ -187,8 +187,18 @@ func (e *Engine) Start(name string, o controller.Options, w ...Watch) error {
 // NamedController is a controller that's not yet started. It gives access to
 // the underlying cache, which may be used e.g. to add indexes.
 type NamedController interface {
+	// Start the named controller. Start does not block.
 	Start(ctx context.Context) error
+	// GetCache returns the cache used by the named controller. If the
+	// controller hasn't been started, neither has the cache.
+	//
+	// Warning: Be careful when calling methods with a context before the
+	// controller has been started. The pass context will be used to e.g. start
+	// informers, and the life-cycle will differ from that of the controller. In
+	// particular, use IndexField instead to add indexes to the cache before the
+	// controller is started.
 	GetCache() cache.Cache
+	IndexField(obj client.Object, field string, extractValue client.IndexerFunc)
 }
 
 type namedController struct {
@@ -196,6 +206,14 @@ type namedController struct {
 	e    *Engine
 	ca   cache.Cache
 	ctrl controller.Controller
+
+	indexes []index
+}
+
+type index struct {
+	obj          client.Object
+	field        string
+	extractValue client.IndexerFunc
 }
 
 // Create the named controller. Each controller gets its own cache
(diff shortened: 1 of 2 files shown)


pkg/controller/engine.go
```diff
@@ -187,15 +187,33 @@ func (e *Engine) Start(name string, o controller.Options, w ...Watch) error {
...
+	// GetCache returns the cache used by the named controller. If the
+	// controller hasn't been started, neither has the cache.
+	//
+	// Warning: Be careful when calling methods with a context before the
+	// controller has been started. The pass context will be used to e.g. start
+	// informers, and the life-cycle will differ from that of the controller. In
+	// particular, use IndexField instead to add indexes to the cache before the
+	// controller is started.
 	GetCache() cache.Cache
+	IndexField(obj client.Object, field string, extractValue client.IndexerFunc)
```
negz, line 201: Could you add a doc string for this method, since the others all have them.

pkg/controller/engine.go
```diff
@@ -281,7 +305,18 @@ func (c *namedController) Start(ctx context.Context) error {
 	return nil
 }
 
+// IndexField queues an index for addition to the cache on start.
```
negz, line 308: Would `QueueIndexField` or `IndexFieldOnStart` or something similar make this more self-descriptive?

What happens if this is called after the controller/cache is started? Nothing I guess? (Worth mentioning in doc string.)

pkg/controller/engine.go
```diff
@@ -187,15 +187,33 @@ func (e *Engine) Start(name string, o controller.Options, w ...Watch) error {
...
 type namedController struct {
 	name string
 	e    *Engine
 	ca   cache.Cache
 	ctrl controller.Controller
+
+	indexes []index
+}
+
+type index struct {
```
negz, line 213: Would it make sense to export this type and have Start take a variadic array of it? I'm not sure how this is called, so I'm happy if it's more ergonomic to use the method approach instead.

e.g.:

```go
c.Start(ctx,
    Index{Object: o, Field: "spec.hi", ExtractValue: ifn},
    Index{Object: o, Field: "spec.hello", ExtractValue: ifn2})
```

pkg/controller/engine.go
sttts: Good points.

sttts: Actually, standby with this PR. The context passed to `GetInformer` and with that to `IndexField` is only used for waiting, not for the informer life-cycle. I think this was a misunderstanding.

negz:
https://github.com/crossplane/crossplane-runtime/pull/689 moved the engine implementation to c/c. The implementation is being reworked there so this PR may no longer be relevant but either way it won't happen in this repo. 🙂

PR #675, opened 2024-02-26
Add a "remote" `ExternalClient` implementation

negz:

Fixes https://github.com/crossplane/crossplane/issues/4723

This PR adds a "remote" `ExternalClient` implementation. The idea is that the controller frontend of a provider could be decoupled from the CRUD logic at the `ExternalClient` line. The provider could use an `ExternalClient` that made RPCs to a backend that implemented the CRUD logic. See https://github.com/crossplane/crossplane/issues/4723 for the motivation.

I'm opening a draft because I realized this direction won't work without larger changes.

The problem is that most (all?) `ExternalConnector` implementations assume they will run inside a Kubernetes cluster and have an API server client. They need this to load ProviderConfigs and connection secrets. They may also need to be in a cluster to setup IRSA.

This means that moving the real `ExternalConnector` into a backend that doesn't have access to a Kubernetes cluster won't work. 

I have:


negz:
I think for this direction to work we'd need to isolate the `ExternalConnector` from the API server. We'd need to load credentials and pass them into an `ExternalConnector`. Its signature would become more like:

```go
/ An Connector produces a new Client for an external system given the supplied
// managed resource.
type Connector interface {
	// Connect to the provider specified by the supplied managed resource and
	// produce an Client.
	Connect(ctx context.Context, mg resource.Managed, pc resource.ProviderConfig, creds ...Credential) (Client, error)
}
```

Issue #555, opened 2023-09-21
Expose MR Metrics


humoflife:
### What problem are you facing?

Crossplane users would benefit from the ability to offer service level objectives (SLOs) to their users. To do this they need MR related metrics.

### How could Crossplane help solve your problem?

At the Crossplane pod port 8080/metrics interface, expose the following:
- number of MRs total, by claim, by composition
- number of MRs in Synched False State total, by claim, by composition
- number of MRs in Synched True State total, by claim, by composition
- number of MRs in Synched Unkown State total, by claim, by composition
- number of MRs in Ready False State total, by claim, by composition
- number of MRs in Ready True State total, by claim, by composition
- number of MRs in Ready Unknown State total, by claim, by composition
- number of MRs in Synched True and Ready True State total, by claim, by composition
- number of MRs in Synched False and Ready True State total, by claim, by composition
- number of MRs in Synched Unknown and Ready True State total, by claim, by composition
- number of MRs in Synched True and Ready False State total, by claim, by composition
- number of MRs in Synched False and Ready False State total, by claim, by composition
- number of MRs in Synched Unknown and Ready False State total, by claim, by composition
- for each MR, time to readiness (Synched True -> Ready True)
- for each MR, time to non-readiness (Ready True -> Synched !True)
- for each MR, number of Synched state changes
- for each MR, number of Ready state changes
- for each MR, time from deletion request to external resource and MR deletion
- for each claim, time to readiness (from claim created to ready)
- for each claim, time to delete all associated resources (from claim to external resources removed)

- if possible, control plane availability and uptime (since last Crossplane pod restart).

The implementation of the metric gathering could be performed in the Crossplane runtime reconciler code, line 675+.
https://github.com/crossplane/crossplane-runtime/blob/master/pkg/reconciler/managed/reconciler.go#L675

An example time to readiness calculation is located in Uptest at https://github.com/upbound/uptest/blob/6e567ebd9ed30f1b1670d2cbbb679fde9beebc6b/cmd/perf/internal/managed/managed.go#L171

negz:
Something to double-check for whoever works on this: do we need to add all of these metrics as individual time series? Let's make sure we focus on adding low-level, flexible time series that can be composed to produce useful higher level metrics.

blut:
For anyone working with crossplane https://github.com/crossplane-contrib/x-metrics might provide useful metrics to aggregate the requested metrics.

pierluigilenoci:
[...]

PR #683, opened 2024-04-03
Introduce High Level MR metrics

ezgidemirel:
This pull request introduces the following high level metrics in the MR reconciler and cherry-picks the relevant [commit](https://github.com/crossplane/crossplane-runtime/pull/489/commits/cc9e48612922a8e5f14a9f6ed833333f539e6f10) in this [PR](https://github.com/crossplane/crossplane-runtime/pull/489).
- crossplane_managed_resource_exists{"gvk"}
- crossplane_managed_resource_ready{"gvk"}
- crossplane_managed_resource_synced{"gvk"}
- crossplane_managed_resource_first_time_to_reconcile_seconds{"gvk"}
- crossplane_managed_resource_first_time_to_readiness_seconds{"gvk"}
- crossplane_managed_resource_deletion_seconds{"gvk"}
- crossplane_managed_resource_drift_seconds{"gvk"}

The `crossplane_managed_resource_time_to_first_reconcile_seconds` and `crossplane_managed_resource_first_time_to_readiness_seconds` metrics are pushed only once for each resource, capturing the initial reconciliation time after creation and the first instance the resource becomes ready, respectively. It's important to note that these metrics won't be recorded in the event of a pod restart. One potential solution to this issue is to introduce an additional status subresource, such as "observations," where these values can be stored persistently.

Example provider PR which consumes changes-> https://github.com/crossplane-contrib/provider-kubernetes/pull/224

Fixes #

I have:


pkg/reconciler/managed/metrics.go
```diff
@@ -0,0 +1,204 @@
...
+
+var (
+	drift = prometheus.NewHistogramVec(prometheus.HistogramOpts{ //nolint:gochecknoglobals // metrics should be registered once in init
+		Subsystem: subSystem,
+		Name:      "resource_drift_seconds",
+		Help:      "ALPHA: How long since the previous successful reconcile when a resource was found to be out of sync; excludes restart of the provider",
+		Buckets:   kmetrics.ExponentialBuckets(10e-9, 10, 10),
+	}, []string{"group", "kind"})
+
+	mr = prometheus.NewGaugeVec(prometheus.GaugeOpts{ //nolint:gochecknoglobals // metrics should be registered once in init
```
negz, line 51: Is it possible to use counters, rather than gauges, to track the number of MRs?

Here's a good explanation of why: https://youtu.be/eq4CnIzw-pE?si=LAhhDrgrqBtrcacy&t=1008

In short, counters are nice because they don't lose meaning when sampled. Gauges can.

If we want to answer the question "how many MRs exist?" we can do that most accurately with two counters:

```
crossplane_managed_resources_created_total = 42
crossplane_managed_resources_deleted_total = 10
```

42 - 10 = 32 MRs exist.

You can also answer questions like "how many MRs were created over the last month" accurately, which you can't do with a gauge - we'd miss 10 MRs being created then deleted between sampling points.

The downside of counters is they don't work so well for questions like "how many MRs are currently ready?" when MRs can transition back and forth between being ready and unready (or being deleted).

We might want to consider a combination of counters and gauges. For example:

```
crossplane_managed_resources_created_total = 42
crossplane_managed_resources_deleted_total = 10
crossplane_managed_resources{condition_status_ready="True",condition_status_synced="True"} = 30
```

pkg/reconciler/managed/metrics.go
```diff
@@ -0,0 +1,204 @@
...
+
+	mrReady = prometheus.NewGaugeVec(prometheus.GaugeOpts{ //nolint:gochecknoglobals // metrics should be registered once in init
+		Subsystem: subSystem,
+		Name:      "managed_resource_ready",
+		Help:      "The number of managed resources in Ready=True state",
+	}, []string{"gvk", "name", "claim", "composite"})
+
+	mrReadyDuration = prometheus.NewHistogramVec(prometheus.HistogramOpts{ //nolint:gochecknoglobals // metrics should be registered once in init
+		Subsystem: subSystem,
+		Name:      "managed_resource_ready_duration_seconds",
```
negz, line 65: Nit: This metric name reads to me as how long the MR has been ready, not how long it took to become ready.

pkg/reconciler/managed/metrics.go
```diff
@@ -0,0 +1,204 @@
...
+		Help:      "The time it took for a managed resource to be deleted",
+		Buckets:   []float64{1, 5, 10, 15, 30, 60, 120, 300, 600, 1800, 3600},
+	}, []string{"gvk", "name", "claim", "composite"})
+)
+
+type metricRecorder struct {
+	firstObservation sync.Map
+	lastObservation  sync.Map
+
+	cluster cluster.Cluster
```
negz, line 95: Does this need the whole `cluster.Cluster`, or just the informer for the relevant GVK?

pkg/reconciler/managed/metrics.go
```diff
@@ -0,0 +1,204 @@
...
+	gvk     schema.GroupVersionKind
+}
+
+func (r *metricRecorder) Start(ctx context.Context) error {
+	inf, err := r.cluster.GetCache().GetInformerForKind(ctx, r.gvk)
+	if err != nil {
+		return errors.Wrapf(err, "cannot get informer for metric recorder for resource %s", r.gvk)
+	}
+
+	registered, err := inf.AddEventHandler(cache.ResourceEventHandlerFuncs{
```
negz, line 105: Will this definitely notice every deleted MR?

Is there a reason to do this cleanup in an edge-driven informer function, rather than doing it immediately after calling `r.metricRecorder.recordDeleted`?

pkg/reconciler/managed/metrics.go
```diff
@@ -0,0 +1,204 @@
...
+	mrReadyDuration = prometheus.NewHistogramVec(prometheus.HistogramOpts{ //nolint:gochecknoglobals // metrics should be registered once in init
+		Subsystem: subSystem,
+		Name:      "managed_resource_ready_duration_seconds",
+		Help:      "The time it took for a managed resource to become ready first time after creation",
+		Buckets:   []float64{1, 5, 10, 15, 30, 60, 120, 300, 600, 1800, 3600},
+	}, []string{"gvk", "name", "claim", "composite"})
+
+	mrDetected = prometheus.NewHistogramVec(prometheus.HistogramOpts{ //nolint:gochecknoglobals // metrics should be registered once in init
+		Subsystem: subSystem,
+		Name:      "managed_resource_detected_time_seconds",
```
negz, line 72:
```suggestion
		Name:      "managed_resource_time_to_first_reconcile_seconds",
```

Would this name be more meaningful to readers?

pkg/reconciler/managed/metrics.go
```diff
@@ -0,0 +1,204 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+
+	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
+	"github.com/crossplane/crossplane-runtime/pkg/errors"
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
+)
+
+func init() { //nolint:gochecknoinits // metrics should be registered once
+	metrics.Registry.MustRegister(drift, mr, mrReady, mrSynced, mrDetected, mrReadyDuration, mrDeletionDuration)
+}
```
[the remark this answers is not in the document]
ezgidemirel: I've already tried this pattern. As you suggested the best (maybe the only) place to initialize and register the metrics is the main function which comes from Upjet for official providers. Passing through these metrics to the reconciler is not straightforward and will make things more complicated than it should be. We also need to think about other providers which are not generated by Upjet. If we decide to follow this pattern, I guess following steps need to be done:
- Define which metrics to have in crossplane-runtime
- Merge and consume that change in Upjet
- Initialize and register those metrics in Upjet and pass them to the controllers (I'm still not sure how complicated this will be)
- In MR reconciler, make sure that those metrics are passed and record metrics.

I'm not sure what we would achieve with this is worth it.

pkg/reconciler/managed/metrics.go
```diff
@@ -0,0 +1,204 @@
...
+		Subsystem: subSystem,
+		Name:      "managed_resource_created",
+		Help:      "The number of managed resources created",
+	}, []string{"gvk", "name", "claim", "composite"})
+
+	mrReady = prometheus.NewGaugeVec(prometheus.GaugeOpts{ //nolint:gochecknoglobals // metrics should be registered once in init
+		Subsystem: subSystem,
+		Name:      "managed_resource_ready",
+		Help:      "The number of managed resources in Ready=True state",
+	}, []string{"gvk", "name", "claim", "composite"})
```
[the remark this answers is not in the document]
ezgidemirel: I think if I don't have MR name as label, I need to store which MRs that I recorded in a map and check whether I sent a metric for that MR or not. Since these metrics are recorded in a reconciliation loop, it's easy to send duplicate metrics for the same MR and the solution I found is using MR name as label to have unique identifier for metrics. 

Regarding the cardinality, we may remove `claim` and `composite` labels. In the MR metrics [issue](https://github.com/crossplane/crossplane-runtime/issues/555), it's also requested to classify metrics per compositions which requires a change in Crossplane.

pkg/reconciler/managed/metrics.go
```diff
@@ -0,0 +1,204 @@
...
+		Name:      "resource_drift_seconds",
+		Help:      "ALPHA: How long since the previous successful reconcile when a resource was found to be out of sync; excludes restart of the provider",
+		Buckets:   kmetrics.ExponentialBuckets(10e-9, 10, 10),
+	}, []string{"group", "kind"})
+
+	mr = prometheus.NewGaugeVec(prometheus.GaugeOpts{ //nolint:gochecknoglobals // metrics should be registered once in init
+		Subsystem: subSystem,
+		Name:      "managed_resource_created",
+		Help:      "The number of managed resources created",
+	}, []string{"gvk", "name", "claim", "composite"})
```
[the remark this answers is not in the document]
ezgidemirel: Yes, that was my first attempt :) The problem I faced again was sending multiple metrics for the same MR each time the condition combination updated.

pkg/reconciler/managed/metrics.go
ezgidemirel: Maybe I'm missing something, but again it's the same problem I've been mentioning about. It's hard to decide when to increase the counter metric in a reconciliation loop.

pkg/reconciler/managed/metrics.go
ezgidemirel: what do you think about `managed_resource_time_to_readiness_seconds`?

pkg/reconciler/managed/metrics.go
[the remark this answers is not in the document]
negz: I see it as adopting the same pattern we already have for other instrumentation - e.g. loggers and event handlers. We initialize these in main and plumb them down to the reconciler implementation.

Yes, this is more effort, but I feel it's worth it. A _library_ like crossplane-runtime should offer you instrumentation options but shouldn't force choices on you, for example by forcing a specific logger or metrics library inherently because you imported a package.

Passing instrumentation implementations as dependencies also allows us to pass in stub/fake implementations during testing if we want to, for example if we wanted to test that we were emitting logs or metrics correctly.

Some interesting reading on the topic:

* https://peter.bourgon.org/blog/2017/06/09/theory-of-modern-go.html
* https://dave.cheney.net/2017/06/11/go-without-package-scoped-variables
* https://github.com/prometheus/docs/issues/506
* https://groups.google.com/g/prometheus-developers/c/l5dXdTkJtl0

It's worth noting that Prom maintainers (e.g. Brian and Björn) are split on this. I'm more aligned with Björn. Perhaps if this wasn't a library the path of least resistance - using global state - would be fine. In this case, it is not.

pkg/reconciler/managed/metrics.go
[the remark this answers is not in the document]
negz: Unfortunately I'm pretty sure including the MR name in metrics is a non-starter cardinality wise. We would be producing a timeseries for every MR that ever existed during the provider's lifetime (i.e. between restarts). If we think of a control plane with 1,000 MRs, that means:

* Every metric with an MR name label gets 1,000 entries at the `/metrics` http endpoint
* Prometheus (or Prom-compatible scraping agent) must store 1,000 entries in its time series database
* Queries, aggregation rules, etc must process 1,000 entries

I don't think removing claim/composite labels will help much.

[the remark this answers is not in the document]
negz: See also https://prometheus.io/docs/practices/instrumentation/#do-not-overuse-labels

pkg/reconciler/managed/metrics.go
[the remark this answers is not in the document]
negz: This could be a sign that this reconciler isn't the right place to be tracking these metrics.

Kubernetes frames this well [here](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/metric-instrumentation.md#instrumentation-types):

> Components have metrics capturing events and states that are inherent to their application logic. Examples are request and error counters, request latency histograms, or internal garbage collection cycles. Those metrics are instrumented directly in the application code.
>
> Secondly, there are business logic metrics. Those are not about observed application behavior but abstract system state, such as desired replicas for a deployment. They are not directly instrumented but collected from otherwise exposed data.

In the context of this PR, we could ask:

* Which metrics are helping us understand how well the MR reconciler you're instrumenting is behaving?
* Which metrics are instead telling us about the state of the system as a whole?

I would argue these metrics are telling us how the reconciler is behaving:

* managed_resource_detected_time_seconds - Average time to notice a resource exists
* managed_resource_ready_duration_seconds - Average time to drive a resource to readiness
* managed_resource_deletion_seconds - Average time to delete a resource
* resource_drift_seconds - Average time to correct drift for a resource

Whereas these metrics are just telling us about the state of the system:

* managed_resource_created - Number of resources that exist
* managed_resource_ready - Number of resources that are ready
* managed_resource_synced - Number of resources that are synced

In Kubernetes, kube-state-metrics would monitor things like this:

> In Kubernetes they are generally captured in the [kube-state-metrics](https://github.com/kubernetes/kube-state-metrics) component, which reads them from the API server. For this types of metric exposition, the [exporter guidelines](https://prometheus.io/docs/instrumenting/writing_exporters/) apply additionally.

There's prior art for kube-state-metrics for Crossplane in https://github.com/crossplane-contrib/x-metrics.

I'm not necessarily making the case that these metrics should move into x-metrics. It would be more convenient if you didn't need to run an extra component to get access to things like resource counts, especially given our resources have a lot of dynamic types.

One way to do handle this could be to add a new crossplane-runtime package that a provider can import to expose KSM-style metrics. I could imagine rather than being a controller, such a package would simply occasionally list a particular kind of MR and expose metrics about how many existed, were ready, etc.

pkg/reconciler/managed/metrics.go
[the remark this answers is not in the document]
ezgidemirel: I've switched to the pattern you suggested.

pkg/reconciler/managed/metrics.go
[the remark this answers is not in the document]
ezgidemirel: Thanks for your detailed review and sharing all these links. They helped a lot! It explains why I was struggling to record those in the reconciler. 

I've moved the state metrics from the reconciler and recording them in a goroutine in the `NewReconciler` function now. But I'm not sure if there might be a better place to call it.

pkg/reconciler/managed/metrics.go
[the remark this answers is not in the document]
sttts: +1 for @negz 's point. We can have GVKs in labels, we CANNOT have names. In general, cardinality must never depend on values the user providers (in Crossplane case that is the claim owner).

Another example: Kube had a CVE long time ago with the http user agent in a label. That could be used to DoS kube-apiserver from outside, even non-authenticated.

pkg/statemetrics/xr_state_metrics.go
```diff
@@ -0,0 +1,160 @@
...
+	r.compositeReady.Collect(ch)
+	r.compositeSynced.Collect(ch)
+	r.compositeComposedCount.Collect(ch)
+}
+
+// Record records the state of managed resources.
+func (r *XRStateRecorder) Record(ctx context.Context, gvk schema.GroupVersionKind) {
+	l := &unstructured.UnstructuredList{}
+	l.SetGroupVersionKind(gvk)
+	err := r.client.List(ctx, l)
```
sttts, line 105: Don't list from a client. List from a cache, i.e. pass in a cache here. Listing can be expensive, imagine 10000 objects.

pkg/reconciler/managed/metrics.go
```diff
@@ -0,0 +1,186 @@
...
+
+func (r *NopMetricRecorder) recordDeleted(_ resource.Managed) {}
+
+func (r *NopMetricRecorder) recordFirstTimeReady(_ resource.Managed) {}
+
+func getMRMetricLabels(managed resource.Managed) prometheus.Labels {
+	l := prometheus.Labels{
+		"gvk":       managed.GetObjectKind().GroupVersionKind().String(),
+		"claim":     "",
+		"composite": managed.GetLabels()["crossplane.io/composite"],
```
sttts, line 178: this is also a name of an object, isn't it? We cannot use that as label.

pkg/reconciler/managed/metrics.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	if managed.GetLabels()["crossplane.io/claim-namespace"] != "" && managed.GetLabels()["crossplane.io/claim-name"] != "" {
+		l["claim"] = managed.GetLabels()["crossplane.io/claim-namespace"] + "/" + managed.GetLabels()["crossplane.io/claim-name"]
```
sttts, line 182: same here. Neither namespace nor name can be used in a label.

pkg/reconciler/managed/reconciler.go
```diff
@@ -701,12 +718,17 @@ func NewReconciler(m manager.Manager, of resource.ManagedKind, o ...ReconcilerOp
...
+		metricRecorder:              NewNopMetricRecorder(),
+		stateRecorder:               statemetrics.NewNopStateRecorder(),
 	}
 
 	for _, ro := range o {
 		ro(r)
 	}
 
+	// State recorder is started in the background to record MR states.
+	go r.stateRecorder.Run(context.Background(), schema.GroupVersionKind(of))
```
sttts, line 730: Never leak go routines from a constructor. This must be started outside with an available`ctx`.

pkg/statemetrics/mr_state_metrics.go
```diff
@@ -0,0 +1,141 @@
...
+	r.mrExists.Collect(ch)
+	r.mrReady.Collect(ch)
+	r.mrSynced.Collect(ch)
+}
+
+// Record records the state of managed resources.
+func (r *MRStateRecorder) Record(ctx context.Context, gvk schema.GroupVersionKind) {
+	l := &unstructured.UnstructuredList{}
+	l.SetGroupVersionKind(gvk)
+	err := r.client.List(ctx, l)
```
sttts, line 95: also here: pass a cache please to make sure this does not hit the apiserver. This can be a very heavy operation, listing thousands of MRs.

pkg/statemetrics/mr_state_metrics.go
```diff
@@ -0,0 +1,141 @@
...
+)
+
+// A MRStateRecorderOption configures a MRStateRecorder.
+type MRStateRecorderOption func(*MRStateRecorder)
+
+// A MRStateRecorder records the state of managed resources.
+type MRStateRecorder struct {
+	client    client.Client
+	log       logging.Logger
+	frequency time.Duration
```
sttts, line 41: nit: this is rather an interval than a frequency.

pkg/statemetrics/xr_state_metrics.go
```diff
@@ -0,0 +1,160 @@
...
+		conditioned := xpv1.ConditionedStatus{}
+		if err := fieldpath.Pave(o.Object).GetValueInto("status", &conditioned); err != nil {
+			r.log.Info("Failed to get conditions of managed resource", "error", err)
+			continue
+		}
+
+		for _, condition := range conditioned.Conditions {
+			if condition.Type == xpv1.TypeReady && condition.Status == corev1.ConditionTrue {
+				numReady++
+			} else if condition.Type == xpv1.TypeSynced && condition.Status == corev1.ConditionTrue {
```
sttts, line 128: is this really an "else"?

pkg/statemetrics/mr_state_metrics.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	r.mrReady.WithLabelValues(label).Set(numReady)
+	r.mrSynced.WithLabelValues(label).Set(numSynced)
+}
+
+// Run records state of managed resources with given frequency.
+func (r *MRStateRecorder) Run(ctx context.Context, gvk schema.GroupVersionKind) {
+	ticker := time.NewTicker(r.frequency)
+	quit := make(chan struct{})
```
sttts, line 129: who closes `quit`?

pkg/statemetrics/xr_state_metrics.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	r.compositeReady.With(labels).Set(numReady)
+	r.compositeSynced.With(labels).Set(numSynced)
+}
+
+// Run records state of managed resources with given frequency.
+func (r *XRStateRecorder) Run(ctx context.Context, gvk schema.GroupVersionKind) {
+	ticker := time.NewTicker(r.frequency)
+	quit := make(chan struct{})
```
sttts, line 148: who closes `quit`?

pkg/statemetrics/mr_state_metrics.go
negz: Would it be sufficient to accept `client.Reader` and add a comment letting the caller know it must be cache-backed?

I don't feel strongly - mostly wondering if there's some way to do this up at the controller-runtime level of abstraction without moving down into client-go things. Would be nice, but not critical.

negz:
Thanks @ezgidemirel! Still needs some work per @sttts's comments but I like this direction much better.

pkg/statemetrics/mr_state_metrics.go
ezgidemirel: What do you think about passing the following option to the `NewManager` function in provider's `main.go`?
```
mgr, err := ctrl.NewManager(ratelimiter.LimitRESTConfig(cfg, *maxReconcileRate), ctrl.Options{
	Client: client.Options{Cache: &client.CacheOptions{Unstructured: true}},
	...
})
```
This makes `shouldByPassCache` method to return false in the [List](https://github.com/kubernetes-sigs/controller-runtime/blob/7ff5801735fbeabed179125571bfd9e6e73068d3/pkg/client/client.go#L374) call.

pkg/statemetrics/xr_state_metrics.go
ezgidemirel: `condition.Type` can only be `TypeReady` or `TypeSynced`

pkg/reconciler/managed/reconciler.go
ezgidemirel: I made state recorders `Runnable` and registered to manager to manage go routine's lifecycle with it.

pkg/statemetrics/mr_state_metrics.go
ezgidemirel: I no longer use unstructured list, so we'll use client's cache.

pkg/reconciler/managed/metrics.go
```diff
@@ -0,0 +1,176 @@
...
+	mrDeletion       *prometheus.HistogramVec
+	mrDrift          *prometheus.HistogramVec
+}
+
+// NewMRMetricRecorder returns a new MRMetricRecorder which records metrics for managed resources.
+func NewMRMetricRecorder() *MRMetricRecorder {
+	return &MRMetricRecorder{
+		mrDetected: prometheus.NewHistogramVec(prometheus.HistogramOpts{
+			Subsystem: subSystem,
+			Name:      "managed_resource_time_to_first_reconcile_seconds",
```
mergenci, line 61: Nit: Does it make sense to make the following names consistent with each other: `managed_resource_time_to_first_reconcile_seconds`, `managed_resource_first_time_to_readiness_seconds`? `time_to_first` sounds easier to the ear, so we can make the latter `managed_resource_time_to_first_readiness_seconds`, to agree with the former. On the other hand, who are familiar with time to readiness (TTR) might find `first_time_to_readiness` more natural. In that case we can leave them as they are. Changing the former to `managed_resource_first_time_to_reconcile_seconds` sounds weird to me.

mergenci (changes requested):

pkg/reconciler/managed/metrics.go
```diff
@@ -0,0 +1,176 @@
...
+	r.mrDrift.Collect(ch)
+}
+
+func (r *MRMetricRecorder) recordUnchanged(name string) {
+	r.lastObservation.Store(name, time.Now())
+}
+
+func (r *MRMetricRecorder) recordFirstTimeReconciled(managed resource.Managed) {
+	if managed.GetCondition(xpv1.TypeSynced).Status == corev1.ConditionUnknown {
+		r.mrDetected.WithLabelValues("gvk", managed.GetObjectKind().GroupVersionKind().String()).
```
mergenci, line 112:
```suggestion
		r.mrDetected.WithLabelValues(managed.GetObjectKind().GroupVersionKind().String()).
```

pkg/reconciler/managed/metrics.go
```diff
@@ -0,0 +1,176 @@
...
+	last, ok := r.lastObservation.Load(name)
+	if !ok {
+		return
+	}
+	lt, ok := last.(time.Time)
+	if !ok {
+		return
+	}
+
+	r.mrDrift.WithLabelValues("gvk", managed.GetObjectKind().GroupVersionKind().String()).
```
mergenci, line 129:
```suggestion
	r.mrDrift.WithLabelValues(managed.GetObjectKind().GroupVersionKind().String()).
```

pkg/reconciler/managed/metrics.go
```diff
@@ -0,0 +1,176 @@
...
+		Observe(time.Since(lt).Seconds())
+
+	r.lastObservation.Store(name, time.Now())
+}
+
+func (r *MRMetricRecorder) recordDeleted(managed resource.Managed) {
+	r.mrDeletion.WithLabelValues("gvk", managed.GetObjectKind().GroupVersionKind().String()).
```
mergenci, line 136:
```suggestion
	r.mrDeletion.WithLabelValues(managed.GetObjectKind().GroupVersionKind().String()).
```

pkg/reconciler/managed/metrics.go
```diff
@@ -0,0 +1,176 @@
...
+
+func (r *MRMetricRecorder) recordFirstTimeReady(managed resource.Managed) {
+	// Note that providers may set the ready condition to "True", so we need
+	// to check the value here to send the ready metric
+	if managed.GetCondition(xpv1.TypeReady).Status == corev1.ConditionTrue {
+		_, ok := r.firstObservation.Load(managed.GetName()) // This map is used to identify the first time to readiness
+		if !ok {
+			return
+		}
+		r.mrFirstTimeReady.WithLabelValues("gvk", managed.GetObjectKind().GroupVersionKind().String()).
```
mergenci, line 148:
```suggestion
		r.mrFirstTimeReady.WithLabelValues(managed.GetObjectKind().GroupVersionKind().String()).
```

mergenci (changes requested):
`"gvk"` seems to be misplaced in these instances.

pkg/reconciler/managed/metrics.go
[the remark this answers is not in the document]
humoflife: Would we still get a claim level "duration to readiness", and a claim level "duration to resources deleted" metric to allow platform creators to develop SLOs?

Technically that would only need to be 1 time series per claim.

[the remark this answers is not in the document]
negz: @humoflife It would be surprising/unusual to me to have SLOs with that level of granularity.

I would expect the type of SLO you mention to be at the _claim type_ granularity, not individual claim granularity. [^1] This would allow you to graph and alert on situations like "the P95 time for `kind: Cluster` claims to become ready is too high", but not as specific as "the `kind: Cluster` claim named `some-claim-42` took too long to become ready".

Think of monitoring an HTTP server. You'd want timeseries metrics that indicate the aggregate health of the system - the request count, error count, and average duration of requests. You wouldn't emit a timeseries for each HTTP request. You probably want to alert on "too many requests are failing", not "a request failed".

[^1]: I'd also expect such metrics to be emitted by core Crossplane, rather than making providers claim-aware.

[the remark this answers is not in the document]
negz: > This would allow you to graph and alert on situations like "the P95 time for kind: Cluster claims to become ready is too high"

This would need a `crossplane_claim_first_time_to_readiness_seconds{gvk}` timeseries to be emitted by core Crossplane. I would support adding such a timeseries. I'm not sure if it's currently planned.

negz:
@ezgidemirel Your PR description mentions some `crossplane_composite_resource_` metrics, but I don't see where those are implemented. Are you planning for them to be part of this PR? I'd expect core Crossplane to emit these metrics.

pkg/reconciler/managed/metrics.go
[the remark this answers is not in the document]
ezgidemirel: Do you think we should have two different types of metrics for claims and composites or having something like below would work? 
```
crossplane_composite_resource_first_time_to_readiness_seconds{"gvk", "claim-namespace"}
```

ezgidemirel:
> @ezgidemirel Your PR description mentions some `crossplane_composite_resource_` metrics, but I don't see where those are implemented. Are you planning for them to be part of this PR? I'd expect core Crossplane to emit these metrics.

I was planning to add the state metrics with this PR, but then removed with this [commit](https://github.com/crossplane/crossplane-runtime/pull/683/commits/1fc52f725662d6109bb1d887cc62af820a99d061). I'll update the PR description.

pkg/reconciler/managed/metrics.go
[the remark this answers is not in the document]
negz: @ezgidemirel That's a good point, I hadn't considered that you could just decorate the XR metrics with details about their claims (if any).

That said, I do think there's some value in having distinct timeseries for claims. An XR being ready doesn't necessarily mean the associated claim is ready. It's possible that the XR is ready but some issue is preventing the claim from being ready.

Without a distinct timeseries for claims, you'd never know if `Foo` claims were taking a lot longer on average than their corresponding `XFoo` XRs to become ready.

pkg/statemetrics/mr_state_metrics.go
```diff
@@ -0,0 +1,145 @@
...
+// Record records the state of managed resources.
+func (r *MRStateRecorder) Record(ctx context.Context, mrList resource.ManagedList) error {
+	if err := r.client.List(ctx, mrList); err != nil {
+		r.log.Info("Failed to list managed resources", "error", err)
+		return err
+	}
+
+	mrs := mrList.GetItems()
+	if len(mrs) == 0 {
+		return nil
```
mergenci, line 108: Issue: This early return seems to prevent setting the metric to zero when all resources are deleted.

mergenci (changes requested):

turkenh (approved):
Thanks @ezgidemirel 🙏 

@negz I see all your comments were resolved and we need to get this in to be able ship this functionality with the upcoming provider releases (i.e. tomorrow), so, I'll merge it now. Feel free add more comments if there are still open points and we can get them resolved with a follow up PR. 

@mergenci, similarly, feel free to open a follow up PR regarding your comment on early return.

eduardo-diaz-em:
> > @ezgidemirel Your PR description mentions some `crossplane_composite_resource_` metrics, but I don't see where those are implemented. Are you planning for them to be part of this PR? I'd expect core Crossplane to emit these metrics.
> 
> I was planning to add the state metrics with this PR, but then removed with this [commit](https://github.com/crossplane/crossplane-runtime/pull/683/commits/1fc52f725662d6109bb1d887cc62af820a99d061). I'll update the PR description.

Hi @ezgidemirel I am running crossplane on 1.17 and can't see the `crossplane_managed_resources_*` metrics either (the rest are working fine. According to the [docs](https://docs.crossplane.io/latest/guides/metrics/#:~:text=crossplane_managed_resource_ready) they should be available, is this a bug or am I missing something? This is my crossplane configuration from ArgoCD:

`
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: crossplane
  namespace: argocd
spec:
  project: system-apps
  source:
    chart: crossplane
    repoURL: https://charts.crossplane.io/stable
    targetRevision: 1.16.0
    helm:
      values: |
        resourcesRBACManager:
          requests:
            cpu: 250m
            memory: 512Mi
          limits:
            cpu: 250m
            memory: 512Mi
        resourcesCrossplane:
          requests:
            cpu: 250m
            memory: 512Mi
          limits:
            cpu: 250m
            memory: 512Mi
        metrics:
          enabled: true
  destination:
    server: 'https://kubernetes.default.svc'
    namespace: crossplane
  syncPolicy:
    automated:
      prune: false
      selfHeal: false
    syncOptions:
      - ServerSideApply=true
      - CreateNamespace=true
`

mergenci:
@eduardo-diaz-em, this is a crossplane-runtime PR. crossplane-runtime is used by Crossplane providers. Your provider version and which version of crossplane-runtime it uses is important, rather than your Crossplane version. For instance, [provider-upjet-aws uses crossplane-runtime v1.16.0-rc.2.0](https://github.com/crossplane-contrib/provider-upjet-aws/blob/main/go.mod#L19).

I acknowledge that it would be helpful if the documentation clarified distinction between Crossplane and Crossplane provider metrics.

pablomdc:
I'm using same xpkg.upbound.io/upbound/provider-aws-sqs:v1.9.0 which is using crossplane-runtime v1.16.0-rc.2.0.2024 along with crossplane v1.16 same happens to me, I can see all metrics but not the crossplane_managed_resources_* ones

@eduardo-diaz-em managed to get this solved?
@ezgidemirel @mergenci will appreciate any clue here

This is my v1.16 chart config : 

extraObjects:
  - apiVersion: aws.upbound.io/v1beta1
    kind: ProviderConfig
    metadata:
      name: default
    spec:
      credentials:
        source: IRSA
  - apiVersion: pkg.crossplane.io/v1beta1
    kind: DeploymentRuntimeConfig
    metadata:
      name: irsa-runtimeconfig
    spec:
      serviceAccountTemplate:
        metadata:
          name: crossplane-aws
          annotations:
            eks.amazonaws.com/role-arn: .....
  - apiVersion: pkg.crossplane.io/v1
    kind: Provider
    metadata:
      name: provider-aws-sqs
    spec:
      package: xpkg.upbound.io/upbound/provider-aws-sqs:v1.9.0
      runtimeConfigRef:
        name: irsa-runtimeconfig

pablomdc:
Please ignore, I found the problem was I was looking at crossplane pod metrics, but not at the provider pods ones :)
metrics are here: provider-aws-sqs-7a356e3a0f47-55bdc59969-vkmpt 
Noting here, in case someone faced same confusion

jbw976:
Thanks for the quick follow up @pablomdc - if anyone else stops by this PR, one helpful resource is also this demo to use the high level metrics that I had put together a few months back with help from @ezgidemirel: https://github.com/jbw976/demo-metrics

eduardo-diaz-em:
Buena esa @pablomdc I was also looking at crossplane pods rather than provider ones. Gracias!

PR #688, opened 2024-05-02
Fix MR state metrics when MRs are deleted

ezgidemirel:

In case of deleting all MRs from a GVK, we should record 0 for `crossplane_managed_resource_exists`, `crossplane_managed_resource_ready` and `crossplane_managed_resource_synced` metrics. 

Fixes #

I have:

Tested with provider-kubernetes
```
Ezgis-MacBook-Pro:phase-2 ezgidemirel$ curl localhost:8080/metrics |grep managed_resource_exists
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 30165    0 30165    0     0  25.3M      0 --:--:-- --:--:-- --:--:-- 28.7M
# HELP crossplane_managed_resource_exists The number of managed resources that exist
# TYPE crossplane_managed_resource_exists gauge
crossplane_managed_resource_exists{gvk="kubernetes.crossplane.io/v1alpha2, Kind=Object"} 0
```

### DIFF
--- a/pkg/statemetrics/mr_state_metrics.go
+++ b/pkg/statemetrics/mr_state_metrics.go
@@ -18,13 +18,16 @@ package statemetrics
 
 import (
 	"context"
+	"strings"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
 	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
+	"github.com/crossplane/crossplane-runtime/pkg/errors"
 	"github.com/crossplane/crossplane-runtime/pkg/logging"
 	"github.com/crossplane/crossplane-runtime/pkg/resource"
 )
@@ -86,9 +89,9 @@ type MRStateRecorder struct {
 }
 
 // NewMRStateRecorder returns a new MRStateRecorder which records the state of managed resources.
-func NewMRStateRecorder(client client.Client, log logging.Logger, metrics *MRStateMetrics, managedList resource.ManagedList, interval time.Duration) *MRStateRecorder {
+func NewMRStateRecorder(c client.Client, log logging.Logger, metrics *MRStateMetrics, managedList resource.ManagedList, interval time.Duration) *MRStateRecorder {
 	return &MRStateRecorder{
-		client:      client,
+		client:      c,
 		log:         log,
 		metrics:     metrics,
 		managedList: managedList,
@@ -97,19 +100,18 @@ func NewMRStateRecorder(client client.Client, log logging.Logger, metrics *MRSta
 }
 
 // Record records the state of managed resources.
-func (r *MRStateRecorder) Record(ctx context.Context, mrList resource.ManagedList) error {
-	if err := r.client.List(ctx, mrList); err != nil {
-		r.log.Info("Failed to list managed resources", "error", err)
-		return err
+func (r *MRStateRecorder) Record(ctx context.Context) error {
+	if err := r.client.List(ctx, r.managedList); err != nil {
+		return errors.Wrap(err, "failed to list managed resources")
 	}
 
-	mrs := mrList.GetItems()
-	if len(mrs) == 0 {
-		return nil
+	labels, err := r.getLabels()
+	if err != nil {
+		return errors.Wrap(err, "failed to get labels")
 	}
 
-	label := mrs[0].GetObjectKind().GroupVersionKind().String()
-	r.metrics.Exists.WithLabelValues(label).Set(float64(len(mrs)))
+	mrs := r.managedList.GetItems()
+	r.metrics.Exists.With(labels).Set(float64(len(mrs)))
 
 	var numReady, numSynced float64 = 0, 0
 	for _, o := range mrs {
@@ -122,8 +124,8 @@ func (r *MRStateRecorder) Record(ctx context.Context, mrList resource.ManagedLis
 		}
 	}
 
-	r.metrics.Ready.WithLabelValues(label).Set(numReady)
-	r.metrics.Synced.WithLabelValues(label).Set(numSynced)
+	r.metrics.Ready.With(labels).Set(numReady)
+	r.metrics.Synced.With(labels).Set(numSynced)
 
 	return nil
 }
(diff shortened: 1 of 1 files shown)


mergenci (approved):
Thank you @ezgidemirel 🙏 LGTM.

pkg/statemetrics/mr_state_metrics.go
```diff
@@ -104,11 +104,8 @@ func (r *MRStateRecorder) Record(ctx context.Context, mrList resource.ManagedLis
 	}
 
 	mrs := mrList.GetItems()
-	if len(mrs) == 0 {
-		return nil
-	}
 
-	label := mrs[0].GetObjectKind().GroupVersionKind().String()
+	label := mrList.GetObjectKind().GroupVersionKind().String()
```
negz, line 108: Does this return the same GVK it used to? I think Lists usually have a different kind, e.g. `InstanceList` instead of `Instance`.

ezgidemirel: It was returning an empty GVK, that's why I couldn't notice when I was just checking the number of MRs. I've passed the GVK info from the provider and used it to record metrics.

pkg/statemetrics/mr_state_metrics.go
```diff
@@ -81,18 +82,20 @@ type MRStateRecorder struct {
...
 	interval    time.Duration
 	managedList resource.ManagedList
+	gvk         schema.GroupVersionKind
 
 	metrics *MRStateMetrics
 }
 
 // NewMRStateRecorder returns a new MRStateRecorder which records the state of managed resources.
-func NewMRStateRecorder(client client.Client, log logging.Logger, metrics *MRStateMetrics, managedList resource.ManagedList, interval time.Duration) *MRStateRecorder {
+func NewMRStateRecorder(client client.Client, log logging.Logger, metrics *MRStateMetrics, managedList resource.ManagedList, gvk schema.GroupVersionKind, interval time.Duration) *MRStateRecorder {
```
negz, line 91:
```suggestion
func NewMRStateRecorder(c client.Client, log logging.Logger, m *MRStateMetrics, ml resource.ManagedList, gvk schema.GroupVersionKind, interval time.Duration) *MRStateRecorder {
```

Nit: With six arguments this function signature is getting pretty long. Some of the variable names could be shorter without meaningful loss of information (e.g. `c` is fine when you can see its a client from the type `client.Client`).

https://github.com/crossplane/crossplane/tree/master/contributing#dont-wrap-function-signatures

pkg/statemetrics/mr_state_metrics.go
```diff
@@ -81,18 +82,20 @@ type MRStateRecorder struct {
 	log         logging.Logger
 	interval    time.Duration
 	managedList resource.ManagedList
+	gvk         schema.GroupVersionKind
```
negz, line 85: I'm pretty sure you could avoid needing this argument by just trimming the `List` suffix from the `ManagedList`'s GVK.

ezgidemirel: I'm testing the code with provider-kubernetes and in the `Record` method, `mrList`'s GVK is empty. That's why I added a separate field for `gvk` and passed it from the provider's main.

ezgidemirel: I'm adding the recorder like below. Therefore, the metadata comes empty. I think setting the values in `&v1alpha2.ObjectList{}` is too implicit. Therefore, I'm suggesting to pass the `gvk` as a separate parameter. @negz what do you think?
```
if err := mgr.Add(statemetrics.NewMRStateRecorder(
		mgr.GetClient(), &v1alpha2.ObjectList{}, v1alpha2.ObjectGroupVersionKind, o.Logger, o.MetricOptions.MRStateMetrics, o.MetricOptions.PollStateMetricInterval)); err != nil {
		return err
	}
```

negz: As far as I can tell the only reason you pass `resource.ManagedList` is to let the `client.Client` know what GVK to query. Internally it uses controller-runtime's [`apiutil.GVKForObject`](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.18.2/pkg/client/apiutil#GVKForObject) to derive a GVK from the supplied list's Go type. So to me it seems redundant to also pass the GVK explicitly.

I would suggest either:

1. Only take a type that satisfies `resource.ManagedList` as an argument, like `&v1alpha2.ObjectList{}`, and use `GVKForObject` to extract its GVK. This will require storing a `runtime.Scheme` (from `mgr.GetScheme()` in the `MRStateRecorder`.
2. Only take a GVK as an argument, and use an `UnstructuredList{}`

Example of option 2:

```go
ul := &unstructured.UnstructuredList{}
ul.SetAPIVersion(gvk.GroupVersion().String())
ul.SetKind(gvk.Kind+"List")
_ = r.client.List(ctx, ul)
```

I prefer option 1, because clients won't cache `Unstructured` types by default.

negz:
@ezgidemirel Two other small nits I noticed that aren't related to this PR specifically. It's fine to address them separately:

* `MRStateRecorder` takes a `client.Client`, but only reads. It could take a smaller interface - `client.Reader`.
* `MRStateRecorder` is constructed with a `managed.ResourceList`, which it then passes as an argument to its own `Record` method every tick. Does `Record` need to take this as an argument? Would you ever expect `Record` to be called with a different `managed.ResourceList` from the one stored in the `MRStateRecorder`?

pkg/statemetrics/mr_state_metrics.go
negz: > This will require storing a runtime.Scheme (from mgr.GetScheme() in the MRStateRecorder.

Actually, the `client.Client` has a `Scheme` method you can use to get one.

ezgidemirel: Thanks @negz for your review. I used the `GVKForObject` to extract the GVK and removed `List` to get the object kind.

ezgidemirel:
> * `MRStateRecorder` takes a `client.Client`, but only reads. It could take a smaller interface - `client.Reader`.

To access the `Schema` I need `client.Client` now. 

> * `MRStateRecorder` is constructed with a `managed.ResourceList`, which it then passes as an argument to its own `Record` method every tick. Does `Record` need to take this as an argument? Would you ever expect `Record` to be called with a different `managed.ResourceList` from the one stored in the `MRStateRecorder`?

Removed the parameter and accessing the struct field in the `Record` method.

negz (approved):

ezgidemirel:
> @ezgidemirel we already cut the release-1.16 branch, do you want to backport this PR so that it is included in the v1.16 release?

@jbw976 yes, we can backport it.

PR #692, opened 2024-05-13
Whitelist management policy Observe & Update

turkenh:

As initially requested in [this ticket](https://github.com/crossplane-contrib/provider-kubernetes/issues/115), we have valid use cases to set a management policy as "Observe & Update", at least with Provider Kubernetes.

This PR whitelisting this variant of management policies so that runtime will not fail early.

I have:

With provider-kubernetes, `managementPolicies: ["Observe", "Update"]`.

### DIFF
--- a/pkg/reconciler/managed/policies.go
+++ b/pkg/reconciler/managed/policies.go
@@ -85,6 +85,10 @@ func defaultSupportedManagementPolicies() []sets.Set[xpv1.ManagementAction] {
 		// Like ObserveOnly, but the external resource is deleted when the
 		// managed resource is deleted.
 		sets.New[xpv1.ManagementAction](xpv1.ManagementActionObserve, xpv1.ManagementActionDelete),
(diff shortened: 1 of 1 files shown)


pkg/reconciler/managed/policies.go
```diff
@@ -85,6 +85,10 @@ func defaultSupportedManagementPolicies() []sets.Set[xpv1.ManagementAction] {
 		// Like ObserveOnly, but the external resource is deleted when the
 		// managed resource is deleted.
 		sets.New[xpv1.ManagementAction](xpv1.ManagementActionObserve, xpv1.ManagementActionDelete),
+		// No Crate and no Delete. Just update/patch the external resource.
```
lsviben, line 88:
```suggestion
		// No Create and no Delete. Just update/patch the external resource.
```

lsviben (approved):
LGTM!

PR #698, opened 2024-05-29
Switch from Makefile to Earthly

negz:

We're trying out a new build tool for one release. See https://github.com/crossplane/crossplane/pull/5711 for context.

I have:

Just `earthly +reviewable` for now.


.github/workflows/promote.yml
```diff
@@ -1,48 +0,0 @@
-name: Promote
```
jbw976, line 1: this promote workflow was never used for crossplane-runtime, and therefore safe to delete?

.github/renovate.json5
```diff
@@ -1,96 +1,208 @@
...
-  "postUpdateOptions": ["gomodTidy"],
-// By default renovate will auto detect whether semantic commits have been used
-// in the recent history and comply with that, we explicitly disable it
-  "semanticCommits": "disabled",
-// All PRs should have a label
-  "labels": ["automated"],
-  "regexManagers": [
+  // The branches renovate should target
+  // PLEASE UPDATE THIS WHEN RELEASING.
+  "baseBranches": [
```
jbw976, line 15: oh interesting, we weren't specifying the set of `baseBranches` before now - consider making a small tweak to the instructions for updating `baseBranches` in https://github.com/crossplane/release/blob/main/.github/ISSUE_TEMPLATE/release.md, so we remember to update it here too in addition to core crossplane.

.github/workflows/renovate.yml
```diff
@@ -0,0 +1,57 @@
+name: Renovate
```
jbw976, line 1: wait, was Renovate not running on crossplane-runtime since switching to self hosted?

jbw976 (approved):
Took this for a quick test drive and all the commands I was running worked successfully! (`+build`, `+reviewable`, `+test`, `+generate`).  Looks like we've incorporated a lot of the lessons from iterating on core crossplane too :)

Just a few questions, nothing blocking on my end!

.github/workflows/promote.yml
negz: Yeah, exactly. I can't think of what it would have done.

.github/workflows/renovate.yml
negz: I think it was still using the Renovate app. See for example https://github.com/crossplane/crossplane-runtime/pull/695 which was opened last week.

I'm guessing we should switch to self hosted for the same reasons as c/c though.

PR #745, opened 2024-07-12
Disconnect() on per reconcile loop scoped external client

### LINKED ISSUE #697: Disconnect ExternalClient
## Problem
I am currently struggling with appropriately disconnecting the external client after it was used. 

My Reconciler is created with `WithExternalConnectDisconnecter`  [here](https://github.com/denniskniep/provider-temporal/blob/3d8cc85d971ddedb4661fc08a5b231efece41b60/internal/controller/searchattribute/searchattribute.go#L73)

I implemented the connect and disconnect method. But disconnect does not have any further specific parameters, so it seems to be intended to close the whole Connector. But there is only one Connector per Reconciler. Meaning if the Connectorś disconnect method is called and then the `Reconciler.Reconcile` gets another call it will call Connect() on an already Disconnected connector.

https://github.com/crossplane/crossplane-runtime/blob/fc036618ffd82023aeefe97554acdeb3a3db972f/pkg/reconciler/managed/reconciler.go#L892-L912

Currently the Disconnect call is:
`r.external.Disconnect(ctx);`

## Proposal

Wouldn´t  it be better to pass either `external` as a parameter to disconnect 
`r.external.Disconnect(ctx, external);`

or directly call Disconnect on external (without `r.`)
`external.Disconnect(ctx)`

Benefit is, that we can close precisely an reconcile call scoped client. And not disconnect the whole connector of the reconciler, though it is still in use and might run another reconcile

ravilr:

Fixes #697

This deprecates the `WithExternalConnectDisconnecter(..)` Reconciler Option, to replace the Disconnect(..) on ExternalConnecter interface added in https://github.com/crossplane/crossplane-runtime/pull/296 with the newly added Disconnect(..) on the ExternalClient interface. 

Any provider implementation has to implement the Disconnect(..) method on its external client type to satisfy the ExternalClient interface. An ExternalClient not requiring to explicitly disconnect to cleanup it resources, can provide a no-op implementation just returning nil.

I have: 

- ~[ ] Linked a PR or a [docs tracking issue] to [document this change].~
- ~[ ] Added `backport release-x.y` labels to auto-backport this PR.~

[contribution process]: https://github.com/crossplane/crossplane/tree/master/contributing
[cheat sheet]: https://github.com/crossplane/crossplane/tree/master/contributing#checklist-cheat-sheet

### DIFF
--- a/pkg/reconciler/managed/reconciler.go
+++ b/pkg/reconciler/managed/reconciler.go
@@ -230,6 +230,9 @@ type ExternalConnecter interface {
 }
 
 // An ExternalDisconnecter disconnects from a provider.
+//
+// Deprecated: Please use Disconnect() on the ExternalClient for disconnecting
+// from the provider.
 type ExternalDisconnecter interface {
 	// Disconnect from the provider and close the ExternalClient.
 	Disconnect(ctx context.Context) error
@@ -330,15 +333,22 @@ type ExternalClient interface {
 	// Delete the external resource upon deletion of its associated Managed
 	// resource. Called when the managed resource has been deleted.
 	Delete(ctx context.Context, mg resource.Managed) error
+
+	// Disconnect from the provider and close the ExternalClient.
+	// Called at the end of reconcile loop. An ExternalClient not requiring
+	// to explicitly disconnect to cleanup it resources, can provide a no-op
+	// implementation which just return nil.
+	Disconnect(ctx context.Context) error
 }
 
 // ExternalClientFns are a series of functions that satisfy the ExternalClient
 // interface.
 type ExternalClientFns struct {
-	ObserveFn func(ctx context.Context, mg resource.Managed) (ExternalObservation, error)
-	CreateFn  func(ctx context.Context, mg resource.Managed) (ExternalCreation, error)
-	UpdateFn  func(ctx context.Context, mg resource.Managed) (ExternalUpdate, error)
-	DeleteFn  func(ctx context.Context, mg resource.Managed) error
+	ObserveFn    func(ctx context.Context, mg resource.Managed) (ExternalObservation, error)
+	CreateFn     func(ctx context.Context, mg resource.Managed) (ExternalCreation, error)
+	UpdateFn     func(ctx context.Context, mg resource.Managed) (ExternalUpdate, error)
+	DeleteFn     func(ctx context.Context, mg resource.Managed) error
+	DisconnectFn func(ctx context.Context) error
 }
 
 // Observe the external resource the supplied Managed resource represents, if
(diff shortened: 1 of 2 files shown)


ravilr:
@negz @denniskniep PTAL.

bobh66:
@ravilr I may not have a complete understanding but I'm curious why this affects `DIsconnect()` but not `Connect()`?  It seems like if the `Disconnect()` method is applicable to a particular scope then there should also be a similarly scoped `Connect()` method?

I'm happy to be wrong, I just want to make sure we're not missing anything.  Thanks

bobh66:
> @ravilr I may not have a complete understanding but I'm curious why this affects `DIsconnect()` but not `Connect()`? It seems like if the `Disconnect()` method is applicable to a particular scope then there should also be a similarly scoped `Connect()` method?

Hm - maybe the `Connect` is implicit in the creation of the `ExternalClient`.

ravilr:
> Hm - maybe the Connect is implicit in the creation of the ExternalClient.

yes, on every reconcile loop, [ExternalConnecter's Connect()](https://github.com/crossplane/crossplane-runtime/blob/v1.16.0/pkg/reconciler/managed/reconciler.go#L229) produces a new ExternalClient, and we want to be able to disconnect that ExternalClient at the end of that reconcile, if in case such an ExternalClient needs calling a close method to release resources it is holding (for example, a GRPC client requiring calling connection close).

Also, linking the crossplane slack channel thread in #core-development:  https://crossplane.slack.com/archives/CEF5N8X08/p1720210518036179, for more context.

negz:
This LGTM, thanks @ravilr!

> Any provider implementation has to implement the Disconnect(..) method on its external client type to satisfy the ExternalClient interface.

It'd be nice to avoid this but I can't think of any simple way to do so.

Could you update the comment on the new `Disconnect` method to mention that clients that don't need to explicitly disconnect can just `return nil` without doing anything (or similar wording). I want to make sure folks don't get confused, thinking they must implement `Disconnect`.

ravilr:
> Could you update the comment on the new `Disconnect` method to mention that clients that don't need to explicitly disconnect can just `return nil` without doing anything (or similar wording). I want to make sure folks don't get confused, thinking they must implement `Disconnect`.

Updated go doc comments to add :

```
        // Disconnect from the provider and close the ExternalClient.
	// Called at the end of reconcile loop. An ExternalClient not requiring
	// to explicitly disconnect to cleanup it resources, can just provide a no-op
	// implementation returning nil.
```

ravilr:
@bobh66 @negz  is this good to be merged ?

denniskniep:
@ravilr LGTM
Thanks!

negz (approved):

PR #754, opened 2024-08-03
Change logs support to track all changes being made to managed resources

jbw976:

This PR contains the `crossplane-runtime` changes to support cloud change logs.

* Issue: https://github.com/crossplane/crossplane/issues/5477
* Design: https://github.com/crossplane/crossplane/pull/5822

Further PRs for changes to the providers themselves (to consume these runtime changes) and for the change logs sidecar container are coming soon, but should not block this PR from being merged. Previews of those changes can be found in:

* https://github.com/crossplane-contrib/provider-kubernetes/compare/main...jbw976:provider-kubernetes:change-logs
* https://github.com/jbw976/change-log-sidecar 

### Testing

I have prepared some testing instructions at https://github.com/jbw976/change-log-sidecar/blob/main/README.md that reviewers can try out fairly easily on their local laptops.

Here is some example output that demonstrates what a subset of fields of a change log entry looks like, for example when an object is created, updated, then deleted:
```
❯ kubectl -n crossplane-system logs -l pkg.crossplane.io/provider=provider-kubernetes -c change-log-sidecar | jq '.timestamp + " " + .provider + " " + .name + " " + .operation'
"2024-08-09T04:48:43Z provider-kubernetes:v0.0.6 object-0 OPERATION_TYPE_CREATE"
"2024-08-09T04:49:22Z provider-kubernetes:v0.0.6 object-0 OPERATION_TYPE_UPDATE"
"2024-08-09T04:50:32Z provider-kubernetes:v0.0.6 object-0 OPERATION_TYPE_DELETE"
```

A full change log entry looks like the below:
```
{"apiVersion":"kubernetes.crossplane.io/v1alpha2","externalName":"object-0","kind":"Object","name":"object-0","operation":"OPERATION_TYPE_CREATE","provider":"provider-kubernetes:v0.0.6","snapshot":{"apiVersion":"kubernetes.crossplane.io/v1alpha2","kind":"Object","metadata":{"annotations":{"crossplane.io/composition-resource-name":"resource-object-0","crossplane.io/external-name":"object-0"},"creationTimestamp":"2024-08-09T04:48:43Z","generateName":"traceperf-tester-wnw9g-","generation":2,"labels":{"crossplane.io/claim-name":"traceperf-tester","crossplane.io/claim-namespace":"default","crossplane.io/composite":"traceperf-tester-wnw9g"},"managedFields":[{"apiVersion":"kubernetes.crossplane.io/v1alpha2","fieldsType":"FieldsV1","fieldsV1":{"f:metadata":{"f:annotations":{"f:crossplane.io/composition-resource-name":{}},"f:generateName":{},"f:labels":{"f:crossplane.io/claim-name":{},"f:crossplane.io/claim-namespace":{},"f:crossplane.io/composite":{}},"f:ownerReferences":{"k:{\"uid\":\"c21ef3e7-adb8-4b1f-9b10-b7fbf33bf623\"}":{}}},"f:spec":{"f:forProvider":{"f:manifest":{"f:apiVersion":{},"f:data":{".":{},"f:key":{}},"f:kind":{},"f:metadata":{"f:namespace":{}}}}}},"manager":"apiextensions.crossplane.io/composed/010d7ea543f4d932601bc4472c6b11a93716eecdfa070bfdcea16def1d7efd74","operation":"Apply","time":"2024-08-09T04:48:43Z"},{"apiVersion":"kubernetes.crossplane.io/v1alpha2","fieldsType":"FieldsV1","fieldsV1":{"f:metadata":{"f:annotations":{"f:crossplane.io/external-name":{}}},"f:spec":{"f:readiness":{".":{},"f:policy":{}}}},"manager":"crossplane-kubernetes-provider","operation":"Update","time":"2024-08-09T04:48:43Z"}],"name":"object-0","ownerReferences":[{"apiVersion":"trace-perf.crossplane.io/v1alpha1","blockOwnerDeletion":true,"controller":true,"kind":"XTracePerf","name":"traceperf-tester-wnw9g","uid":"c21ef3e7-adb8-4b1f-9b10-b7fbf33bf623"}],"resourceVersion":"998","uid":"905af6bb-eb03-4569-9fe1-b257fa8019e9"},"spec":{"deletionPolicy":"Delete","forProvider":{"manifest":{"apiVersion":"v1","data":{"key":"value-0-tester-foo"},"kind":"ConfigMap","metadata":{"namespace":"default"}}},"managementPolicies":["*"],"providerConfigRef":{"name":"default"},"readiness":{"policy":"SuccessfulCreate"}},"status":{"atProvider":{"manifest":null}}},"timestamp":"2024-08-09T04:48:43Z"}
```

I have: 

[contribution process]: https://github.com/crossplane/crossplane/tree/master/contributing
[cheat sheet]: https://github.com/crossplane/crossplane/tree/master/contributing#checklist-cheat-sheet


apis/proto/v1alpha1/changelog.proto
```diff
@@ -0,0 +1,81 @@
...
+
+// ChangeLogResponse is the response returned by the ChangeLogService after a
+// change log entry is sent.
+message ChangeLogResponse {
+  // A boolean indicating whether the change log entry was successfully sent.
+  bool success = 1;
+
+  // A message that provides further information about the result of sending the
+  // change log entry.
+  string message = 2;
```
jbw976, line 80: These fields on the `ChangeLogResponse` don't feel particularly useful right now, i.e. not sure if we need to return anything at all for successful sends. Is it possible that the only useful information the server side may need to return (e.g. something has gone wrong) is something that can just be returned in an `error` instead of a specific response type?

pkg/reconciler/managed/reconciler.go
```diff
@@ -1248,3 +1291,92 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (resu
...
+			if err = ads.UnmarshalJSON(b); err != nil {
+				log.Debug("Cannot convert additional details to Struct", "error", err)
+				record.Event(managed, event.Warning(reasonCannotSendChangeLog, err))
+				ads = nil
+			}
+		}
+	}
+
+	// determine the full namespaced name of the managed resource
+	var namespacedName string
```
sttts, line 1326: `namespacedName := types.NamespacedName{Namespace: managed.GetNamespace(), Name: managed.GetName()}.String()`

pkg/reconciler/managed/reconciler.go
```diff
@@ -1248,3 +1291,92 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (resu
...
+	if !r.features.Enabled(feature.EnableAlphaChangeLogs) {
+		// change log feature isn't enabled, just return immediately
+		return nil
+	}
+
+	// capture the full state of the managed resource from before we performed the change
+	snapshot, err := resource.AsStruct(managed)
+	if err != nil {
+		log.Debug("Cannot snapshot managed resource", "error", err)
+		record.Event(managed, event.Warning(reasonCannotSendChangeLog, err))
```
sttts, line 1377: Are events the right vehicle to report such an error?

pkg/reconciler/managed/reconciler.go
```diff
@@ -1248,3 +1291,92 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (resu
...
+		Type:              managed.GetObjectKind().GroupVersionKind().String(),
+		Name:              namespacedName,
+		ExternalName:      meta.GetExternalName(managed),
+		Operation:         opType,
+		Snapshot:          snapshot,
+		ErrorMessage:      changeErrMessage,
+		AdditionalDetails: ads,
+	}
+
+	clr, err := r.changeLogClient.SendChangeLog(ctx, cle)
```
jbw976, line 1352: I'm considering adding the `WaitForReady()` call option here, so we automatically retry calls that fail due to a transient failure state. Some resources about using `WaitForReady()`:

* https://pkg.go.dev/google.golang.org/grpc#WaitForReady
* https://grpc.io/docs/guides/wait-for-ready/
* https://github.com/grpc/grpc-go/blob/master/Documentation/anti-patterns.md

> By default, RPCs will fail if the ClientConn enters the "transient failure" state, but setting WaitForReady(true) on a call will cause it to queue even in the "transient failure" state, and it will only ever fail due to a deadline, a server response, or a connection loss after the RPC was sent to a server.

If we were to use `WaitForReady()`, we should probably have a new `context` here that is separate from the main `Reconcile()` context and has a shorter timeout/deadline.

I was curious what Crossplane does for the function pipeline gRPC calls:

* the main function pipeline looks to use a [round robin load balancer](https://github.com/crossplane/crossplane/blob/master/internal/xfn/function_runner.go#L229-L232) approach for all connections as part of the default service config
* the `render` CLI command appears to use [`waitForReady`](https://github.com/crossplane/crossplane/blob/master/cmd/crank/beta/render/render.go#L110-L112) for all connections as part of the default service config

I think it could make these calls more robust for dealing with transient failures if we are using the `WaitForReady()` call option with a reasonably short context deadline. any strong opinions here?

pkg/reconciler/managed/reconciler.go
jbw976: yep yep, I actually tried that first too, but didn't like the output for when there is no namespace. [`NamespacedName.String()`](https://github.com/kubernetes/apimachinery/blob/master/pkg/types/namespacedname.go#L38 ) always uses a separator char even when there is no namespace, so the output looks like `/my-resource`.

I was doubting even including the namespace at all because managed resources aren't namespaced right now in Crossplane v1.  Perhaps that will change for a v2, so I may have been getting ahead of myself with future-proofing by including handling namespaces already at this point 😅 

I could just remove this entirely and include only the name of the resource, without any consideration for namespace.

pkg/reconciler/managed/reconciler.go
jbw976: Great question @sttts! I'd love to hear your opinion if you have one 🙇‍♂️ 

I was thinking that the core reconciliation of the managed resource is the most important thing, and if the change log fails to send then I'd still like reconciliation to complete. A (somewhat) visible way to note that an error happened, but to not alter the rest of the reconciliation attempt, was to just surface it as a warning event on the resource.

Do you have an opinion or other ideas?

sttts: Fearing lots of such events when something goes wrong with the audit log collector.

sttts: Alternative: log error and emit metrics.

jbw976:
Pushed https://github.com/crossplane/crossplane-runtime/pull/754/commits/2dd945877040682f34ca708230f5283634013c20 to address all the protobuf linter errors found in https://github.com/crossplane/crossplane-runtime/actions/runs/10229176882/job/28302502501?pr=754.  

Looks like all those are green now in https://github.com/crossplane/crossplane-runtime/actions/runs/10257477240/job/28378586893?pr=754 🟢

apis/changelogs/proto/v1alpha1/changelog.proto
```diff
@@ -0,0 +1,81 @@
...
+
+// SendChangeLogResponse is the response returned by the ChangeLogService after a
+// change log entry is sent.
+message SendChangeLogResponse {
+  // A boolean indicating whether the change log entry was successfully sent.
+  bool success = 1;
+
+  // A message that provides further information about the result of sending the
+  // change log entry.
+  string message = 2;
```
negz, line 80: I think gRPC has built in support to return this level of detail as part of the actual error.

https://grpc.io/docs/guides/error/
https://github.com/grpc/grpc-go/blob/e524655/examples/features/error_details/server/main.go#L54

I believe it'd be idiomatic gRPC to have the ChangeLogService just return an empty message if everything worked okay, and return an error if there was one. You could return the [well-known Empty message](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/empty.proto#L43), but probably defining SendChangeLogResponse as an empty message would be more future proof. That way we could add response fields in future if we ever realized we needed them.

pkg/controller/options.go
```diff
@@ -97,3 +101,10 @@ type MetricOptions struct {
 	// MRStateMetrics to use for recording state metrics.
 	MRStateMetrics *statemetrics.MRStateMetrics
 }
+
+// ChangeLogOptions for recording changes to managed resources into the change
+// logs.
+type ChangeLogOptions struct {
+	ChangeLogClient changelogs.ChangeLogServiceClient
+	ProviderVersion string
```
negz, line 109: Is this really a changelog-specific option? I wonder if anything else might need access to the provider version in future.

pkg/reconciler/managed/reconciler.go
```diff
@@ -138,6 +144,11 @@ func (fn CriticalAnnotationUpdateFn) UpdateCriticalAnnotations(ctx context.Conte
 // resource, for example usernames, passwords, endpoints, ports, etc.
 type ConnectionDetails map[string][]byte
 
+// AdditionalDetails represent any additional details the external client wants
+// to return about an operation that has been performed. These details will be
+// included in the change logs.
+type AdditionalDetails map[string]any
```
negz, line 150: Should the values of this map really be as open-ended as `any`? Could we get away with requiring string values?

pkg/reconciler/managed/reconciler.go
```diff
@@ -466,6 +477,10 @@ type ExternalCreation struct {
 	// unless an existing key is overwritten. Crossplane may publish these
 	// credentials to a store (e.g. a Secret).
 	ConnectionDetails ConnectionDetails
+
+	// AdditionalDetails represent any additional details the external client
+	// wants to return about the creation operation that was performed.
+	AdditionalDetails *AdditionalDetails
```
negz, line 483: Given this is a map type I'd avoid the pointer. The zero value of a map is nil anyway.

pkg/reconciler/managed/reconciler.go
```diff
@@ -700,6 +721,16 @@ func WithReconcilerSupportedManagementPolicies(supported []sets.Set[xpv1.Managem
 	}
 }
 
+// WithChangeLogs enables support for capturing change logs during
+// reconciliation.
+func WithChangeLogs(c changelogs.ChangeLogServiceClient, pv string) ReconcilerOption {
```
negz, line 726:
```suggestion
func WithChangeLogs(c changelogs.ChangeLogServiceClient, providerVersion string) ReconcilerOption {
```

I'm usually the first to argue for short variable names in Go, but in cases like this where it's not obvious from the function name or type, I think more descriptive names are valuable.

pkg/resource/resource.go
```diff
@@ -425,3 +433,32 @@ func StableNAndSomeMore(n int, names []string) string {
 	sort.Strings(cpy)
 	return FirstNAndSomeMore(n, cpy)
 }
+
+// AsStruct converts the given object to a structpb.Struct for usage with gRPC
+// connections.
+// Copied from Crossplane's internal composite controller package
+// "github.com/crossplane/crossplane/internal/controller/apiextensions/composite"
+// Source code at:
```
negz, line 441:
```suggestion
// Copied from:
```

Nit: This is probably sufficient context.

pkg/reconciler/managed/reconciler.go
```diff
@@ -1248,3 +1291,92 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (resu
...
 }
+
+// recordChangeLog constructs a change log entry from the provided information
+// and sends it to the change log service, if the feature is enabled. If errors
+// are encountered during this process, then logs/events will be recorded for
+// each, but the process will continue. We are choosing to send all the
+// information we can to the change logs as opposed to sending nothing.
+func (r *Reconciler) recordChangeLog(ctx context.Context, managed resource.Managed, snapshot *structpb.Struct, log logging.Logger,
+	record event.Recorder, opType changelogs.OperationType, changeErr error, ad *AdditionalDetails,
+) {
```
negz, line 1302: This function signature is pretty long for my comfort.

https://github.com/crossplane/crossplane/tree/master/contributing#dont-wrap-function-signatures

I wonder, does this need to be a method on `Reconciler`? It seems to have most of its dependencies passed in as arguments. I suggest making it a self-contained injectable type that wraps the 'raw' `changelogs.ChangeLogServiceClient` gRPC generates.

Something like this:

```go
type ChangeLogger struct {
  client changelings.ChangeLogServiceClient
}

func (c *ChangeLogger) RecordChangeLog(ctx context.Context, mg resource.Managed, cl *changelogs.ChangeLog) error {
  // Logic goes here
}
```

I would then:

* Not pass a logger and event recorder to this type. Instead, return descriptive errors up to the `Reconciler` and let that log the errors and/or create events.
* Define a `ChangeLog` message at the Protobuf level, so that snapshot, opType, and ad all become part of one message you can pass into `RecordChangeLog`.

apis/changelogs/proto/v1alpha1/changelog.proto
```diff
@@ -0,0 +1,81 @@
...
+// ChangeLogService is a service that provides the ability to send change log
+// entries.
+service ChangeLogService {
+  // SendChangeLog sends a change log entry to the change log service.
+  rpc SendChangeLog (SendChangeLogRequest) returns (SendChangeLogResponse) {}
+}
+
+// SendChangeLogRequest represents a single change log entry, with detailed
+// information about the resource that was changed.
+message SendChangeLogRequest {
```
negz, line 32: I'd consider taking everything that constitutes "a change log" and making it a distinct message (e.g. `message ChangeLog`) that is included in a `SendChangeLogRequest`.

I suspect right now that means we'll just have:

```protobuf
message SendChangeLogRequest {
  ChangeLog log = 1;
}

message ChangeLog {
  // All the fields that are currently in SendChangeLogRequest
}
```

The reason I suggest this is:

* If in future we want to add fields that pertain to the request to send a change log but that aren't _actually part of the change log_ (e.g. request metadata) there'll be a more obvious distinction.
* It seems like it might give you a generated struct that's more appropriate to use as a "container type" for all the things that constitute a change log in the Reconciler code. (See related suggestion in that code.)

apis/changelogs/proto/v1alpha1/changelog.proto
```diff
@@ -0,0 +1,81 @@
...
+package changelogs.proto.v1alpha1;
+
+option go_package = "github.com/crossplane/crossplane-runtime/apis/changelogs/proto/v1alpha1";
+
+// ChangeLogService is a service that provides the ability to send change log
+// entries.
+service ChangeLogService {
+  // SendChangeLog sends a change log entry to the change log service.
+  rpc SendChangeLog (SendChangeLogRequest) returns (SendChangeLogResponse) {}
+}
```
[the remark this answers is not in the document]
jbw976: Interesting idea - I can see it making sense to minimize the number of calls/responses to the server in the change logs sidecar.  However, I'm having trouble understanding the lifecycle/management of the calls on the client side in the managed reconciler. Right now, the unary calls are directly tied to a single reconciliation - the call/response is complete within a single reconcile.

In a streaming approach, when does the client initiate the stream and then when would it decide to call `CloseAndRecv` to get the response from the server for the batch? 

The unary approach has a clear response within a single reconciliation, so we can easily attach it to the given resource before it finishes being reconciled. I'm not sure how to tie a response to a given reconcile in the streaming approach where we may not get a response until long after the reconcile is finished.

I'm sure we could think through how to manage the client side stream and getting responses reported appropriately for their corresponding resources, but do you feel like that extra complexity is worth it? Maybe it's more simple than I'm thinking 😇

pkg/reconciler/managed/reconciler.go
jbw976: I was thinking the flexibility would be nice for externalClients to be able to return JSON-y type of details (e.g. `map[string]object`), but I'm not a full 100% on that.

What's the biggest benefit you think we'd get from imposing a more constrained type of `map[string]string`?

negz: It's simpler to reason about, and eliminates any issues with folks trying to return values that can't easily be serialized. Unless there's a clear reason to support arbitrary/complex values I'd keep it simple.

jbw976: Cool! that's reasonable and I can make that change. The more PRs I review, the more I appreciate things being simple to reason about 😅

apis/changelogs/proto/v1alpha1/changelog.proto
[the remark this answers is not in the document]
negz: I've never used streaming gRPC before, but it my thinking is that you'd create the `stream` (i.e. the client with `Send` and `CloseAndRecv` methods) at provider startup. You'd then call `CloseAndRecv` only when the provider shuts down (so basically never).

You'd call the `Send` method on each reconcile to send a change log. This method returns an error, so presumably you'll get an error if the change log fails to send and can record that.

Given that AFAICT we don't really have a response for the server (sidecar) to send the client other than "yep I got your change log" I don't think it really matters if we never call `CloseAndRecv`.

This is all just based on a few minutes reading the gRPC streaming docs though, so 🤷‍♂️.

apis/changelogs/proto/v1alpha1/changelog.proto
jbw976: Great idea, with https://github.com/crossplane/crossplane-runtime/pull/754/commits/afb1c6b1148b69d12412523e3133266c8c810ea2, the response is now empty and any information being returned with the response is through errors.

Server side error handling can be seen in https://github.com/jbw976/change-log-sidecar/blob/main/server/server.go.

pkg/controller/options.go
jbw976: Elevated this provider version controller option to a higher level in https://github.com/crossplane/crossplane-runtime/pull/754/commits/afb1c6b1148b69d12412523e3133266c8c810ea2

pkg/reconciler/managed/reconciler.go
jbw976: Done! Additional details are now constrained to `map[string]string` in https://github.com/crossplane/crossplane-runtime/pull/754/commits/afb1c6b1148b69d12412523e3133266c8c810ea2.

Note I also realized that `Delete()` didn't have the ability to specify additional details because it currently has no return type - it can only return errors. I also updated that with https://github.com/crossplane/crossplane-runtime/pull/754/commits/afb1c6b1148b69d12412523e3133266c8c810ea2#diff-97300a2543f95f5a2ada3560bf47dd7334e237e27976574d15d1cddef2e66c01R345, but let me know if you'd like to skip that because it does force consumers of the managed reconciler to update to this new `Delete()` signature whenever they update their crossplane-runtime dependency next, a la https://github.com/jbw976/provider-kubernetes/commit/4a9146acc2b2444517ccc66f43cb3a2aa650a2e1#diff-475eea76845069ccefe80a5e3e03cf95a235099281bb9bcc874d19f243ed8e2c.

apis/changelogs/proto/v1alpha1/changelog.proto
jbw976: I have made this change to capture all the change log entry specific details in its own message type, which is the included in the Request type.  If we need to add new fields specific to the request itself in the future, we now have a clean separation for that. Good idea! 

https://github.com/crossplane/crossplane-runtime/pull/754/commits/afb1c6b1148b69d12412523e3133266c8c810ea2#diff-0da98c4e35025a6b55e0791d1d95743104fe96e9bb3d0466e27deda37120ee24

pkg/reconciler/managed/reconciler.go
jbw976: I've made a bunch of data format changes to the protobuf types as suggested, and I'm starting to look into this now.  I also already thought the `recordChangeLog()` method signature was overly long, so I'm not against exploring this suggestion too.

However, one thing I liked about my first implementation was that everything was encapsulated in the `recordChangeLog` method, including change log entry preparation, error handling, and event recording. This allowed the `Reconcile()` function, which is pretty complex and long, to just get a single line addition where needed to call `recordChangeLog`. With the approach you suggest, I believe we'll have additional lines within `Reconcile()` to prep the change log entry with data and to handle errors / record events in all the places we call it (create/update/delete - success/failure).

That's the downside I see to this approach, but it may end up being palatable.  I will play with it and see! 🤓

apis/changelogs/proto/v1alpha1/changelog.proto
```diff
@@ -0,0 +1,81 @@
...
+
+// SendChangeLogRequest represents a request to send a single change log entry.
+message SendChangeLogRequest {
+  // The change log entry to send as part of this request.
+  ChangeLogEntry entry = 1;
+}
+
+// ChangeLog represents a single change log entry, with detailed information
+// about the resource that was changed.
+message ChangeLogEntry {
```
negz, line 38: Nit: Looks like the type is ChangeLogEntry but the comment says just ChangeLog.

apis/changelogs/proto/v1alpha1/changelog.proto
```diff
@@ -0,0 +1,81 @@
...
+  // The name and version of the provider that is making the change to the
+  // resource.
+  string provider = 1;
+
+  // The type of the resource that was changed, e.g. Group, Version, Kind.
+  string type = 2;
```
negz, line 44: Would it make sense to have distinct `apiVersion` and `kind` fields? This might make formatting clearer.

jbw976: You know, I've been recently thinking more and more of doing that myself. The current `type` field looks like this in the change logs:
```
"type":"kubernetes.crossplane.io/v1alpha2, Kind=Object"
```

I can see it being better for formatting, but also maybe better for indexing/searching to separate those into something like:
```
"apiVersion":"kubernetes.crossplane.io/v1alpha2","kind":"Object"
```

jbw976: Done in https://github.com/crossplane/crossplane-runtime/pull/754/commits/874a4a4f39c560a7ea64648d3154807e665857ec

pkg/reconciler/managed/reconciler.go
jbw976: Took a quick stab at this with https://github.com/crossplane/crossplane-runtime/pull/754/commits/ae4de71d3d844f1697530e4af8a40810ce09ea41. The change logs functionality has been broken out from the managed reconciler into a `changeLogger` type with a few key methods:

* `takeSnapshot()` to capture the state of the managed resource early during reconciliation before any operations are performed
* `newChangeLogEntry()` to create and prep a change log entry with given values/details
* `recordChangeLog()` to actually send the change log entry to the underlying change log service client

Note that the error handling and event recording is being done inside this `changeLogger`, so the main `Reconcile()` doesn't get many additional lines added. All that `Reconcile()` needs to have added is basically `newChangeLogEntry(a, b, c)` followed by `recordChangeLog(entry)`.

I think I like keeping changes to `Reconcile()` very minimal, but still playing with this idea a bit 😇

pkg/reconciler/managed/changelogger.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/crossplane/crossplane-runtime/pkg/meta"
+	"github.com/crossplane/crossplane-runtime/pkg/reference"
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
+	"google.golang.org/protobuf/types/known/structpb"
+)
+
+// changeLogger processes changes to resources and helps to send them to the
+// change log service
+type changeLogger struct {
+	enabled         bool
```
negz, line 35: Rather than `enabled bool` I'd suggest:

* Define a ChangeLogger interface
* Have a NopChangeLogger implementation that does nothing when called, and use that by default
* Add a WithChangeLogger option for the Reconciler
* Use WithChangeLogger to pass in this real implementation if the feature flag is enabled

pkg/reconciler/managed/changelogger.go
```diff
@@ -0,0 +1,129 @@
...
+	client          changelogs.ChangeLogServiceClient
+	providerVersion string
+	log             logging.Logger
+	record          event.Recorder
```
negz, line 39: This is maybe controversial, but I generally prefer to keep logs and events emitted by the code that orchestrates everything else - i.e. the reconcile loop. Especially if I'm just logging/eventing for errors, I think it's better to write dependencies (like this change logger) as more of a library (without loggers etc) and have the caller (the Reconciler) log the errors it returns. This way you get one part of the codebase that's responsible for all logging/eventing etc.

Generally the only exception I'll make to this is if a dependency (like the change logger) has something really interesting/useful to log that _isn't_ an error.

pkg/reconciler/managed/reconciler.go
```diff
@@ -347,7 +354,7 @@ type ExternalClientFns struct {
 	ObserveFn    func(ctx context.Context, mg resource.Managed) (ExternalObservation, error)
 	CreateFn     func(ctx context.Context, mg resource.Managed) (ExternalCreation, error)
 	UpdateFn     func(ctx context.Context, mg resource.Managed) (ExternalUpdate, error)
-	DeleteFn     func(ctx context.Context, mg resource.Managed) error
+	DeleteFn     func(ctx context.Context, mg resource.Managed) (ExternalDelete, error)
```
negz, line 357: Changing this return signature will be a breaking change for all provider implementations.

pkg/reconciler/managed/reconciler.go
```diff
@@ -29,6 +29,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
+	changelogs "github.com/crossplane/crossplane-runtime/apis/changelogs/proto/v1alpha1"
```
negz, line 32:
```suggestion
	"github.com/crossplane/crossplane-runtime/apis/changelogs/proto/v1alpha1"
```

I think we normally import generated proto types in a similar way to Kubernetes types. The convention is that if it's a Crossplane type (as opposed to upstream Kubernetes or third party) we just import it using its version - so `v1alpha1`. If we end up with two different `v1alpha` packages then we go to e.g. `changelogsv1alpha1` to differentiate.

pkg/reconciler/managed/reconciler.go
```diff
@@ -972,18 +1008,26 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (resu
 		return reconcile.Result{Requeue: true}, nil
 	}
 
+	// take a snapshot of the managed resource now that we've called Observe()
+	// and have not performed any external operations
+	changeLogger := newChangeLogger(r.features.Enabled(feature.EnableAlphaChangeLogs), r.changeLogClient, log, record, r.providerVersion)
+	changeLogger.takeSnapshot(managed)
```
negz, line 1014: I'd pass in the whole change logger using the `WithChangeLogs` option.

pkg/reconciler/managed/changelogger.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	// determine the full namespaced name of the managed resource
+	var namespacedName string
+	namespace := managed.GetNamespace()
+	if namespace != "" {
+		namespacedName = fmt.Sprintf("%s/%s", namespace, managed.GetName())
+	} else {
+		namespacedName = managed.GetName()
+	}
```
negz, line 91: I think it's safe to assume managed resources aren't namespaced.

pkg/reconciler/managed/changelogger.go
```diff
@@ -0,0 +1,129 @@
...
+		record:          record,
+	}
+}
+
+// takeSnapshot attempts to take a snapshot of the given managed resource by
+// converting it to a Struct that will be sent to the change logs after an
+// external operation is performed, if the change logs feature is enabled. If
+// there is an error during this process, an event will be recorded and no
+// snapshot will be saved.
+func (c *changeLogger) takeSnapshot(managed resource.Managed) {
```
negz, line 59: I wonder whether this type really needs three methods (snapshot, new entry, record entry).

If I was designing this I'd probably:

1. Take a deep copy of `managed` where you currently take the snapshot
2. Pass the deep copy, opType, error and additional details to `recordChangeLog` and have it form the entry

pkg/reconciler/managed/reconciler.go
jbw976: Namespace usage removed in https://github.com/crossplane/crossplane-runtime/pull/754/commits/c23f09b8314e5a7d11e0438243572a0b38cfaff1 as per https://github.com/crossplane/crossplane-runtime/pull/754#discussion_r1707711834.

pkg/reconciler/managed/reconciler.go
jbw976: Addressed in a more cleanly designed way as discussed in https://github.com/crossplane/crossplane-runtime/pull/754/commits/c23f09b8314e5a7d11e0438243572a0b38cfaff1

pkg/reconciler/managed/changelogger.go
jbw976: This approach of a `ChangeLogger` interface and `GRPCChangeLogger`/`NopChangeLogger` implementations is done in https://github.com/crossplane/crossplane-runtime/pull/754/commits/c23f09b8314e5a7d11e0438243572a0b38cfaff1.

pkg/reconciler/managed/changelogger.go
jbw976: Cool cool!! more than happy to follow the existing patterns in the `Reconcile()` code base. Now with the `ChangeLogger` interface in https://github.com/crossplane/crossplane-runtime/pull/754/commits/c23f09b8314e5a7d11e0438243572a0b38cfaff1, an error is returned from `RecordChangeLog()` and `Reconcile()` handles that with debug logging and recording a warning event.

pkg/reconciler/managed/reconciler.go
jbw976: Yes, when providers update to the v1.17 version of crossplane-runtime, they will need to update their `ExternalClient.Delete` implementations to now return an `ExternalDelete` struct, which will very likely be empty if they have no additional details to provide about their delete operation.  Example from provider-kubernetes: https://github.com/jbw976/provider-kubernetes/commit/4a9146acc2b2444517ccc66f43cb3a2aa650a2e1#diff-475eea76845069ccefe80a5e3e03cf95a235099281bb9bcc874d19f243ed8e2c 

As discussed, you were on the fence and I'm erring on the side of consistency for all 3 CUD operations.

pkg/reconciler/managed/changelogger.go
jbw976: Yep, I like this approach as done in https://github.com/crossplane/crossplane-runtime/pull/754/commits/c23f09b8314e5a7d11e0438243572a0b38cfaff1

apis/changelogs/proto/v1alpha1/changelog.proto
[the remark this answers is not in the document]
jbw976: I explored a client side streaming approach and I concluded that it's not a good fit for our scenario. Some artifacts of my exploration (please excuse the sloppy prototyping code 😇):

* https://github.com/jbw976/crossplane-runtime/commit/3754f69beee135226c74915e085bce84afbe81e3
* https://github.com/jbw976/provider-kubernetes/commit/29b4cc59fc9e3d6d5c12be6e57968ee72f7a7b83
* https://github.com/jbw976/change-log-sidecar/commit/59d09869bf5351a924d54e1388a10f31af0b9a76

Some observations from the prototype:

* I tried two different approaches for the client (reconciler) handing errors/messages from the server (change logs sidecar)
  * client only calls `Send()`: A server side internal processing error, as opposed to a transport/networking error, is not returned to the client from `stream.Send()`. A non transport error can only be received at the very end when the client finally calls `stream.CloseAndRecv()` and it only gets the first error the server encountered.
  * client calls `RecvMsg()` after `Send()`: The first time the client calls `RecvMsg()` it will get a full response and good server side internal error details back OK, but this closes the stream. Any subsequent attempt by the client to call `Send()` again will get `EOF`.
* bi-directional streaming could maybe work but that seems to be intended more for asynchronous sends/receives on different channels. IMHO we have a need for a classic call/response, where a single entry is sent and we get back an immediate single response. Just sticking with unary seems reasonable to me here 🤓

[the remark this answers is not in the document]
negz: Got it, thanks for investigating! Sounds like unary is the best approach here after all.

pkg/reconciler/managed/changelogger.go
```diff
@@ -0,0 +1,94 @@
...
+		providerVersion: providerVersion,
+	}
+}
+
+// RecordChangeLog sends the given change log entry to the change log service.
+func (g *GRPCChangeLogger) RecordChangeLog(ctx context.Context, managed resource.Managed, opType v1alpha1.OperationType, changeErr error, ad AdditionalDetails) error {
+	// get an error message from the error if it exists
+	var changeErrMessage *string
+	if changeErr != nil {
+		changeErrMessage = reference.ToPtrValue(changeErr.Error())
```
negz, line 56: Huh, I actually forgot we had these pointer helpers in runtime.

We usually use https://pkg.go.dev/k8s.io/utils/ptr instead these days. I'd suggest that. We should probably deprecate the ones we have in runtime, but that can happen in a separate PR.

pkg/reconciler/managed/changelogger.go
```diff
@@ -0,0 +1,94 @@
...
+		ExternalName:      meta.GetExternalName(managed),
+		Operation:         opType,
+		Snapshot:          snapshot,
+		ErrorMessage:      changeErrMessage,
+		AdditionalDetails: ad,
+	}
+
+	// send everything we've got to the change log service
+	_, err = g.client.SendChangeLog(ctx, &v1alpha1.SendChangeLogRequest{Entry: entry})
+	return errors.Wrap(err, "Cannot send change log entry")
```
negz, line 81:
```suggestion
	return errors.Wrap(err, "cannot send change log entry")
```

Nit: Go recommends not capitalizing error strings, since they'll usually be concatenated together. I think there's a linter for this but unfortunately it only knows about `errors.Errorf` and not `errors.Wrap`.

pkg/reconciler/managed/reconciler.go
```diff
@@ -972,18 +1007,29 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (resu
...
 				// We'll hit this condition if we can't delete our external
 				// resource, for example if our provider credentials don't have
 				// access to delete it. If this is the first time we encounter
 				// this issue we'll be requeued implicitly when we update our
 				// status with the new error condition. If not, we want requeue
 				// explicitly, which will trigger backoff.
 				log.Debug("Cannot delete external resource", "error", err)
+				if err := r.changeLogger.RecordChangeLog(ctx, managedPreOp, v1alpha1.OperationType_OPERATION_TYPE_DELETE, err, deletion.AdditionalDetails); err != nil {
+					log.Debug(errRecordChangeLog, "error", err)
+					record.Event(managed, event.Warning(reasonCannotSendChangeLog, err))
```
negz, line 1031: I could pretty easily be convinced that we should just log this, and not create an event.

Change logging not working does seem like more of an operator issue than something an MR author would be concerned with. So maybe we don't emit an event, but log the error at info level?

pkg/reconciler/managed/reconciler.go
```diff
@@ -506,6 +531,7 @@ type Reconciler struct {
 	log            logging.Logger
 	record         event.Recorder
 	metricRecorder MetricRecorder
+	changeLogger   ChangeLogger
```
negz, line 534: If you renamed `RecordChangeLog` to just `Log` and renamed this field to `change` then callers would write `r.change.Log(ctx, ...)`. I like the way that reads (more like prose, less stutter-ey).

negz (commented):
A few last small nits, but this looks great to me. Thanks @jbw976!

pkg/reconciler/managed/reconciler.go
jbw976: @negz and I ended up at just logging in https://github.com/crossplane/crossplane-runtime/pull/754/commits/f7fb3a733a63a95ba0d4eb5f5555105d247dabd8

pkg/reconciler/managed/reconciler.go
jbw976: @negz I went ahead and made this change in https://github.com/crossplane/crossplane-runtime/pull/754/commits/f7fb3a733a63a95ba0d4eb5f5555105d247dabd8. With the reading and prototyping I did with this, I like the behavior we get here. We'll wait/retry if the server (change logs sidecar) is temporarily unreachable for a default of 10s, and that value is configurable with a new `WithSendTimeout()` option. 

Let me know if you don't think this is a good approach, but with my testing I think it adds a bit more robustness for sending the change logs without making the `Reconcile()` hang for too long.

pkg/reconciler/managed/reconciler.go
jbw976: Sounds good, we are no longer recording events and we are logging at info level in https://github.com/crossplane/crossplane-runtime/pull/754/commits/f7fb3a733a63a95ba0d4eb5f5555105d247dabd8

pkg/reconciler/managed/reconciler.go
jbw976: niiice, I like the way this reads too! good style here 😎 - did this in https://github.com/crossplane/crossplane-runtime/pull/754/commits/f7fb3a733a63a95ba0d4eb5f5555105d247dabd8

jbw976:
@negz I've got all of your feedback integrated now, thanks for the awesome reviews dude! 😎 

Pay special attention to https://github.com/crossplane/crossplane-runtime/pull/754#discussion_r1710728868 (using `WaitForReady()`, since that's not something you asked for, but I do think it's an improvement personally.

I've pushed all changes to my `provider-kubernetes` fork and the https://github.com/jbw976/change-log-sidecar prototype server implementation:

* https://github.com/crossplane-contrib/provider-kubernetes/compare/main...jbw976:provider-kubernetes:change-logs
* https://github.com/jbw976/change-log-sidecar 

Testing instructions and published images are also up to date in https://github.com/jbw976/change-log-sidecar/blob/main/README.md. Let me know if there's anything else I can improve with this PR! 🙇‍♂️

apis/changelogs/proto/v1alpha1/changelog.proto
```diff
@@ -0,0 +1,84 @@
...
+
+// SendChangeLogRequest represents a request to send a single change log entry.
+message SendChangeLogRequest {
+  // The change log entry to send as part of this request.
+  ChangeLogEntry entry = 1;
+}
+
+// ChangeLogEntry represents a single change log entry, with detailed information
+// about the resource that was changed.
+message ChangeLogEntry {
```
negz, line 38: One thing I totally spaced on - I think we should record a timestamp client side and send it in the RPC. I see you're currently recording it in the sidecar. It probably doesn't make a huge difference, but given we'll tolerate a delay of up to 10 seconds when sending the RPC, recording the timestamp in the sidecar could cause changes to appear to happen slightly later than they actually did.

I see there's a well-known type we can use: https://protobuf.dev/reference/protobuf/google.protobuf/#timestamp

negz (approved):
Thanks @jbw976! One final thing I forgot, but otherwise this looks good. Feel free to squash and merge once you've added the timestamp.

apis/changelogs/proto/v1alpha1/changelog.proto
jbw976: yep that is a good idea for this scenario. Typically server side timestamps are used, but in this scenario with just one client and a potential sending delay, it make sense to standardize on the client side. 

Done in https://github.com/crossplane/crossplane-runtime/pull/754/commits/b3b7e6da82e2823c45b552e186ef6bfb8277110d

jbw976:
Awesome, 0 unresolved conversations, so I will likely merge this today after a bit more testing 😎

PR #778, opened 2024-09-26
Add `paved.GetFloat()` function

rickard-von-essen:
This adds `paved.GetFloat()` function similar to `paved.GetString()` and `paved.GetInteger()` but returning a `float64`.

I have: 

### DIFF
--- a/pkg/fieldpath/paved.go
+++ b/pkg/fieldpath/paved.go
@@ -263,6 +263,20 @@ func (p *Paved) GetString(path string) (string, error) {
 	return s, nil
 }
 
+// GetFloat value of the supplied field path.
+func (p *Paved) GetFloat(path string) (float64, error) {
+	v, err := p.GetValue(path)
+	if err != nil {
+		return 0, err
+	}
+
+	f, ok := v.(float64)
+	if !ok {
+		return 0, errors.Errorf("%s: not a number", path)
+	}
+	return f, nil
+}
+
 // GetStringArray value of the supplied field path.
 func (p *Paved) GetStringArray(path string) ([]string, error) {
 	v, err := p.GetValue(path)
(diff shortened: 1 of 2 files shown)


negz:
Thanks! Can you provide some context around what you need this for?

I vaguely recall that we actually used to have something like this but intentionally dropped it, because you're not supposed to use floats in Kubernetes APIs. From https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types:

> Avoid floating-point values as much as possible, and never use them in spec. Floating-point values cannot be reliably round-tripped (encoded and re-decoded) without changing, and have varying precision and representations across languages and architectures.

rickard-von-essen:
Sorry I should have provided some context. 

I needed this in this [PR provider-upjet-gcp#623](https://github.com/crossplane-contrib/provider-upjet-gcp/pull/623/files#diff-b07deabbe5860b3369e426a8ab6680cdf9855ecd824eabc2f28dbcfeb3f66fedR43-R47) to read the `port` (as in IPv4 port). 

I first assumed it was a _integer_ since inherently a port is an int and the [Terraform schema](https://github.com/hashicorp/terraform-provider-google/blob/main/google/services/redis/resource_redis_cluster.go#L283-L287) mark it as `schema.TypeInt`. But after some debugging I realize it was [passed as a json `number`](https://github.com/crossplane-contrib/provider-upjet-gcp/pull/623/files#diff-01ebba9f28e340b3a7edc2839a4b640664e8fb2da5c947de235befb30e982d74R595-R598) which translates into a float.

negz:
@rickard-von-essen I'm pretty confident this is a bug in upjet - these fields should not be floats in the Go structs. That violates the Kubernetes API conventions I linked above. It also doesn't make sense in most of the cases - ports, replica counts, etc presumably can never actually be a floating point number.

So, I suggest we fix the bug (which I assume is in upjet) rather than adding a `GetFloat` utility here.

rickard-von-essen:
@negz Thanks for your insight on this. I'll close this and see if I have some time to chase down the issue in upjet.

PR #786, opened 2024-11-04
Fix unknown fields warnings and possibility to suppress finalizer ones

turkenh:

This PR aims to clean up warnings observed in https://github.com/crossplane/crossplane/issues/6043 by:
- Fixing "unknown field" warnings using the correct structs matching schemas.
- ~Suppressing "prefer a domain-qualified finalizer" until we properly migrate them to the suggested format.~

Together with the [accompanying PR](https://github.com/crossplane/crossplane/pull/6064) in Crossplane, I could validate that former warnings disappeared with the changes here.

~This PR introduces a new package, `warning`, with the `warning.LoggingHandler` implementing [rest.WarningHandler](https://github.com/kubernetes/client-go/blob/master/rest/warnings.go#L31) interface. This supports deduplication just like the [current handler](https://github.com/crossplane/crossplane/blob/e4d72e53d33c5d3c982a76197d43e850c482206c/cmd/crossplane/core/core.go#L147) Crossplane uses, but it also uses provided crossplane runtime logger so that warnings logged as structured logs* and also enables discarding logs matching provided regular expressions. I opted to put this into crossplane runtime since we may want to leverage it in other repositories, e.g. providers, as well.~

I have: 

- [x] Added `backport release-x.y` labels to auto-backport this PR.

### DIFF
--- a/pkg/resource/fake/mocks.go
+++ b/pkg/resource/fake/mocks.go
@@ -35,7 +35,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
-	"github.com/crossplane/crossplane-runtime/pkg/resource/unstructured/claim"
+	"github.com/crossplane/crossplane-runtime/pkg/resource/unstructured/reference"
 )
 
 // Conditioned is a mock that implements Conditioned interface.
@@ -50,13 +50,13 @@ func (m *Conditioned) GetCondition(ct xpv1.ConditionType) xpv1.Condition {
 }
 
 // ClaimReferencer is a mock that implements ClaimReferencer interface.
-type ClaimReferencer struct{ Ref *claim.Reference }
+type ClaimReferencer struct{ Ref *reference.Claim }
 
 // SetClaimReference sets the ClaimReference.
-func (m *ClaimReferencer) SetClaimReference(r *claim.Reference) { m.Ref = r }
+func (m *ClaimReferencer) SetClaimReference(r *reference.Claim) { m.Ref = r }
 
 // GetClaimReference gets the ClaimReference.
-func (m *ClaimReferencer) GetClaimReference() *claim.Reference { return m.Ref }
+func (m *ClaimReferencer) GetClaimReference() *reference.Claim { return m.Ref }
 
 // ManagedResourceReferencer is a mock that implements ManagedResourceReferencer interface.
 type ManagedResourceReferencer struct{ Ref *corev1.ObjectReference }
@@ -184,15 +184,15 @@ func (m *CompositionSelector) SetCompositionSelector(s *metav1.LabelSelector) { 
 func (m *CompositionSelector) GetCompositionSelector() *metav1.LabelSelector { return m.Sel }
 
 // CompositionRevisionReferencer is a mock that implements CompositionRevisionReferencer interface.
-type CompositionRevisionReferencer struct{ Ref *corev1.ObjectReference }
+type CompositionRevisionReferencer struct{ Ref *corev1.LocalObjectReference }
 
 // SetCompositionRevisionReference sets the CompositionRevisionReference.
-func (m *CompositionRevisionReferencer) SetCompositionRevisionReference(r *corev1.ObjectReference) {
+func (m *CompositionRevisionReferencer) SetCompositionRevisionReference(r *corev1.LocalObjectReference) {
 	m.Ref = r
 }
 
 // GetCompositionRevisionReference gets the CompositionRevisionReference.
-func (m *CompositionRevisionReferencer) GetCompositionRevisionReference() *corev1.ObjectReference {
+func (m *CompositionRevisionReferencer) GetCompositionRevisionReference() *corev1.LocalObjectReference {
 	return m.Ref
 }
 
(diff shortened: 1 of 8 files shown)


pkg/resource/unstructured/claim/claim.go
```diff
@@ -63,21 +64,6 @@ type Unstructured struct {
 	unstructured.Unstructured
 }
 
-// Reference to a claim.
-type Reference struct {
```
turkenh, line 67: If I simply added a Reference type under composite package, I was getting import cycle. So, I moved them into a separate `reference` package.

pkg/resource/unstructured/claim/claim.go
```diff
@@ -112,16 +98,16 @@ func (c *Unstructured) SetCompositionReference(ref *corev1.ObjectReference) {
...
+	out := &corev1.LocalObjectReference{}
 	if err := fieldpath.Pave(c.Object).GetValueInto("spec.compositionRevisionRef", out); err != nil {
 		return nil
 	}
 	return out
 }
 
 // SetCompositionRevisionReference of this resource claim.
-func (c *Unstructured) SetCompositionRevisionReference(ref *corev1.ObjectReference) {
+func (c *Unstructured) SetCompositionRevisionReference(ref *corev1.LocalObjectReference) {
```
turkenh, line 110: See https://github.com/crossplane/crossplane/blob/e4d72e53d33c5d3c982a76197d43e850c482206c/internal/xcrd/schemas.go#L222-L227

pkg/resource/unstructured/claim/claim.go
```diff
@@ -170,22 +156,22 @@ func (c *Unstructured) GetCompositeDeletePolicy() *xpv1.CompositeDeletePolicy {
...
+	out := &reference.Composite{}
 	if err := fieldpath.Pave(c.Object).GetValueInto("spec.resourceRef", out); err != nil {
 		return nil
 	}
 	return out
 }
 
 // SetResourceReference of this composite resource claim.
-func (c *Unstructured) SetResourceReference(ref *corev1.ObjectReference) {
+func (c *Unstructured) SetResourceReference(ref *reference.Composite) {
```
turkenh, line 168: See https://github.com/crossplane/crossplane/blob/e4d72e53d33c5d3c982a76197d43e850c482206c/internal/xcrd/schemas.go#L256-L263

pkg/resource/unstructured/composite/composite.go
```diff
@@ -99,16 +99,16 @@ func (c *Unstructured) SetCompositionReference(ref *corev1.ObjectReference) {
 }
 
 // GetCompositionRevisionReference of this Composite resource.
-func (c *Unstructured) GetCompositionRevisionReference() *corev1.ObjectReference {
-	out := &corev1.ObjectReference{}
+func (c *Unstructured) GetCompositionRevisionReference() *corev1.LocalObjectReference {
+	out := &corev1.LocalObjectReference{}
```
turkenh, line 103: See https://github.com/crossplane/crossplane/blob/e4d72e53d33c5d3c982a76197d43e850c482206c/internal/xcrd/schemas.go#L95-L100

phisco (approved):
not a huge fan of suppressing warnings, I would have probably preferred to keep them there and fix them in a follow-up PR.

pkg/resource/fake/mocks.go
```diff
@@ -50,22 +50,22 @@ func (m *Conditioned) GetCondition(ct xpv1.ConditionType) xpv1.Condition {
...
-func (m *ClaimReferencer) SetClaimReference(r *claim.Reference) { m.Ref = r }
+func (m *ClaimReferencer) SetClaimReference(r *reference.Claim) { m.Ref = r }
 
 // GetClaimReference gets the ClaimReference.
-func (m *ClaimReferencer) GetClaimReference() *claim.Reference { return m.Ref }
+func (m *ClaimReferencer) GetClaimReference() *reference.Claim { return m.Ref }
 
 // ManagedResourceReferencer is a mock that implements ManagedResourceReferencer interface.
-type ManagedResourceReferencer struct{ Ref *corev1.ObjectReference }
+type ManagedResourceReferencer struct{ Ref *reference.Composite }
```
jbw976, line 62: are these updates to `ManagedResourceReferencer` correct? Because it is called `ManagedResource*`, I would expect them not to only specifically point to a `Composite` now 🤔

pkg/warning/warning.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	return k
+}
+
+// DomainQualifiedFinalizer returns a regular expression that matches warnings
+// that suggest a domain-qualified finalizer name should be used to avoid
+// accidental conflicts with other finalizer writers.
+func DomainQualifiedFinalizer(domain string) regexp.Regexp {
+	return *regexp.MustCompile(
+		fmt.Sprintf("^metadata.finalizers:.*%s.*prefer a domain-qualified finalizer name to avoid accidental conflicts with other finalizer writers$", domain),
```
jbw976, line 109: will this check start to fail once we pick up https://github.com/kubernetes/kubernetes/pull/127303/files which changes the warning message?

pkg/warning/warning.go
```diff
@@ -0,0 +1,111 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package warning contains utilities for handling warnings.
+package warning
```
jbw976, line 18: i'm wondering how essential this is actually - it doesn't fix the underlying issue right, it just hides it from the logs? i'm wondering if we just live with the warnings in the logs now, and resolve them with a real fix during the 1.19 milestone, instead of hiding them for now and maybe forgetting about them. What do you think?

jbw976 (commented):
Thanks for continuing to tackle issues quickly for v1.18 @turkenh!! a few questions for you 🙇‍♂️

pkg/resource/fake/mocks.go
turkenh: Great catch! Fixing linter issues one by one and not sure how/why I changed this one 🤔

pkg/warning/warning.go
turkenh: I'll drop the last commit, e.g., finalizer log suppressing, since both you and @phisco prefer keeping logs around instead, and I am not too opinionated, at least to consider something that important doing last minute :)

PR #788, opened 2024-11-10
feat: support additional management policy combination

### LINKED ISSUE #681: Support "Update" without "Create" management policy combinations

Suppose you want to extend functionality of a resource of which you should not be managing the lifecycle of. To extend functionality of this resource you may need to add some annotations as is common in Kubernetes.

A more concrete example is providing a Kubernetes service account with AWS credentials via IRSA. This requires creating an IAM Role with the required policies and adding an annotation to the service account to link it to the IAM Role. However, this service account is created via a helm chart and you do not want to take "ownership" of the resource.

The problem is deciding on the right supported combination of management policies for the service account resource. All combinations of `Update` require `Create` to also be given. But this is not in the scope of this composition, it is designed to only annotate the service account if it exists and should fail otherwise. We are assuming a pre-existing resource, and will try again later if it is missing, presumably when the helm chart is installed.

Crossplane could solve this problem by allowing a combination of `Observe` and `Update` management policies so that the provider can correctly annotate the service account without creating it if it is missing. The managed resource should fail to sync if it is missing and therefore cause the containing composition to not be ready.

man-ish-k:
Add two new management policy combinations to support safer resource importing and external lifecycle management scenarios:

1. **Import Mode**: A new policy combination that allows safe importing of existing resources by only permitting observation and late initialization, without making any changes to the external resource.
   - Actions: `[Observe, LateInitialize]`
   - Use case: Safely discover and import existing cloud resources into Crossplane without risk of modifying them

2. **External Lifecycle Management**: A policy combination for resources whose lifecycle (creation/deletion) is managed outside of Crossplane, but allows Crossplane to make updates and track state.
   - Actions: `[Observe, Update, LateInitialize]`
   - Use case: Integration with external provisioning systems while maintaining Crossplane's ability to manage resource configuration

Fixes #681, 

I have: 


jake-ciolek:
Hi @jbw976 and @negz . Could you please have a look at this one? It's been open for a while. Thanks!

pkg/reconciler/managed/policies_test.go
```diff
@@ -0,0 +1,362 @@
...
+
+import (
+	"testing"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
+)
+
+func TestManagementPoliciesResolver_Validate(t *testing.T) {
```
lsviben, line 11: All these tests below exist in [reconciler_test.go](https://github.com/crossplane/crossplane-runtime/blob/main/pkg/reconciler/managed/reconciler_test.go). I agree that as at one point the policy resolver moved to `policies.go` that its tests should reside in `policies_test.go`. But we shouldnt duplicate and that can be done as a separate change. 

The unit tests that actually need to be added is testing the new combinations in [reconciler_test.go](https://github.com/crossplane/crossplane-runtime/blob/main/pkg/reconciler/managed/reconciler_test.go) `TestReconcile`.

lsviben (changes requested):
Firstly, sorry for waiting that long for a review, this slipped my mind and I got reminded just recently 🙏 .

The policies added make sense to me. 

We need to add the unit tests in the right place to check the reconciler flow with the new policies.

pkg/reconciler/managed/policies_test.go
man-ish-k: Thanks @lsviben I have made the requested changes.

lsviben (approved):
Looks good! Thanks @man-ish-k ! 🎉

turkenh (approved):

PR #804, opened 2025-01-17
Fix bug causing status to be cleared after adding Finalizer

### LINKED ISSUE #791: AddFinalizer clears Status subobject leading to delayed object readiness
### What happened?

We create managed resources using _only_ the `Observe` management policy and the provider will initially set the `Synced` condition to `True` but will not set the `Ready` condition at all.
We can see in the provider logs that the object seems to be successfully reconciled and is requeued according to the configured `sync interval`.
Once the sync interval is hit, the resource becomes ready.

### How can we reproduce it?

1. Create any MR in the cluster with only the `Observe` management policy. We can easily reproduce this with `provider-upjet-azure` in any resource, for instance `VirtualNetwork` with the following spec:
```yaml
apiVersion: network.azure.upbound.io/v1beta2
kind: VirtualNetwork
metadata:
  annotations:
    crossplane.io/external-name: vnet-test-crossplane
  name: vnet-test-crossplane
spec:
  deletionPolicy: Delete
  forProvider:
    resourceGroupName: rg-test-crossplane
  initProvider: {}
  managementPolicies:
  - Observe
  providerConfigRef:
    name: default
```
2. You will see the object become `Synced` almost immediately, but the `Ready` condition will stay `nil`. Also, the whole `atProvider` struct stays `nil`. Provider logs will show the following:
```
2024-12-06T20:57:33Z    DEBUG    provider-azure    Reconciling    {"controller": "managed/network.azure.upbound.io/v1beta1, kind=virtualnetwork", "request": {"name":"vnet-test-crossplane"}}
2024-12-06T20:57:33Z    DEBUG    provider-azure    Connecting to the service provider    {"uid": "bb757cb5-8934-471d-981b-7862513ac9b1", "name": "vnet-test-crossplane", "gvk": "network.azure.upbound.io/v1beta1, Kind=VirtualNetwork"}
2024-12-06T20:57:33Z    DEBUG    provider-azure    Instance state not found in cache, reconstructing...    {"uid": "bb757cb5-8934-471d-981b-7862513ac9b1", "name": "vnet-test-crossplane", "gvk": "network.azure.upbound.io/v1beta1, Kind=VirtualNetwork"}
2024-12-06T20:57:33Z    DEBUG    provider-azure    Observing the external resource    {"uid": "bb757cb5-8934-471d-981b-7862513ac9b1", "name": "vnet-test-crossplane", "gvk": "network.azure.upbound.io/v1beta1, Kind=VirtualNetwork"}
2024-12-06T20:57:34Z    DEBUG    provider-azure    Diff detected    {"uid": "bb757cb5-8934-471d-981b-7862513ac9b1", "name": "vnet-test-crossplane", "gvk": "network.azure.upbound.io/v1beta1, Kind=VirtualNetwork", "instanceDiff": "*terraform.InstanceDiff{mu:sync.Mutex{state:0, sema:0x0}, Attributes:map[string]*terraform.ResourceAttrDiff{\"location\":*terraform.ResourceAttrDiff{Old:\"canadacentral\", New:\"\", NewComputed:false, NewRemoved:true, NewExtra:interface {}(nil), RequiresNew:true, Sensitive:false, Type:0x0}}, Destroy:false, DestroyDeposed:false, DestroyTainted:false, RawConfig:cty.NilVal, RawState:cty.NilVal, RawPlan:cty.NilVal, Meta:map[string]interface {}(nil)}"}
2024-12-06T20:57:34Z    DEBUG    provider-azure    Skipping update due to managementPolicies. Reconciliation succeeded    {"controller": "managed/network.azure.upbound.io/v1beta1, kind=virtualnetwork", "request": {"name":"vnet-test-crossplane"}, "uid": "bb757cb5-8934-471d-981b-7862513ac9b1", "version": "1176637016", "external-name": "vnet-test-crossplane", "requeue-after": "2024-12-07T02:49:57Z"} 
```
4. You can either wait for the requeue to trigger or you can restart the provider.
5. After that, the `Ready` condition will be set to `true`.

Crossplane version: 1.18.0

### Additional information

After step-by-step debugging, we have found that the bug seems to come from the `AddFinalizer` function in the `APIFinalizer` struct. This struct is used by `provider-upjet-azure` and probably all `upjet` based providers (although I didn't validate).
The `AddFinalizer` function calls the kube client's `Update` function which will apply an object in the kube API, but also return the updated object returned by the API to the caller. This, in turn, nullifies any struct that is not applied by the `Update` function such as the `Status` sub-struct.

Here are the relevant parts of the code:
- [AddFinalizer function](https://github.com/crossplane/crossplane-runtime/blob/19d95a69cc03690c4b867ff91d89681fcf872a93/pkg/resource/api.go#L161C1-L168C2)
- [Reconciler's call to AddFinalizer](https://github.com/crossplane/crossplane-runtime/blob/19d95a69cc03690c4b867ff91d89681fcf872a93/pkg/reconciler/managed/reconciler.go#L1156C1-L1166C3)

gravufo:

This PR will fix the issue where the status sub-object would be cleared when adding the Finalizer. This would cause Observe-only resources to wait until the next SyncPeriod/PollPeriod to become Ready=true even though it should have been set at the first loop.

Fixes #791 

Note: I am really unsure if these are the right places to do those changes, so please don't hesitate to leave comments pointing to the right locations.

I have: 


turkenh (commented):
Thanks @gravufo for the PR and sorry for the late response here!

The PR looks good to me in general, except adding `Statused` to the `managed` interface feels a bit off since there is `ConditionedStatus` which also controls part of status and related to why unit test failing. 

If we want to proceed with the proposed approach, we will also need a controller tools PR so that all managed resources satisfy the extended interface.

Another solution could be, instead of changing managed reconciler, we could fix `APIFinalizer`'s `AddFinalizer()` method such that it won't override the status. A simple deepcopy and using copied object in the Update call would be helpful, also eliminating the need to update managed interface/reconciler. 

```diff
func (a *APIFinalizer) AddFinalizer(ctx context.Context, obj Object) error {
	if meta.FinalizerExists(obj, a.finalizer) {
                return nil
        }
        meta.AddFinalizer(obj, a.finalizer)
-       return errors.Wrap(a.client.Update(ctx, obj), errUpdateObject)
+       // Copy object to ensure we don't modify the original object, e.g. not to lose status.
+       o := obj.DeepCopyObject().(Object)
+       return errors.Wrap(a.client.Update(ctx, o), errUpdateObject)
 }
```

While this change looks simpler (and wouldn't require a crossplane tools), I am not comfortable since it will have a bigger blast radius due to APIFinalizer has other usages. If there are callers of that method assuming they got the latest object after the add finalizer call, they might get broken. For example, the next update call will fail since the object don't have the latest resource version any more 🤔

apis/common/v1/status.go
```diff
@@ -0,0 +1,29 @@
...
+*/
+
+package v1
+
+import "k8s.io/apimachinery/pkg/runtime"
+
+// See https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties
+
+// A Status reflects the observed status of a resource.
+type Status map[string]runtime.RawExtension
```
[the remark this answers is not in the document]
n3wscott: I think to be able to reach in and get/set conditions directly?

pkg/resource/unstructured/composed/composed.go
```diff
@@ -89,6 +89,20 @@ func (cr *Unstructured) SetConditions(c ...xpv1.Condition) {
 	_ = fieldpath.Pave(cr.Object).SetValue("status.conditions", conditioned.Conditions)
 }
 
+// GetStatus of this Composed resource.
+func (cr *Unstructured) GetStatus() xpv1.Status {
+	status := xpv1.Statused{}
+	if err := fieldpath.Pave(cr.Object).GetValueInto("status", &status); err != nil {
```
n3wscott, line 95: I am concerned about `status.conditions` and `.status` being stored independently.

pkg/resource/interfaces.go
```diff
@@ -28,13 +28,20 @@ import (
...
+// A Conditioned may have conditions set or retrieved. Conditions typically
 // indicate the status of both a resource and its reconciliation process.
 type Conditioned interface {
 	SetConditions(c ...xpv1.Condition)
 	GetCondition(ct xpv1.ConditionType) xpv1.Condition
 }
 
+// A Statused may have status set or retrieved. Status is typically
+// a struct containing various runtime status properties.
+type Statused interface {
```
n3wscott, line 40: I am wondering if we need this... perhaps we just need `Pave.SyncStatus(old, new)`?

pkg/reconciler/managed/reconciler.go
```diff
@@ -1164,6 +1164,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (resu
 		managed.SetConditions(xpv1.ReconcileError(err))
 		return reconcile.Result{Requeue: true}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
 	}
+	// Set the status back to its original value before the AddFinalizer call.
+	// The AddFinalizer call may overwrite the managed object with what is
+	// currently in the kube API which would make us lose information that may
+	// have been populated by the Observe call.
+	managed.SetStatus(managedPreOp.GetStatus())
```
n3wscott, line 1171: Can we try to make a new pave function that copies the original pre-op status into managed rather than introduce the .Set/GetStatus?

I think it has the same effect and does not spread out the API

turkenh:
> No problem, I didn't push too much on this one, because I am very hesitant with my proposed fix. The problem is that this fixes only a single use-case, but doesn't really fix all the other use-cases (wanting to modify status during a Create/Update for instance). Basically, any call that will update the object spec makes us lose the status sub-object modifications. In this case it's a little worse because we literally lose the full sub-object since it's a brand new MR which had no status inside kubernetes yet.

FWIW, we are good with other cases in the managed reconciler, and this (after observe, patching finalizer) is the only remaining one.

- It does a `kube.Update()` after `Create` for late-initialization, but this triggers a new reconciler and next time the status is persisted properly. See [this comment](https://github.com/crossplane/crossplane-runtime/blob/532a1c432f02f4b7c2ece99b882cd9be4100b18d/pkg/reconciler/managed/reconciler.go#L1253-L1259).
- There is no `kube.Update()` after `external.Update()`, so status being persisted successfully.

It is clear what we need to do, the question is simply how :D

a. Follow the existing convention and define interfaces to work with Status (what this PR doing).
  i. Need crossplane-tools update to generate new interfaces.
  ii. Status and status conditions being managed independently via 2 different getter/setter.
b. Make an exception and directly patch unstructured object by paving this time.
  i. The first time we will be using pave/unstructured in managed reconciler.
c. Update `APIFinalizer`'s `AddFinalizer()` method, so that it doesn't lose un-persisted status.
  i. Possible behavior change for other callers of `AddFinalizer()`.  
  ii. We will still need unstructured/pave.

 I still lean on (a) but no strong opinions if we chose (b) or (c). @negz, interested to hear your thoughts.

turkenh:
> > * It does a `kube.Update()` after `Create` for late-initialization, but this triggers a new reconciler and next time the status is persisted properly. See [this comment](https://github.com/crossplane/crossplane-runtime/blob/532a1c432f02f4b7c2ece99b882cd9be4100b18d/pkg/reconciler/managed/reconciler.go#L1253-L1259).
> 
> Maybe I didn't understand properly, but usually the create is only called once, so if that first time it's called we lose the status, we'll never restore what the Create wanted to save, no?

You're right. However, in practice, I am not aware of any information that is unique to the Create call and not obtained in an Observe call. The only exception could be sensitive information that external API returning only after Create (e.g. API secret key), but those are stored into secrets as connection details instead of object status. So, the next Observe is expected to form the latest status by making a fresh observation and we're not worried about the status after Create getting lost.

negz:
@turkenh and I chatted about this a bit the other day. I agree updating resource.Managed to have GetStatus and SetStatus methods would be the most idiomatic way to do it. That'd require a lot of updates across the ecosystem though, and ultimately those methods would need to return and accept opaque data. So maybe we may as well just use fieldpath.Pave?

It kinda kills me that we have to do this (save and restore status) at all though. It seems like an anti-pattern in that it basically works around optimistic concurrency. If we ever added another controller that wanted to update MR status we'd have a race:

1. MR reconciler saves status
2. MR reconciler adds finalizer
3. Other reconciler updates status (bumping `resourceVersion`)
4. MR reconciler restores status, undoing 3

PR #789, opened 2024-11-12
chore: added observedGeneration to the Ready and Sync condition

pintonunes:
### Description of your changes

Today is pretty impossible to know when a managed resource is fully reconciled after an update.
On creation we can check the `Ready` and `Sync` conditions to be `True` but on updates we never know if what we are seeing is the actual or the previous state. 

Imagine we are updating an RDS cluster (long running async update) and by looking to the CR we cant tell if the update is done, in progress or if the controller is actually dead. 

This PR adds the `observedGeneration` to the 'Ready' and 'Sync' conditions in the main reconcile loop since it was introduced with #633 

Related with: 
https://github.com/crossplane/crossplane/issues/4695
https://github.com/crossplane/crossplane/issues/4655

I have: 

- [ ] Read and followed Crossplane's [contribution process].
- [ ] Run `earthly +reviewable` to ensure this PR is ready for review.
- [ ] Added or updated unit tests.
- [ ] Linked a PR or a [docs tracking issue] to [document this change].
- [ ] Added `backport release-x.y` labels to auto-backport this PR.

Need help with this checklist? See the [cheat sheet].

[contribution process]: https://github.com/crossplane/crossplane/tree/main/contributing
[docs tracking issue]: https://github.com/crossplane/docs/issues/new
[document this change]: https://docs.crossplane.io/contribute/contribute
[cheat sheet]: https://github.com/crossplane/crossplane/tree/main/contributing#checklist-cheat-sheet

PR #828, opened 2025-05-01
Update Managed reconciler to store observedGeneration in conditions

n3wscott:

Update `Managed` reconciler to store observedGeneration on conditions and also update that controllers unit tests.

This uses the new `conditions.Manager` to help propagate the `observedGeneration` to the changed conditions within the reconciler.

Relates to https://github.com/crossplane/crossplane/issues/6420
Superseedes https://github.com/crossplane/crossplane-runtime/pull/789
Depends on https://github.com/crossplane/crossplane-runtime/pull/831/

I have: 

### DIFF
--- a/apis/common/v1/condition.go
+++ b/apis/common/v1/condition.go
@@ -96,12 +96,13 @@ type Condition struct {
 }
 
 // Equal returns true if the condition is identical to the supplied condition,
-// ignoring the LastTransitionTime and ObservedGeneration.
+// ignoring the LastTransitionTime.
 func (c Condition) Equal(other Condition) bool {
 	return c.Type == other.Type &&
 		c.Status == other.Status &&
 		c.Reason == other.Reason &&
-		c.Message == other.Message
+		c.Message == other.Message &&
+		c.ObservedGeneration == other.ObservedGeneration
 }
 
 // WithMessage returns a condition by adding the provided message to existing
@@ -167,28 +168,24 @@ func (s *ConditionedStatus) GetCondition(ct ConditionType) Condition {
 // SetConditions sets the supplied conditions, replacing any existing conditions
 // of the same type. This is a no-op if all supplied conditions are identical,
 // ignoring the last transition time, to those already set.
-// Observed generation is updated if higher than the existing one.
 func (s *ConditionedStatus) SetConditions(c ...Condition) {
-	for _, new := range c {
+	for _, cond := range c {
 		exists := false
 		for i, existing := range s.Conditions {
-			if existing.Type != new.Type {
+			if existing.Type != cond.Type {
 				continue
 			}
 
-			if existing.Equal(new) {
+			if existing.Equal(cond) {
 				exists = true
-				if existing.ObservedGeneration < new.ObservedGeneration {
-					existing.ObservedGeneration = new.ObservedGeneration
-				}
 				continue
 			}
 
-			s.Conditions[i] = new
+			s.Conditions[i] = cond
 			exists = true
 		}
 		if !exists {
-			s.Conditions = append(s.Conditions, new)
+			s.Conditions = append(s.Conditions, cond)
 		}
 	}
 }
--- a/pkg/reconciler/managed/reconciler.go
+++ b/pkg/reconciler/managed/reconciler.go
@@ -1107,9 +1107,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (resu
 				log.Info(errRecordChangeLog, "error", err)
 			}
 			record.Event(managed, event.Normal(reasonDeleted, "Successfully requested deletion of external resource"))
-			managed.SetConditions(
-				xpv1.Deleting().WithObservedGeneration(managed.GetGeneration()),
-				xpv1.ReconcileSuccess().WithObservedGeneration(managed.GetGeneration()))
+			status.MarkConditions(xpv1.Deleting(), xpv1.ReconcileSuccess())
 			return reconcile.Result{Requeue: true}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedStatus)
 		}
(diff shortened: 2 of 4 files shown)


apis/common/v1/condition.go
```diff
@@ -101,7 +101,8 @@ func (c Condition) Equal(other Condition) bool {
 	return c.Type == other.Type &&
 		c.Status == other.Status &&
 		c.Reason == other.Reason &&
-		c.Message == other.Message
+		c.Message == other.Message &&
+		c.ObservedGeneration == other.ObservedGeneration
```
n3wscott, line 105: Updating to not treat ObservedGeneration as something to ignore between versions of conditions.

apis/common/v1/condition.go
```diff
@@ -169,26 +170,23 @@ func (s *ConditionedStatus) GetCondition(ct ConditionType) Condition {
 // ignoring the last transition time, to those already set.
 // Observed generation is updated if higher than the existing one.
 func (s *ConditionedStatus) SetConditions(c ...Condition) {
-	for _, new := range c {
+	for _, cond := range c {
```
n3wscott, line 173: scope creep, but it is not good to use a semi-reserved word as a local var.

negz: We lint for this in c/c. Wonder if this repo's golangci-lint config is lagging behind.

apis/common/v1/condition.go
negz: Looks like the godoc string needs an update for this too.

pkg/reconciler/managed/reconciler.go
```diff
@@ -180,7 +180,7 @@ type ConnectionDetailsFetcher interface {
 	FetchConnection(ctx context.Context, so resource.ConnectionSecretOwner) (ConnectionDetails, error)
 }
 
-// A Initializer establishes ownership of the supplied Managed resource.
+// Initializer establishes ownership of the supplied Managed resource.
```
negz, line 183: I started running https://writewithharper.com/docs/integrations/language-server in my editor recently and it's finding these eeeeeverywhere heh.

pkg/reconciler/managed/reconciler.go
```diff
@@ -880,7 +880,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (resu
 		log.Debug("Reconciliation is paused either through the `spec.managementPolicies` or the pause annotation", "annotation", meta.AnnotationKeyReconciliationPaused)
 		record.Event(managed, event.Normal(reasonReconciliationPaused, "Reconciliation is paused either through the `spec.managementPolicies` or the pause annotation",
 			"annotation", meta.AnnotationKeyReconciliationPaused))
-		managed.SetConditions(xpv1.ReconcilePaused())
+		managed.SetConditions(xpv1.ReconcilePaused().WithObservedGeneration(managed.GetGeneration()))
```
negz, line 883: Is there ever a case where we want to set a condition inside a reconciler, but _not_ its generation to the generation of the resource?

```go
cm := resource.NewConditionManager(managed)
cm.SetConditions(...)
```

I'm imagining something like this that'd wrap `SetConditions` and automatically propagate the generation from `managed` to the conditions being set. I could be convinced though that this isn't worth doing now if the plan is to move to a more heavyweight condition manager.

n3wscott: I had that same thought and we can get there, but I was not sure yet about how all the interfaces work inside of crossplane so I did it hard mode and added the tests to make sure it is working and then I can look at removing stuff as we tighten up condition accessors.

n3wscott: > move to a more heavyweight condition manager

This is my plan, yes. But I need more context on the project before I can understand what the project needs :D

apis/common/v1/condition.go
n3wscott: Made an issue to follow up.

pkg/reconciler/managed/reconciler.go
n3wscott: Done. But we should land https://github.com/crossplane/crossplane-runtime/pull/831/f first and then I can rebase.

pkg/reconciler/managed/reconciler.go
```diff
@@ -1102,7 +1107,9 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (resu
 				log.Info(errRecordChangeLog, "error", err)
 			}
 			record.Event(managed, event.Normal(reasonDeleted, "Successfully requested deletion of external resource"))
-			managed.SetConditions(xpv1.Deleting(), xpv1.ReconcileSuccess())
+			managed.SetConditions(
+				xpv1.Deleting().WithObservedGeneration(managed.GetGeneration()),
+				xpv1.ReconcileSuccess().WithObservedGeneration(managed.GetGeneration()))
```
negz, line 1112: Should this use `MarkConditions`?

pkg/reconciler/managed/reconciler.go
```diff
@@ -1298,7 +1305,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (resu
 		// https://github.com/crossplane/crossplane/issues/289
 		reconcileAfter := r.pollIntervalHook(managed, r.pollInterval)
 		log.Debug("External resource is up to date", "requeue-after", time.Now().Add(reconcileAfter))
-		managed.SetConditions(xpv1.ReconcileSuccess())
+		managed.SetConditions(xpv1.ReconcileSuccess().WithObservedGeneration(managed.GetGeneration()))
```
negz, line 1308: Ditto here.

negz (approved):
I like the UX of this new API. Approving because the overall approach LGTM - I think you missed two spots in the reconciler that need updating though.

pkg/reconciler/managed/reconciler.go
n3wscott: yes thanks!!

PR #829, opened 2025-05-02
Draft: Add a namespaced event recorder

frigaut-orange:

Added a namespaced APIRecorder to create events only on namespaced resources.

Related issue : https://github.com/crossplane/crossplane/issues/6348
Related doc PR : https://github.com/crossplane/crossplane/pull/6419

I have: 

- [ ] Added or updated unit tests. (todo)
- [ ] Added `backport release-x.y` labels to auto-backport this PR. (todo)

### DIFF
--- a/pkg/controller/options.go
+++ b/pkg/controller/options.go
@@ -69,6 +69,8 @@ type Options struct {
 
 	// ChangeLogOptions for recording change logs.
 	ChangeLogOptions *ChangeLogOptions
+
+	NamespacedEvents bool
 }
 
 // ForControllerRuntime extracts options for controller-runtime.
--- a/pkg/event/event.go
+++ b/pkg/event/event.go
@@ -18,6 +18,7 @@ limitations under the License.
 package event
 
 import (
+	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/tools/record"
 )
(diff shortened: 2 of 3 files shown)


pkg/event/event.go
```diff
@@ -101,6 +102,40 @@ func (r *APIRecorder) WithAnnotations(keysAndValues ...string) Recorder {
...
+		APIRecorder{kube: r, annotations: map[string]string{}},
+	}
+}
+
+// Event records the supplied event.
+func (r *NamespacedAPIRecorder) Event(obj runtime.Object, e Event) {
+	m, err := meta.Accessor(obj)
+	// If we cannot determine if the object is namespace (i.e. err is not nil), don't do anything
+	if err == nil && m.GetNamespace() != "" && m.GetNamespace() != "default" {
+		r.APIRecorder.Event(obj, e)
```
chlunde, line 124: are the annotations dropped here, don't you need `AnnotatedEventf`? If you do, check out https://github.com/crossplane/crossplane-runtime/pull/838/files

frigaut-orange: The `APIRecorder` struct is embedded in `NamespacedAPIRecorder`, so it re-uses the `Event` method from that struct (defined L90), if that is what you mean ?

pkg/event/event.go
```diff
@@ -101,6 +102,40 @@ func (r *APIRecorder) WithAnnotations(keysAndValues ...string) Recorder {
...
+	return &NamespacedAPIRecorder{
+		APIRecorder{kube: r, annotations: map[string]string{}},
+	}
+}
+
+// Event records the supplied event.
+func (r *NamespacedAPIRecorder) Event(obj runtime.Object, e Event) {
+	m, err := meta.Accessor(obj)
+	// If we cannot determine if the object is namespace (i.e. err is not nil), don't do anything
+	if err == nil && m.GetNamespace() != "" && m.GetNamespace() != "default" {
```
negz: I wonder: instead of a new type, could we have `NewEventRecorder` accept and use a variadic number of filter functions?

```go
type FilterFn func(runtime.Object, e Event)

func NewAPIRecorder(r record.EventRecorder, fns ...FilterFn) *APIRecorder { }


func (r *APIRecorder) Event(obj runtime.Object, e Event) {
  for _, filter := range r.filterFns {
    if filter(obj, e) {
      return
    }
  }
   // Proceed as normal
}
```

I think this way you could define a `NamespacedObjectsOnly` `FilterFn` and pass it in.

frigaut-orange: That's a great idea, I will make the change

frigaut-orange: Done (updated PR and commit titles, updated https://github.com/crossplane/crossplane/pull/6425 accordingly)

pkg/controller/options.go
```diff
@@ -69,6 +69,8 @@ type Options struct {
 
 	// ChangeLogOptions for recording change logs.
 	ChangeLogOptions *ChangeLogOptions
+
+	NamespacedEvents bool
```
negz: I think this can be removed now.

pkg/event/event.go
```diff
@@ -77,16 +77,24 @@ type Recorder interface {
 type APIRecorder struct {
 	kube        record.EventRecorder
 	annotations map[string]string
+	filterFns   []FilterFn
 }
 
+type FilterFn func(obj runtime.Object, e Event) bool
```
negz, line 85: This needs a godoc string - I expect the linter will complain when it runs.

pkg/event/event.go
```diff
@@ -77,16 +77,24 @@ type Recorder interface {
...
+
 // NewAPIRecorder returns an APIRecorder that records Kubernetes events to an
 // APIServer using the supplied EventRecorder.
-func NewAPIRecorder(r record.EventRecorder) *APIRecorder {
-	return &APIRecorder{kube: r, annotations: map[string]string{}}
+func NewAPIRecorder(r record.EventRecorder, filterFns ...FilterFn) *APIRecorder {
```
negz:
```suggestion
func NewAPIRecorder(r record.EventRecorder, fns ...FilterFn) *APIRecorder {
```

Nit: I prefer a shorter var name. It's a short function and the `FilterFn` type in the signature lets you know what it's for.

negz:
Few nits but otherwise this LGTM.

Not sure why CI is reluctant to run - closed and reopened the issue to try to get it to start.

pkg/event/event.go
frigaut-orange: agreed, done

pkg/event/event.go
frigaut-orange: done

PR #831, opened 2025-05-02
Introduce a conditions manager.

n3wscott:

Introduce a conditions package with a light weight manager to help enable coordinated condition updates. This will grow over time, for now the focus is just on helping make `observedGeneration` propagate to the condition that is being updated.

Relates to https://github.com/crossplane/crossplane-runtime/pull/828


pkg/conditions/manager.go
```diff
@@ -0,0 +1,79 @@
...
+// ConditionSet holds operations for interacting with an object's conditions.
+type ConditionSet interface {
+	// Mark adds or updates the conditions onto the managed resource object. Unlike a "Set" method, this also can add
+	// contextual updates to the condition such as propagating the correct observedGeneration to the conditions being
+	// changed.
+	Mark(condition ...xpv1.Condition)
+}
+
+// New returns an implementation of a Manager.
+func New() Manager {
```
negz, line 46: Does this need to return an interface?

We do have a small handful of places in the codebase where we need to return an interface, generally I prefer the "accept interfaces, return structs" pattern. I'm not a fan of using "go to definition" in my IDE to find how a returned type works, only to find it's returning an interface and I need to dig further to find the real implementation.

pkg/conditions/manager.go
```diff
@@ -0,0 +1,79 @@
...
+	return &managerImpl{}
+}
+
+// Check that conditionsImpl implements ConditionManager.
+var _ Manager = (*managerImpl)(nil)
+
+// managerImpl is the top level factor for producing a ConditionSet on behalf of a ObjectWithConditions resource.
+// managerImpl implements Manager.
+type managerImpl struct{}
```
negz, line 55: Nit: I'm not a fan of `Foo` interface with a `fooImpl`. I'd rather try to find some way to describe the implementation even if it's a bit contrived.

In this case, I wonder whether this package needs to define the `condition.Manager` interface? I could imagine this package just defining a concrete `condition.Manager` implementation, with callers defining the interface they want to work with if/as needed.

Historically we have defined a bunch of interfaces in crossplane-runtime, but I've been trying to move toward callers defining the interfaces they need.

pkg/conditions/manager.go
```diff
@@ -0,0 +1,79 @@
...
+	// For returns an implementation of a ConditionSet to operate on a specific ObjectWithConditions.
+	For(o ObjectWithConditions) ConditionSet
+}
+
+// ConditionSet holds operations for interacting with an object's conditions.
+type ConditionSet interface {
+	// Mark adds or updates the conditions onto the managed resource object. Unlike a "Set" method, this also can add
+	// contextual updates to the condition such as propagating the correct observedGeneration to the conditions being
+	// changed.
+	Mark(condition ...xpv1.Condition)
```
negz, line 42: WDYT about `MarkConditions`?

While I generally like the UX for this, looking at e.g. https://github.com/crossplane/crossplane-runtime/pull/828/files#diff-97300a2543f95f5a2ada3560bf47dd7334e237e27976574d15d1cddef2e66c01L883 `status.MarkConditions(Available())` would give me more of a hint to what it's doing than `status.Mark(Available()`.

I think I'd prefer to just stick with `SetConditions` despite the extra magic happening behind the scenes, but I don't feel as strongly about that.

pkg/conditions/manager.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+// New returns an implementation of a Manager.
+func New() Manager {
+	return &managerImpl{}
+}
+
+// Check that conditionsImpl implements ConditionManager.
+var _ Manager = (*managerImpl)(nil)
```
negz, line 51: Nit: We put these checks in `_test.go` files by convention, since they're essentially a kind of test.

pkg/conditions/manager.go
negz: Same question for `ConditionSet`. How likely is it that there'll be multiple meaningfully different `ConditionSet` implementations?

pkg/conditions/manager_test.go
```diff
@@ -0,0 +1,136 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := New(); !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("New() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
+
+func Test_conditionSet_Mark(t *testing.T) {
```
negz, line 49: Nit: The prevalent pattern in our tests is to use names like `TestConditionSetMark` - no underscores in test names.

To be fair this is pure taste, but I'd rather stick to the established pattern. I've always found it a bit strange that Go style avoids underscores in names, except for test function names where they're fine. 🤔

pkg/conditions/manager_test.go
```diff
@@ -0,0 +1,136 @@
...
+	manager := New()
+
+	tests := []struct {
+		name  string
+		start []xpv1.Condition
+		mark  []xpv1.Condition
+		want  []xpv1.Condition
+	}{{
+		name:  "provide no conditions",
```
negz, line 58: Nit: Established pattern is to use a `CamelCase` subtest name and include a human-readable `reason` string that's printed if the test fails.

We also tend to use a `map[string]struct{}` where the string is the test name, but less worried about that.

I find this makes the output a lot easier to read when the test fails. Using the names to provide context like this Go'll mangle them to `provide_no_conditions`, `existing_status_a_new_condition_is_appended` etc. Not so bad here but it's not great when your test descriptions get a little longer.

negz (commented):
Lots of nits but I like the overall direction. Thanks!

pkg/conditions/manager.go
n3wscott: There will be multiple meaningful implementations of ConditionSet and Manager in the fullness of time and it also allows for downstream testing in an easier way because this has to be in the hot path of all interactions. For example, it allows for a `Recorder` like interface to validate that a reconciler does attempt to call the correct mark methods.

pkg/conditions/manager.go
n3wscott: I will change this to `MarkConditions`, I think something that is not `Set` makes it clear there is magic happening or at least clear it is not a normal `Getter/Setter` pattern.

pkg/conditions/manager.go
n3wscott: Interesting, I use this for two things: validating that I have implemented the interface, and asking my IDE to implement the shape of the missing methods. But I can move them.

pkg/conditions/manager_test.go
n3wscott: This is what goland generates and I think this style is to be able to see if the method you are testing is private or public, like if you had:

```
func (*Foo) Mark() {}
func (*Foo) mark() {}

// and maybe also

func (*foo) Mark() {}
func (*foo) mark() {}

```

To test these with the suggested style would be:

```
TestFooMark
TestFoomark
TestfooMark
Testfoomark
```

vs how goland will generate it:

```
Test_Foo_Mark
Test_Foo_mark
Test_foo_Mark
Test_foo_mark
```

In this style you know exactly what is public or private without conflicts.

pkg/conditions/manager_test.go
n3wscott: I gave it a try to implement as you mentioned

pkg/conditions/manager.go
negz: I'm wondering if we should rip the bandaid and go straight to the full manager implementation after all. I'm still skeptical/reluctant about returning interfaces here, and it's hard to judge how to proceed without understanding the final destination.

pkg/conditions/manager_test.go
negz: I'd prefer to stick to the prevalent pattern in the codebase. I see the rationale but AFAIK we haven't had a meaningful issue with conflicts/readability so far.

Generally - I'm open to changing the prevailing style if there's a strong reason to. This seems like a non-zero reason but I'm not convinced it's worth the churn to adopt.

pkg/conditions/manager.go
n3wscott: I think the real value is for testing integrators of the manager. The patterns I have used in other repos are that this manager (which is a factory) is built up with the conditions and other settings that apply to all kinds for this reconciler implementation. the .`For` method gives you an instance of a condition mutator that allows you make changes to your conditions. Because it is so implementation specific to how the controllers work I think it would be best to stick with the interface/factory pattern for future improvements and testability. The current crossplane controllers do not use controllers in a standard way, but it is self consistent so there are at least two ways to deal with conditions out of the box. 

I also worry that the current code only updates a condition if it does work, which is also non-standard. The assumption is that every condition should be updated for every reconciliation even if there is no meaningful change, and then you don't update just time. Which means today as we add ObservedGeneration, and we have `Synced` and `Ready`, I am going to bet that `Ready` in the way it was described in a linked issue will lag behind in `ObservedGeneration` vs `Generation`. So a non-standard implementation of the ConditionSet we might need to implement is to update all other conditions (we know there are only two) to the current Generation on each `Mark`.

I agree to keep it simple, but I also think this is not a simple situation and and having the flexibility to add temporary implementations of the manager and condition set components without the downstream consumers needing to understand this deeply will let us move faster.

pkg/conditions/manager_test.go
n3wscott: ok updated.

pkg/conditions/manager_test.go
```diff
@@ -0,0 +1,133 @@
...
+			ut := newManaged(42, tt.start...)
+			c := manager.For(ut)
+			c.MarkConditions(tt.mark...)
+			if diff := cmp.Diff(tt.want, ut.Conditions, test.EquateConditions(), cmpopts.EquateApproxTime(1*time.Second)); diff != "" {
+				t.Errorf("\nReason: %s\n-want, +got:\n%s", tt.reason, diff)
+			}
+		})
+	}
+
+	t.Run("Manage a nil object", func(t *testing.T) {
```
negz, line 89:
```suggestion
	t.Run("ManageNilObject", func(t *testing.T) {
```

Nit - same naming rationale as the entries in the map above.

negz (approved):
Thanks @n3wscott! Excited to see where this goes.

Issue #340, opened 2022-07-14
Workspace resource is stuck with external-create-pending annotation if provider-terraform pod is deleted during Create


bobh66:
### What happened?
The provider-terraform Create() implementation runs terraform apply which can take "a while" - certainly tens of seconds and often several minutes.

If the provider pod is terminated while terraform apply is running, the external-create-pending annotation can get stuck on the Workspace resource and the managed Reconciler will not process it without manual intervention to remove the annotation.

This is most easily demonstrated when provider-terraform has code that detected the context cancellation and returns failure from Create():

https://github.com/crossplane-contrib/provider-terraform/pull/76

This should cause the Reconciler to set the ExternalCreateFailed annotation, which would allow the resource to be retried on the next cycle.   However the update of the resource to set that annotation fails because the context has already been cancelled, and so the resource is stuck with the pending annotation.

The logs indicate that the critical annotations cannot be updated because the context has been canceled:

```
[...]

PR #841, opened 2025-05-28
Avoid non-requeue for MRs with determinstic external names.

nolancon:
- Add a new ~~Annotation `crossplane.io/external-name-is-deterministic`~~ `WithDeterministicExternalName` reconciler option which allows the user to declare that a given resource is deterministically named by the external system.
- For MRs which are deterministically named, ignore the safeguard against leaked resources.

These changes enables the user to essentially "opt-out" of the current behaviour which blocks re-queuing of MRs in a scenario where creation is deemed incomplete. This safeguard is perfectly justified for MRs which are given non-deterministic names by the external system, preventing leaked resources (this scenario is described in detail [here](https://github.com/crossplane/crossplane/issues/3037#issuecomment-1110142427)). 

However, for resources which **are** deterministically named by the external system, the decision to not re-queue the MR is unnecessary as these resources are not in danger of leaking. For instance, [provider-ceph](https://github.com/linode/provider-ceph) creates S3 buckets by reconciling Bucket MRs - these s3 buckets are deterministically named and created idempotently. Therefore there is no danger in re-queuing after an initial "incomplete creation" failure (in fact it is expected). This kind of failure can (and does) happen regularly and in a system with 1000s of buckets leads to dozens of MRs which are not re-queued and therefore left in this pending state requiring manual intervention.

I have tested this patched version of the runtime against provider ceph. More details can be seen on this ~~[demo PR](https://github.com/linode/provider-ceph/pull/329)~~ [demo PR](https://github.com/linode/provider-ceph/pull/333) which also contains details on the existing workaround we use to attempt to mitigate the above scenario (but this is only best-effort and does not eradicate the problem completely).

Also, I'm happy to open a docs PR on this but would like to get feedback on this approach first, thanks.

Fixes # 
https://github.com/crossplane/crossplane-runtime/issues/340 (maybe not fixes, but allows an "opt-out")

I have: 


negz:
Thanks for taking a shot at this! I was thinking about something similar.

Could this be configured at the entire-type level though? I assume the if one `kind: Widget` has a deterministic external name, that'll be true for all `kind: Widget` resources, right? Ideally you wouldn't need to add this annotation to all of them.

Configuring this at the type level feels like the right way to go, but it's harder. It seems like a good use for the ManagedResourceDefinition type proposed in https://github.com/crossplane/crossplane/issues/2262.

negz:
Actually I guess this could just be handled at the code level via a new option to the MR reconciler. That'd mean provider authors would need to adopt it rather than being something end-users could specify though.

nolancon:
@negz I reverted the earlier changes and added a new commit with the ReconcilerOption approach to make the diff more readable. Here is an example PR for how this can be implemented by a downstream provider https://github.com/linode/provider-ceph/pull/333

pkg/reconciler/managed/reconciler.go
```diff
@@ -941,9 +955,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (resu
 
 	// If we started but never completed creation of an external resource we
 	// may have lost critical information. For example if we didn't persist
-	// an updated external name we've leaked a resource. The safest thing to
-	// do is to refuse to proceed.
-	if meta.ExternalCreateIncomplete(managed) {
+	// an updated external name which is non-deterministic, we have leaked a
+	// resource. The safest thing to do is to refuse to proceed. However, if
+	// the resource has a deterministic external name, it is safe to proceed.
+	if meta.ExternalCreateIncomplete(managed) && !r.deterministicExternalName {
```
negz: WDYT about debug logging this case (but not returning an error) even when there's a deterministic external name?

negz (approved):
One suggestion but this LGTM. Thank you!

pkg/reconciler/managed/reconciler.go
nolancon: Sure, something like this https://github.com/crossplane/crossplane-runtime/pull/841/commits/6120cd37b63623eae817c75fc6efe893288621f6?

negz: Yes that looks great.

PR #843, opened 2025-06-06
feat: enable namespaced reference resolution

jbw976:

This PR updates the Resolve methods to take into a namespace when resolving references. This is needed for v2 compatible providers that have namespaced managed resources.

I considered updating [`ResolutionRequest.Reference`](https://github.com/crossplane/crossplane-runtime/blob/main/pkg/reference/reference.go#L187) itself to become a reference type that includes a namespace field, but I decided against that for the following reasons:

* it's not needed - cluster scoped legacy MRs refer to other cluster scoped MRs and don't need to specify a namespace in the reference itself. namespaced MRs refer to other namespaced MRs and don't need to specify a namespace in the reference either - it's inferred from the namespace of the MR.
* it would be a (semi) breaking change to add a new field to that type that is used all over the crossplane ecosystem

This PR is against `main` since Crossplane core development has moved onto v2 now.

Part of https://github.com/crossplane/crossplane/issues/6381

Related PR: https://github.com/crossplane/crossplane-tools/pull/110 

These changes have been tested as demonstrated in this demo repo: https://github.com/jbw976/demo-upjet-namespaced-refs

I have: 

### DIFF
--- a/pkg/reference/reference.go
+++ b/pkg/reference/reference.go
@@ -190,6 +190,7 @@ type ResolutionRequest struct {
 	Selector     *xpv1.Selector
 	To           To
 	Extract      ExtractValueFn
+	Namespace    string
 }
 
 // IsNoOp returns true if the supplied ResolutionRequest cannot or should not be
@@ -244,6 +245,7 @@ type MultiResolutionRequest struct {
 	Selector      *xpv1.Selector
 	To            To
 	Extract       ExtractValueFn
+	Namespace     string
 }
 
 // IsNoOp returns true if the supplied MultiResolutionRequest cannot or should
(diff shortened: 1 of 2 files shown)


pkg/reference/reference.go
```diff
@@ -335,7 +337,7 @@ func (r *APIResolver) Resolve(ctx context.Context, req ResolutionRequest) (Resol
 	}
 
 	// The reference was not set, but a selector was. Select a reference.
-	if err := r.client.List(ctx, req.To.List, client.MatchingLabels(req.Selector.MatchLabels)); err != nil {
+	if err := r.client.List(ctx, req.To.List, client.MatchingLabels(req.Selector.MatchLabels), client.InNamespace(req.Namespace)); err != nil {
```
negz, line 343: Might be worth adding a note that `InNamespace` is a no-op if the namespace is empty. I always find that ambiguous when reading it.

negz (approved):

pkg/reference/reference.go
jbw976: good call, just pushed an update to include a comment on this behavior!

Issue #842, opened 2025-06-06
[v2] Remove external secret store support


negz:
### What problem are you facing?

Per https://github.com/crossplane/crossplane/issues/5522 we're removing external secret store support in Crossplane v2. Support is already gone from core - XRs and claims can no longer write to external secret stores.

### How could Crossplane help solve your problem?

We should remove support from crossplane-runtime as well, which then implies removing it from the provider ecosystem as well.

PR #849, opened 2025-06-18
Backport c/c internal/xresource

negz:

Fixes https://github.com/crossplane/crossplane/issues/6505
Closes https://github.com/crossplane/crossplane-runtime/pull/844

This is my attempt to address https://github.com/crossplane/crossplane/issues/6505 without removing external secret store support completely - yet (see #842).

I have: 

### DIFF
--- a/pkg/resource/unstructured/composite/composite.go
+++ b/pkg/resource/unstructured/composite/composite.go
@@ -22,6 +22,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/utils/ptr"
 
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 	"github.com/crossplane/crossplane-runtime/pkg/errors"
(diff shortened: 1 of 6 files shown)


negz:
@phisco @jbw976 @bobh66 @turkenh - Could you prioritize reviewing this please. Mostly because if it looks good I'd like to get https://github.com/crossplane/crossplane/pull/6553 merged before I suffer too many merge conflicts. :)

pkg/resource/unstructured/composite/composite.go
```diff
@@ -60,110 +82,182 @@ func New(opts ...Option) *Unstructured {
 // +k8s:deepcopy-gen=true
 // +kubebuilder:object:root=true
 
-// An Unstructured composed resource.
+// An Unstructured composite resource.
 type Unstructured struct {
 	unstructured.Unstructured
+
+	Schema Schema
```
phisco, line 89: So, if I got this correctly, the default schema is modern, for people upgrading that would mean that if they don't explicitly set the legacy one everything will be broken in a not so obvious way. Would it make sense to switch the two?

phisco (approved):
one question about the breaking change, otherwise LGTM 🙏

pkg/resource/unstructured/composite/composite.go
negz: Hrm good point. I think I feel okay with this breaking change though. I imagine there's very few users of this package in practice, apart from c/c. The Go SDK for example has its own version. So I doubt anyone will break in practice.

PR #850, opened 2025-06-25
Better support for non-deterministic external-names by updating critical annotations

twobiers:

This PR ensures that critical annotations are stored when a resource is created. Some resources have non-deterministic external-names, which are never updated when the `mangementPolicies` don't contain `LateInitialize`.
So far, there is an implicit contract that an Observation updating the external-name will be eventually stored as part of the LateInitialize process. However, that should only affect updates to the `spec` and not the annotations like external-name.
More Context can be found here: https://github.com/crossplane/crossplane/issues/5918

I'm not sure how a unit test can be expressed, as I'm not really familiar with the setup. If the change is approved, I think it would make sense to backport it aswell.

Fixes:
- https://github.com/crossplane/crossplane/issues/5918
- https://github.com/crossplane-contrib/provider-upjet-azuread/issues/159
- https://github.com/crossplane-contrib/provider-upjet-aws/issues/1567
- https://github.com/crossplane-contrib/provider-upjet-azure/issues/802

Maybe fixes aswell:
- https://github.com/crossplane-contrib/provider-upjet-aws/issues/1482

I have: 

- [ ] ~Added or updated unit tests.~ 

### DIFF
--- a/pkg/reconciler/managed/api.go
+++ b/pkg/reconciler/managed/api.go
@@ -277,15 +277,25 @@ func NewRetryingCriticalAnnotationUpdater(c client.Client) *RetryingCriticalAnno
 // UpdateCriticalAnnotations updates (i.e. persists) the annotations of the
 // supplied Object. It retries in the face of any API server error several times
 // in order to ensure annotations that contain critical state are persisted.
-// Pending changes to the supplied Object's spec, status, or other metadata
-// might get reset to their current state according to the API server, e.g. in
-// case of a conflict error.
+// Only annotations will be updated as part of this operation, other fields of the
+// supplied Object will not be modified.
 func (u *RetryingCriticalAnnotationUpdater) UpdateCriticalAnnotations(ctx context.Context, o client.Object) error {
 	a := o.GetAnnotations()
 	err := retry.OnError(retry.DefaultRetry, func(err error) bool {
 		return !errors.Is(err, context.Canceled)
 	}, func() error {
-		err := u.client.Update(ctx, o)
+		patchMap := map[string]interface{}{
+			"metadata": map[string]any{
+				"annotations": a,
+			},
+		}
+
+		patchData, err := json.Marshal(patchMap)
+		if err != nil {
+			return err
+		}
+
+		err = u.client.Patch(ctx, o, client.RawPatch(types.MergePatchType, patchData), client.FieldOwner(fieldOwnerAPISimpleRefResolver), client.ForceOwnership)
 		if kerrors.IsConflict(err) {
 			if getErr := u.client.Get(ctx, client.ObjectKeyFromObject(o), o); getErr != nil {
 				return getErr
--- a/pkg/reconciler/managed/api.go
+++ b/pkg/reconciler/managed/api.go
@@ -55,6 +55,15 @@ const (
 	errUpdateCriticalAnnotations = "cannot update critical annotations"
 )
 
+var (
+	criticalAnnotations = []string{
+		meta.AnnotationKeyExternalCreateFailed,
+		meta.AnnotationKeyExternalCreatePending,
+		meta.AnnotationKeyExternalCreateSucceeded,
+		meta.AnnotationKeyExternalName,
+	}
+)
+
 // NameAsExternalName writes the name of the managed resource to
 // the external name annotation field in order to be used as name of
 // the external resource in provider.
@@ -280,7 +289,18 @@ func NewRetryingCriticalAnnotationUpdater(c client.Client) *RetryingCriticalAnno
 // Only annotations will be updated as part of this operation, other fields of the
 // supplied Object will not be modified.
 func (u *RetryingCriticalAnnotationUpdater) UpdateCriticalAnnotations(ctx context.Context, o client.Object) error {
-	a := o.GetAnnotations()
+	a := make(map[string]string)
+	for _, k := range criticalAnnotations {
+		if v, ok := o.GetAnnotations()[k]; ok {
+			a[k] = v
+		}
+	}
+
+	if len(a) == 0 {
+		// No critical annotations to update.
+		return nil
+	}
+
 	err := retry.OnError(retry.DefaultRetry, func(err error) bool {
 		return !errors.Is(err, context.Canceled)
 	}, func() error {
--- a/pkg/reconciler/managed/api.go
+++ b/pkg/reconciler/managed/api.go
@@ -55,6 +55,7 @@ const (
 	errUpdateCriticalAnnotations = "cannot update critical annotations"
 )
 
+//nolint:gochecknoglobals // this is a list of critical annotations that should be retried when updating, and is not expected to be modified at runtime.
 var (
 	criticalAnnotations = []string{
 		meta.AnnotationKeyExternalCreateFailed,
@@ -304,7 +305,7 @@ func (u *RetryingCriticalAnnotationUpdater) UpdateCriticalAnnotations(ctx contex
 	err := retry.OnError(retry.DefaultRetry, func(err error) bool {
 		return !errors.Is(err, context.Canceled)
 	}, func() error {
-		patchMap := map[string]interface{}{
+		patchMap := map[string]any{
 			"metadata": map[string]any{
 				"annotations": a,
 			},
--- a/pkg/reconciler/managed/reconciler.go
+++ b/pkg/reconciler/managed/reconciler.go
@@ -1412,8 +1412,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (resu
 
 	if observation.ResourceExists {
 		// When a resource exists or is just created, it might have received
-		// a non-deterministic external name, which we need to persist.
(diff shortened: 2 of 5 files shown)


twobiers:
@jeanduplessis Thanks for your kind comment. I'm able to work on this and have rebased the branch onto master.

pkg/reconciler/managed/reconciler.go
```diff
@@ -1408,6 +1408,21 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (resu
...
 
+	if observation.ResourceExists {
+		// When a resource exists or is just created, it might have received
+		// a non-deterministic external name after its creation, which we need to persist.
+		// We do this by updating the critical annotations.
+		// This is needed because some resources might not receive an external-name directly 
+		// after the creation, but later as part of an asynchronous process.
+		// When Crossplane supports asynchronous creation of resources natively, this logic
+		// might not be needed anymore and can be revisited. 
+		if err := r.managed.UpdateCriticalAnnotations(ctx, managed); err != nil {
```
erhancagirici, line 1421: a thing to consider (possible nit):
most reconciliation loops will enter this codepath regardless of a need to update critical annotations. I wonder if this one is no-op in terms of k8s API access or brings some extra load on the apiserver.

Especially in async creations with long-running creation times, currently (with no native async support) the external clients return `observation.ResourceExists` as `true` to avoid further actions for the observations during creation of the resource.

pkg/reconciler/managed/api.go
```diff
@@ -277,15 +277,25 @@ func NewRetryingCriticalAnnotationUpdater(c client.Client) *RetryingCriticalAnno
...
+				"annotations": a,
+			},
+		}
+
+		patchData, err := json.Marshal(patchMap)
+		if err != nil {
+			return err
+		}
+
+		err = u.client.Patch(ctx, o, client.RawPatch(types.MergePatchType, patchData), client.FieldOwner(fieldOwnerAPISimpleRefResolver), client.ForceOwnership)
```
erhancagirici, line 319: Looks like we patch all annotations here.
Just wondering how does this interact with the field ownerships of annotations that XP does not manage (like a custom annotation set by a user or some other controllers, tooling etc), and whether it is a thing to worry about regarding server-side apply etc. Would we cause a race regarding field ownerships?

pkg/reconciler/managed/api.go
```diff
@@ -277,15 +277,25 @@ func NewRetryingCriticalAnnotationUpdater(c client.Client) *RetryingCriticalAnno
...
-// might get reset to their current state according to the API server, e.g. in
-// case of a conflict error.
+// Only annotations will be updated as part of this operation, other fields of the
+// supplied Object will not be modified.
 func (u *RetryingCriticalAnnotationUpdater) UpdateCriticalAnnotations(ctx context.Context, o client.Object) error {
 	a := o.GetAnnotations()
 	err := retry.OnError(retry.DefaultRetry, func(err error) bool {
 		return !errors.Is(err, context.Canceled)
 	}, func() error {
-		err := u.client.Update(ctx, o)
```
ulucinar, line 288: I like this update operation is being replaced by SSA.

pkg/reconciler/managed/api.go
ulucinar, line 319: We had better use a different manager name than `managed.crossplane.io/api-simple-reference-resolver` as the annotations have nothing to do with the API resolver. Maybe something like: `managed.crossplane.io/critical-annotation-updater`?

pkg/reconciler/managed/api.go
```diff
@@ -277,15 +277,25 @@ func NewRetryingCriticalAnnotationUpdater(c client.Client) *RetryingCriticalAnno
...
+		patchMap := map[string]interface{}{
+			"metadata": map[string]any{
+				"annotations": a,
```
ulucinar, line 310: Should we be more selective in what's being patched here, i.e., not all the annotations on the MR are the critical ones and we would only like to manage the critical annotations by this manager? This will probably not be an issue as this manager will not have an opinion on "non-critical" annotations and their respective managers will dictate their values. But one potential issue is when the other manager (who should really be owning the annotation) actually wants to delete the annotation. The `managed.crossplane.io/critical-annotation-updater` will still be owning the annotation. So the critical-annotation-updater had better not own non-critical ones...

erhancagirici (commented):
Thanks for the PR! I've left some comments regarding the mechanics, otherwise LGTM conceptually. As you mentioned in the doc comments, this will be sort of a band-aid until a more async-aware implementation.

pkg/reconciler/managed/reconciler.go
```diff
@@ -1408,6 +1408,21 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (resu
...
+			log.Debug(errUpdateManagedAnnotations, "error", err)
+			record.Event(managed, event.Warning(reasonCannotUpdateManaged, errors.Wrap(err, errUpdateManagedAnnotations)))
+			return reconcile.Result{Requeue: true}, errors.Wrap(r.client.Status().Update(ctx, managed), errUpdateManagedAnnotations)
+		}
+	}
```
ulucinar, line 1426: As @lsviben explains [here](https://github.com/crossplane/crossplane/issues/5918#issuecomment-2326716642), upjet should not be relying on setting `ResourceLateInitialized` to get the critical annotations updated in the first place. The async mode implemented by upjet breaks the assumptions of the managed reconciler. I believe we need to first address this discrepancy between upjet and the managed reconciler...

ulucinar (commented):
Thanks @twobiers for working on this issue. Left some comments for us to discuss.

pkg/reconciler/managed/api.go
negz: Would it make sense for the MR reconciler to just use one manager name, regardless of operation?

Possibly too late if we're already using `api-simple-reference-resolver` for some.

pkg/reconciler/managed/reconciler.go
negz: Perhaps we should make writing these annotations optional. I'd suggest making them opt-out since that's the safest path. The option would be specified by the provider author - so they can disable these annotations if they know for sure they're not needed (i.e. naming is deterministic, API is strongly consistent). I remember discussing this with someone recently, but can't find a tracking issue.

(If we do make them optional, I think we could do it pretty easily by injecting a no-op implementation of the annotation updater.)

pkg/reconciler/managed/api.go
twobiers: I think it would be a reasonable addition for this function to have knowledge about what is a "critical annotation". Currently that comes from the execution order of the reconciler, this way we would make it explicit.
From what I can see, the ciritcal annotations should be:
- `crossplane.io/external-name`
- `crossplane.io/external-create-pending`
- `crossplane.io/external-create-succeeded`
- `crossplane.io/external-create-failed`

Am I missing any?

pkg/reconciler/managed/api.go
twobiers: See https://github.com/crossplane/crossplane-runtime/pull/850/files#r2585575881

pkg/reconciler/managed/reconciler.go
twobiers: It should also be straightforward to keep a state in the reconciler whether a critical annotation was added/changed and depending on that update the annotations or perform a no-op.

On the other hand I'm wondering whether this kind of optimization is needed for "Critical" annotations. Shouldn't the priority be here to add those annotations? Otherwise it would leave room for errors.

pkg/reconciler/managed/api.go
twobiers: I've added a very naive approach in 11e21362a81ac917efc056dfb588942044003926

twobiers:
I just had the idea that a function in the pipeline could also ensure those annotations are persisted. It would not be an ideal solution, but at least a possible hotfix while async is not fully supported.

jeanduplessis:
@bobh66 this PR now needs a review and approval from a @crossplane/crossplane-maintainers to move it forward.

jwefers:
Does this fix https://github.com/crossplane/upjet/issues/632 ?

bobh66:
@twobiers can you check the lint and unit-test errors?  I'd like to get this moving if there are no other solutions in the works, and I don't think there are.

twobiers:
@bobh66 I've rebased the branch and fixed the issues in the checks. They do pass locally now. Not sure if you have to kick off the workflows here again manually.
Let me know if you have any further comments on the change

niraj1201:
Hello, any idea when this fix will be applied? Also any workaround until this is fixed. We have paused the sync by adding crossplane.io/paused=true since CertificateAuthority on acmpca MR is creating duplicate privateCAs because of missing external-name.

jwefers:
> Hello, any idea when this fix will be applied? Also any workaround until this is fixed. We have paused the sync by adding crossplane.io/paused=true since CertificateAuthority on acmpca MR is creating duplicate privateCAs because of missing external-name.

if you know the upstream id, you can manually set the external-name annotation to lock it in

PR #854, opened 2025-07-08
add v2 apis

erhancagirici:

New V2 apis and removals to facilitate namespaced MRs.

**v2 `ManagedResourceSpec` API:**
  - `writeConnectionSecretToRef` drops `namespace` field and accepts local secret references.
  - `providerConfigRef` becomes typed and requires `kind`. `apiGroup` is intentionally left out, it is expected to be inferred by the provider code. Namespaced
  - `deletionPolicy` is removed. New resources should rely on `managementPolicies` to control external resource deletion.

**External Secret Store support is removed**
 `publishConnectionDetailsTo` is removed from both cluster-scoped (legacy) and namespace-scoped (modern) MR specs.
`ConnectionDetailsPublisherTo` interface is removed.

**Cross-namespace cross-resource references:**
Introduced `NamespacedResolutionRequest`, `NamespacedResolutionResponse`, `NamespacedMultiResolutionRequest`, `NamespacedMultiResolutionResponse` types along with `APINamespacedResolver` . In the new types, references and selectors in the resolution requests can include optional namespace parameter. Otherwise, they will default to MR namespaces.

I have: 


apis/common/v2/resource.go
```diff
@@ -0,0 +1,55 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v2
+
+import v1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
```
negz: Is there a way to avoid having the v2 package depend on v1 types?

pkg/reconciler/managed/api.go
```diff
@@ -130,6 +130,63 @@ func (a *APISecretPublisher) UnpublishConnection(_ context.Context, _ resource.C
...
+
+	return true, nil
+}
+
+// UnpublishConnection is no-op since PublishConnection only creates resources
+// that will be garbage collected by Kubernetes when the managed resource is
+// deleted.
+func (a *APILocalSecretPublisher) UnpublishConnection(_ context.Context, _ resource.LocalConnectionSecretOwner, _ ConnectionDetails) error {
+	return nil
+}
```
negz, line 194: iirc this'll always be a no-op now that external secret stores are gone.

No strong feelings about removing it now though - we could follow up later.

pkg/reconciler/managed/publisher.go
```diff
@@ -55,24 +54,32 @@ func (pc PublisherChain) UnpublishConnection(ctx context.Context, o resource.Con
...
+type LocalPublisherChain []LocalConnectionPublisher
 
-// PublishConnection returns a proper error when API used but the feature was
-// not enabled.
-func (m *DisabledSecretStoreManager) PublishConnection(_ context.Context, so resource.ConnectionSecretOwner, _ ConnectionDetails) (bool, error) {
-	if so.GetPublishConnectionDetailsTo() != nil {
-		return false, errors.New(errSecretStoreDisabled)
+// PublishConnection calls each ConnectionPublisher.PublishConnection serially. It returns the first error it
+// encounters, if any.
+func (pc LocalPublisherChain) PublishConnection(ctx context.Context, o resource.LocalConnectionSecretOwner, c ConnectionDetails) (bool, error) {
```
negz: Do we still need this with external secret stores gone? I think we'll always publish to one secret.

(Again not urgent to remove it though.)

pkg/reconciler/managed/reconciler.go
```diff
@@ -579,10 +595,30 @@ func defaultMRManaged(m manager.Manager) mrManaged {
...
+
+func (m mrManaged) PublishConnection(ctx context.Context, managed resource.Managed, c ConnectionDetails) (bool, error) {
+	switch so := managed.(type) {
+	case resource.LocalConnectionSecretOwner:
+		return m.LocalConnectionPublisher.PublishConnection(ctx, so, c)
+	case resource.ConnectionSecretOwner:
+		return m.ConnectionPublisher.PublishConnection(ctx, so, c)
+	default:
+		return false, errors.New(errManagedNotImplemented)
+	}
```
negz, line 628: Nice, I like this pattern.

pkg/resource/interfaces.go
```diff
@@ -192,14 +195,21 @@ type Object interface {
...
+type LegacyManaged interface {
+	Managed
 	ConnectionSecretWriterTo
-	ConnectionDetailsPublisherTo
-	Manageable
+	ProviderConfigReferencer
 	Orphanable
-
-	Conditioned
 }
```
negz: Naming nit: I don't love `ModernManaged` but I also can't think of a better alternative. Only thing that comes to mind would be:

* `BaseManaged`
* `Managed`
* `LegacyManaged`

negz (commented):
This looks great overall - I like the direction. Thanks @erhancagirici!

pkg/resource/interfaces.go
erhancagirici: I also do not like this, however having the "base managed" named other than `Managed` requires renaming in lots of places (this repo itself, upjet, crossplane-tools) and in providers themselves. 
This way, only ProviderConfig-related codepaths needs to adapt.
LMK if this is not enough to justify this decision and we should go for renaming this.

I think eventually `ModernManaged` and `Managed` will converge and become aliases, in a future major version bump.

pkg/reconciler/managed/publisher.go
erhancagirici: Agreed, this was needed for the `WithLocalConnectionPublishers(p ...ConnectionPublisher)`  reconciler option, which is already irrelevant.

A question is, the original(legacy) one of becomes irrelevant too. Should we remove this one to make our intention clear? or keep it for the sake of not causing a code changes in the (native) providers for existing legacy resources?
https://github.com/crossplane/crossplane-runtime/blob/f5f608c93ded2ef2ea30daa38a6fd7aaed53efb2/pkg/reconciler/managed/reconciler.go#L702-L706
Maybe make it no-op with deprecated comment?

apis/common/v2/resource.go
erhancagirici: Technically yes, but I see two ways, both having its own drawback:
most of `apis/common/v1` actually has many shared constructs not particularly "v1", like ManagementPolicies, Conditions, Selector, Reference types etc.
- If we duplicate them, we end up with same but different structs, which needs superfluous handling in various places.
- One option can be to refactor common stuff mentioned above to its own package, but that also kind of breaks existing imports in the ecosystem.
Open to ideas here

apis/common/v1/resource.go
```diff
@@ -124,6 +148,19 @@ type Reference struct {
...
 }
 
+// A NamespacedReference to a named object.
+type NamespacedReference struct {
+	// Name of the referenced object.
+	Name string `json:"name"`
+
+	// Namespace of the referenced object
+	// +optional
+	Namespace string `json:"namespace,omitempty"`
```
negz: Nit: I'd prefer consistent whitespace between struct fields.

pkg/reconciler/managed/policies.go
```diff
@@ -118,6 +126,25 @@ func NewManagementPoliciesResolver(managementPolicyEnabled bool, managementPolic
 	return r
 }
 
+// NewLegacyManagementPoliciesResolver returns an ManagementPolicyChecker based
+// on the management policies and if the management policies feature
+// is enabled.
+// Deprecated: this is intended for LegacyManaged resources that had deletionPolicy
+// ModernManaged resources should use NewManagementPoliciesResolver.
+func NewLegacyManagementPoliciesResolver(managementPolicyEnabled bool, managementPolicy xpv1.ManagementPolicies, deletionPolicy xpv1.DeletionPolicy, o ...ManagementPoliciesResolverOption) ManagementPoliciesChecker {
```
negz, line 134: Nit: Some of these long var names are a bit redundant. Not blocking though, and I can see it matches the existing implementation.

pkg/reconciler/managed/publisher.go
negz: I'd lean toward just removing it. Either way it's going to be work for upstream provider authors. If it's gone, it's more obvious they need to do that work.

I don't feel strongly though.

pkg/reference/namespaced_reference.go
```diff
@@ -0,0 +1,312 @@
...
+	valueMap := make(map[string]xpv1.NamespacedReference)
+
+	// The references are already set - resolve them.
+	if len(req.References) > 0 {
+		for i := range req.References {
+			ns := req.References[i].Namespace
+			if ns == "" {
+				ns = r.from.GetNamespace()
+			}
+			// TODO(erhan): check ns
```
negz: Is this TODO still relevant? I see the namespace in the Get call.

pkg/reference/namespaced_reference.go
```diff
@@ -0,0 +1,312 @@
...
+		}
+
+		rsp := NamespacedResolutionResponse{ResolvedValue: req.Extract(req.To.Managed), ResolvedReference: req.Reference}
+
+		return rsp, getResolutionError(req.Reference.Policy, rsp.Validate())
+	}
+
+	// The reference was not set, but a selector was. Select a reference. If the
+	// request has no namespace, then InNamespace is a no-op.
+	// TODO(erhan): check ns
```
negz: Looks like this TODO might be stale as well?

pkg/resource/resource.go
```diff
@@ -51,6 +51,8 @@ const (
 	ExternalResourceTagKeyName     = "crossplane-name"
 	ExternalResourceTagKeyProvider = "crossplane-providerconfig"
```
negz: Nit: Should this whitespace be here? Looks like it should be grouped with the three consts above.

pkg/reference/namespaced_reference.go
```diff
@@ -0,0 +1,312 @@
...
+		sortedKeys, sortedRefs := sortGenericMapByKeys(valueMap)
+
+		rsp := MultiNamespacedResolutionResponse{ResolvedValues: sortedKeys, ResolvedReferences: sortedRefs}
+
+		return rsp, rsp.Validate()
+	}
+
+	// No references were set, but a selector was. Select and resolve
+	// references. If the request has no namespace, then InNamespace is a no-op.
+	// TODO(erhan): check ns
```
negz: Here too. :)

pkg/reconciler/managed/reconciler_modern_test.go
```diff
@@ -0,0 +1,2715 @@
...
+	"github.com/crossplane/crossplane-runtime/pkg/errors"
+	"github.com/crossplane/crossplane-runtime/pkg/meta"
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
+	"github.com/crossplane/crossplane-runtime/pkg/resource/fake"
+	"github.com/crossplane/crossplane-runtime/pkg/test"
+)
+
+var _ reconcile.Reconciler = &Reconciler{}
+
+func TestModernReconciler(t *testing.T) {
```
negz, line 46: I do wonder whether there's diminishing returns in completely duplicating the test file. I imagine a lot of these subtest scenarios use the same code regardless of whether it's a legacy or a modern MR, right?

It seems like having two test files could be a bit of a pain - we'll need to add and update them both whenever we update the tests. On the other hand it does seem like the most defensive/conservative approach to make sure both legacy and modern MRs work with all parts of the reconciler.

negz:
Officially deferring my reviewer powers to @turkenh if there's any other questions or concerns while I'm out on vacation. 😉

pkg/resource/interfaces.go
negz: FWIW I've warmed to `ModernManaged` and `LegacyManaged`. It's pretty close to what we did with XR - i.e. they have `SchemaModern` and `SchemaLegacy`.

apis/common/v2/resource.go
negz: Maybe we could:

* Move the shared stuff to `apis/common/common.go`
* Have both common/v1 and common/v2 import common
* Add [type aliases](https://go.dev/blog/alias-names) in the v1 package to avoid breaking folks

I think I prefer that approach. I agree that v1 is a bit of a misnomer in the first place.

The other options I considered are:

* Duplicate the types in both places. I like that this decouples the two APIs, but it forces us to have two implementations of any function that uses these types, which I think is a non-starter.
* Duplicate the types in both places, also add internal versions, update our functions to work on the internal versions, and generate code to translate between v1/v2 and internal. This seems like the best approach in theory, but it's a ton of work for these little utility types.

pkg/reconciler/managed/publisher.go
erhancagirici: I'll remove this and `PublisherChain` and keep an unexported version for unit testing purposes for cases we need to mock.

erhancagirici: done

pkg/reconciler/managed/reconciler_modern_test.go
erhancagirici: As you mentioned, this was mostly for being defensive. I tried to pinpoint the "common" tests ( relying only on the "base" managed functional), though ended up duplicating most already to cover some unvisited codepaths.
We might consider following up later with a refactored test setup, that runs the tests as a matrix for legacy vs modern maybe?

pkg/reconciler/managed/api.go
erhancagirici: let's have this as a follow-up

apis/common/v2/resource.go
turkenh: We just had an offline sync with @erhancagirici on this and the above proposal sounds good in general. We discussed some details like linter complaining about `common` as a package name and whether we should put a deprecation notice on v1 aliased types/consts/funcs.

At the end, we feel good about suppressing linter on package name and going with `common` as suggested. `apis/common` is there since the begining and we're fine for it as a package name. Regarding the deprecation notice, we believe it would be good not to force consumers to immediately switch to the new place to minimize the required changes for now. So, if linter complains and forces consumers to migrate, then we should better not do it now (and track with an issue to do it later). This is something @erhancagirici will check and see.

erhancagirici: I have pushed the changes with the type aliasing approach.
One side note: I had to bump the controller-tools to 0.18.0 to consume following fixes about type aliases.

https://github.com/kubernetes-sigs/controller-tools/pull/1061
https://github.com/kubernetes-sigs/controller-tools/pull/1122
This bump transitively upgraded `google.golang.org/protobuf` to `v1.36.5` which had some minor changes in the generated protobuf files (backward-compatible).

I haven't added the depreciation notices to aliased types in `apis/common/v1` yet, as they cause large amount of linter warnings. I'll create a follow-up issue for this.

.golangci.yml
```diff
@@ -219,6 +219,13 @@ linters:
           - staticcheck
         text: 'SA1019: .+ is deprecated: Use Composition Functions instead.'
 
+      # Some shared structs in apis/common/v1 are moved to
+      # apis/common. To preserve a backward-compatible directory structure
+      # package had to be named common, which we suppress.
+      - linters:
+          - revive
+        text: "var-naming: avoid meaningless package names"
+        path: apis/common
```
turkenh, line 228: Could we add a `//nolint` directive instead? https://github.com/crossplane/crossplane/tree/main/contributing#explain-nolint-directives

pkg/reference/namespaced_reference.go
```diff
@@ -0,0 +1,310 @@
...
+	}
+
+	// No references were set, but a selector was. Select and resolve
+	// references. If the request has no namespace, then InNamespace is a no-op.
+	ns := req.Selector.Namespace
+	if ns == "" {
+		ns = r.from.GetNamespace()
+	}
+
+	if err := r.client.List(ctx, req.To.List, client.MatchingLabels(req.Selector.MatchLabels), client.InNamespace(req.Namespace)); err != nil {
```
turkenh: Should this use `ns` instead of `req.Namespace` ?

turkenh (approved):
Looks good to me, thanks @erhancagirici 🙌

pkg/reference/namespaced_reference.go
erhancagirici: good catch, fixed

.golangci.yml
erhancagirici: opted for this, because had to repeat this for all the new files, and nolint directives were appearing at the very top and potentially in godoc. not strongly opinionated though

PR #855, opened 2025-07-08
Delay external delete call when other finalizers exist

### LINKED ISSUE #851: Delay deletion of external managed resource until all 3rd-party finalizers are removed
When a Managed Resource is marked for deletion the Managed Reconciler immediately calls the external.Delete() method to trigger deletion of the external managed resource.  This makes it impossible to perform any "pre-delete" actions such as taking a database backup before the resource has been deleted.

A common pattern in kubernetes is for controllers that have an interest in a resource they don't manage to place a finalizer on the resource so that they can be notified when the resource is deleted and perform some processing before the resource is garbage collected.  This works for native and "local" kubernetes resources that do not have external manifestations to consider as the important data is in the CR instance.

This pattern does not work for Crossplane MRs because the reconciler does not consider the existence of other finalizers before deleting the remote resource.

Crossplane could allow a similar pattern by delaying the call to external.Delete() until there are no finalizers on the MR other than the one Crossplane uses.  This would allow scenarios such as `Operation Functions` that want to perform some action prior to the external resource being deleted.

bobh66:
Delay the external.Delete() call on a Managed Resource if there are extra finalizers in addition to the one that Crossplane uses.  This will allow external controllers to complete their deletion handling before the external resource is deleted.

One example could be a database backup Operation Function that needs to execute when a Database MR is marked as deleted but before the actual external database is removed by the provider.

Additional testing is needed but I wanted to get the review started for any feedback.

Fixes #851 

I have: 

~- [ ] Linked a PR or a [docs tracking issue] to [document this change].~

### DIFF
--- a/pkg/reconciler/managed/reconciler.go
+++ b/pkg/reconciler/managed/reconciler.go
@@ -1163,6 +1163,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (resu
 	if meta.WasDeleted(managed) {
 		log = log.WithValues("deletion-timestamp", managed.GetDeletionTimestamp())
 
+		if len(managed.GetFinalizers()) > 1 {
+			// There are other controllers monitoring this resource so preserve the external instance
+			// until all other finalizers have been removed
+			log.Debug("Delay external deletion until all finalizers have been removed")
+			return reconcile.Result{Requeue: true}, nil
+		}
+
 		if observation.ResourceExists && policy.ShouldDelete() {
 			deletion, err := external.Delete(externalCtx, managed)
 			if err != nil {
--- a/pkg/reconciler/managed/reconciler_legacy_test.go
+++ b/pkg/reconciler/managed/reconciler_legacy_test.go
@@ -202,7 +202,51 @@ func TestReconciler(t *testing.T) {
 			},
 			want: want{result: reconcile.Result{Requeue: true}},
 		},
-		"ExternalCreatePending": {
+		"ExtraFinalizersDelayDelete": {
+			reason: "The existence of multiple finalizers should trigger a requeue after a short wait.",
+			args: args{
+				m: &fake.Manager{
+					Client: &test.MockClient{
+						MockGet: test.NewMockGetFn(nil, func(obj client.Object) error {
+							mg :=
+								asManaged(obj, 42)
+							mg.SetDeletionTimestamp(&now)
+							mg.SetDeletionPolicy(xpv1.DeletionDelete)
+							mg.SetFinalizers([]string{FinalizerName, "finalizer2"})
+							return nil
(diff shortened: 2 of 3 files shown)


bobh66:
Test Plan

I built a version of provider-kubernetes with this change applied to crossplane-runtime, and created an `Object` with a `Secret`.

I then manually added a finalizer to the `Object` to simulate an external controller such as an Operation Function that wants to be able to perform an operation after the `Object` has been marked for delete but before the external resource `Secret` has actually been deleted.

I then deleted the `Object` and confirmed that while it was marked for deletetion with the `metadata.deletionTimestamp` field, it did NOT delete the `Secret` and the `Object` was not garbage collected as long as the second finalizer was present on the `Object`.

I then removed the second finalizer from the `Object` and confirmed that the `Secret` was deleted and the `Object` was garbage collected.

jbw976 (approved):
This seems like a reasonable approach @bobh66 - it makes sense to honor any other finalizers that are on a MR which would indicate another entity still has work to do before the deletion can be processed.  Thanks for your testing effort too! 🙇‍♂️

pkg/reconciler/managed/reconciler.go
```diff
@@ -1163,6 +1163,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (resu
 	if meta.WasDeleted(managed) {
 		log = log.WithValues("deletion-timestamp", managed.GetDeletionTimestamp())
 
+		if len(managed.GetFinalizers()) > 1 {
+			// There are other controllers monitoring this resource so preserve the external instance
+			// until all other finalizers have been removed
+			log.Debug("Delay external deletion until all finalizers have been removed")
```
negz, line 1169: I wonder if we should emit an event in this scenario, to let folks know what's blocking the delete?

negz:
Sorry I got to this too late.

Honestly I was a bit wary initially because I was wary of Operations adding finalizers to things as a pattern in the first place. I've always thought it was best that controllers only place finalizers on the things they control (i.e. are controller ref of).

Otherwise you run into potential issues like:

1. Install Crossplane
2. Create Operations that add finalizers to arbitrary stuff
3. Uninstall Crossplane
4. Finalizers will never go away, you gotta clean them up manually

I spent a short amount of time researching this yesterday and couldn't find any explicit guidance though.

That all said, even if we didn't choose to allow Operations to add finalizers to MRs I can see the point of blocking the external delete until there are no finalizers.

pkg/reconciler/managed/reconciler.go
bobh66: I was thinking about that but my experience with `Events` has been that they can be unreliable compared to debug messages, so I went with the debug.  I can add an event if we think it's helpful.

PR #856, opened 2025-07-10
Adding Gate and a CustomResourceGate reconciler

n3wscott:

In support of https://github.com/crossplane/crossplane/pull/6577, this change introduces a method for downstream providers (or anything) to add this reconciler that will reconcile CRDs and track their state. Allowing (really any) controller code to register a callback when a set of GVKs has been seen as ready in the cluster. This will dramatically improve CRD race conditions in general.

The main point of this would be to allow our upjet providers to leverage the gate and reconciler to block until a set of kinds has been created and accepted before we attempt another action, like to start a reconciler.

Integration looks something like this:

```
	o := controller.Options{
	        ...
		Gate:                    new(gate.Gate[schema.GroupVersionKind]),
	}	

	if err := customresourcesgate.Setup(mgr, o); err != nil {
		setupLog.Error(err, "unable to create controller", "controller", "CustomResourceGate")
		os.Exit(1)
	}
	
	o.Gate.Register(func() {
		if err = (&custom.Reconciler{}).SetupWithManager(mgr); err != nil {
			setupLog.Error(err, "unable to create controller", "controller", "Custom")
			os.Exit(1)
		}
	}, xpv1.FunctionGroupVersionKind, xpv1.FunctionRevisionGroupVersionKind, customv1.SchemeGroupVersion.WithKind("MyCustomType"))
```

The above example will only start when it sees `Function`s, `FunctionRevision`s, and `MyCustomType`s are installed and ready to go in the cluster.

I have: 

### DIFF
--- a/pkg/controller/gate.go
+++ b/pkg/controller/gate.go
@@ -1,0 +1,16 @@
+package controller
+
+import (
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// A Gate is an interface to allow reconcilers to delay a callback until a set of GVKs are set to true inside the gate.
+type Gate interface {
+	// Register to call a callback function when all given GVKs are marked true. If the callback is unblocked, the
+	// registration is removed.
+	Register(callback func(), gvks ...schema.GroupVersionKind)
+	// True marks a given gvk true, and then looks for unlocked callbacks. Returns if there was an update or not.
+	True(gvk schema.GroupVersionKind) bool
+	// False marks a given gvk false. Returns if there was an update or not.
+	False(gkv schema.GroupVersionKind) bool
+}
--- a/pkg/controller/gate.go
+++ b/pkg/controller/gate.go
@@ -9,8 +9,7 @@ type Gate interface {
 	// Register to call a callback function when all given GVKs are marked true. If the callback is unblocked, the
 	// registration is removed.
 	Register(callback func(), gvks ...schema.GroupVersionKind)
-	// True marks a given gvk true, and then looks for unlocked callbacks. Returns if there was an update or not.
-	True(gvk schema.GroupVersionKind) bool
-	// False marks a given gvk false. Returns if there was an update or not.
-	False(gkv schema.GroupVersionKind) bool
+	// Set marks the associated condition to the given value. If the condition is already set as
+	// that value, then this is a no-op. Returns true if there was an update detected.
+	Set(gvk schema.GroupVersionKind, ready bool) bool
 }
--- a/pkg/gate/gate.go
+++ b/pkg/gate/gate.go
@@ -29,9 +29,13 @@ type Gate[T comparable] struct {
 	fns        []gated[T]
 }
 
+// gated is an internal tracking resource.
 type gated[T comparable] struct {
-	fn       func()
-	depends  []T
+	// fn is the function callback we will invoke when all the dependent conditions are true.
+	fn func()
+	// depends is the list of conditions this gated function is waiting on. This is an AND.
+	depends []T
+	// released means the gated function has been invoked and we can garbage collect this gated function.
 	released bool
 }
 
@@ -46,19 +50,9 @@ func (g *Gate[T]) Register(fn func(), depends ...T) {
 	g.process()
 }
 
-// True marks the associated condition as true. If the condition is already true, then this is a no-op.
-// Returns if there was an update detected. Thread safe.
-func (g *Gate[T]) True(condition T) bool {
-	return g.markCondition(condition, true)
-}
-
-// False marks the associated condition as false. If the condition is already false, then this is a no-op.
-// Returns if there was an update detected. Thread safe.
-func (g *Gate[T]) False(condition T) bool {
-	return g.markCondition(condition, false)
-}
-
-func (g *Gate[T]) markCondition(condition T, ready bool) bool {
+// Set marks the associated condition to the given value. If the condition is already set as that value, then this is a
+// no-op. Returns true if there was an update detected. Thread safe.
+func (g *Gate[T]) Set(condition T, value bool) bool {
 	g.mux.Lock()
 
 	if g.conditions == nil {
@@ -68,9 +62,9 @@ func (g *Gate[T]) markCondition(condition T, ready bool) bool {
 	old, found := g.conditions[condition]
 
 	updated := false
-	if !found || old != ready {
+	if !found || old != value {
 		updated = true
-		g.conditions[condition] = ready
+		g.conditions[condition] = value
 	}
 	// process() would also like to lock the mux, so we must unlock here directly and not use defer.
 	g.mux.Unlock()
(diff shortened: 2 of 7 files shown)


negz:
xref https://github.com/crossplane/crossplane/issues/6611

pkg/controller/gate.go
```diff
@@ -0,0 +1,16 @@
...
+
+// A Gate is an interface to allow reconcilers to delay a callback until a set of GVKs are set to true inside the gate.
+type Gate interface {
+	// Register to call a callback function when all given GVKs are marked true. If the callback is unblocked, the
+	// registration is removed.
+	Register(callback func(), gvks ...schema.GroupVersionKind)
+	// True marks a given gvk true, and then looks for unlocked callbacks. Returns if there was an update or not.
+	True(gvk schema.GroupVersionKind) bool
+	// False marks a given gvk false. Returns if there was an update or not.
+	False(gkv schema.GroupVersionKind) bool
```
negz: Would true/false ever represent something other than existence in the case of a GVK?

"The GVK is true" or "the GVK is false" lacks meaning to me. I'd find methods that made it clearer we were marking the GVK as existing or not would be more intuitive.

pkg/gate/gate.go
```diff
@@ -0,0 +1,109 @@
...
+
+func (g *Gate[T]) process() {
+	g.mux.Lock()
+	defer g.mux.Unlock()
+
+	for i := range g.fns {
+		release := true
+
+		for _, dep := range g.fns[i].depends {
+			if ready := g.conditions[dep]; !ready {
```
negz:
```suggestion
			if g.ready[dep] {
```

WDYT about this? Or `g.satisfied[dep]`?

I like naming maps of bools like way so assigns/checks are human readable. e.g. `satisfied[dep] = true`, `if satisfied[dep]`.

pkg/gate/gate.go
```diff
@@ -0,0 +1,109 @@
...
+
+	return updated
+}
+
+func (g *Gate[T]) process() {
+	g.mux.Lock()
+	defer g.mux.Unlock()
+
+	for i := range g.fns {
+		release := true
```
negz, line 85: FWIW I had to think a lot to wrap my head around what "release" meant. In retrospect makes sense -  "released from the gate" but I'd suggest a clarifying comment or a different name to make it quicker to parse when reading.

pkg/gate/gate_test.go
```diff
@@ -0,0 +1,299 @@
...
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+
+	"github.com/crossplane/crossplane-runtime/pkg/gate"
+)
+
+func TestGate_Register(t *testing.T) {
```
negz:
```suggestion
func TestGateRegister(t *testing.T) {
```

Nit: I really can't stand `_` in test names. I've always found it weird that underscores are verboten in Go function names, except tests. We've avoided them thus far, I'd prefer to keep doing so.

No real reason except that mixing camel case and snake case in the same function name looks gross. 🤷

pkg/reconciler/customresourcesgate/reconciler.go
```diff
@@ -0,0 +1,95 @@
...
+
+// +kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions,verbs=get;list;watch
+
+// Reconcile reconciles CustomResourceDefinitions and reports ready and unready GVKs to the gate.
+func (r *Reconciler) Reconcile(_ context.Context, crd *apiextensionsv1.CustomResourceDefinition) (ctrl.Result, error) {
+	// If there is no gate, then we don't need to do work.
+	if r.gate == nil || reflect.ValueOf(r.gate).IsNil() {
+		r.log.Debug("Gate is not set, skipping reconciliation")
+		return ctrl.Result{}, nil
+	}
```
negz: When would we use this reconciler without a gate?

pkg/reconciler/customresourcesgate/reconciler.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+// Reconciler reconciles a CustomResourceDefinitions in order to gate and wait
+// on CRD readiness to start downstream controllers.
+type Reconciler struct {
+	log  logging.Logger
+	gate controller.Gate
+}
+
+// +kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions,verbs=get;list;watch
```
negz: I don't think we run whatever code generation uses this marker - I'd prefer to drop it.

negz (commented):
Overall looks good, few nits.

pkg/gate/gate.go
n3wscott: I could go in with `g.conditions[dep]`, but I want to keep it generic.

pkg/gate/gate_test.go
n3wscott: I think you are real wrong on this one. it makes testing way more clear and you are fighting upstream on this issue. This is how the generators do it.

pkg/reconciler/customresourcesgate/reconciler.go
n3wscott: Never, but the gate can be nil without us wanting it to be if the gate is never setup.

pkg/reconciler/customresourcesgate/reconciler.go
n3wscott: 😢  it is the standard kubebuidler marker to make rbac

pkg/controller/gate.go
n3wscott: I am bending because of the generic gate impl. I could combine the functions to one thing... let me think.

pkg/gate/gate.go
n3wscott: added comments

pkg/controller/gate.go
n3wscott: mooshed into `Set(condition, value)` and kept it generic. But you can do cute things in the interface that is for the consumer: `Set(gvk schema.GroupVersionKind, ready bool) bool`, which is how that consumer wants to think about what it is setting.

pkg/reconciler/customresourcesgate/reconciler.go
n3wscott: Yeah good call, will throw an error in setup, not here.

negz (approved):

pkg/gate/gate.go
```diff
@@ -0,0 +1,106 @@
...
+	g.mux.Lock()
+	defer g.mux.Unlock()
+
+	for i := range g.fns {
+		// release controls if we should release the function.
+		release := true
+
+		for _, dep := range g.fns[i].depends {
+			if !g.satisfied[dep] {
+				release = false
```
verthais, line 89: can we save some cycles if we just `break` here, since we know that "all deps need to be satisfied". So as soon as we find first not satisfied dep we do not need to check further, the `release` will be `false` so we will skip if check, and jump str8 into next `fn` ?

PR #864, opened 2025-08-06
Add support for Orphan management policy

bobh66:

Added an Orphan management policy value which maps to ['Observe', 'Create', 'Update', 'LateInitialize'] in order to simplify the migration from `deletionPolicy` to `managementPolicies` and facilitate controlled orphaning of resources.

Fixes #6694 (Crossplane) 

I have: 

### DIFF
--- a/pkg/reconciler/managed/reconciler_legacy_test.go
+++ b/pkg/reconciler/managed/reconciler_legacy_test.go
@@ -86,7 +86,7 @@ func TestReconciler(t *testing.T) {
 			},
 			want: want{result: reconcile.Result{}},
 		},
-		"UnpublishConnectionDetailsDeletionPolicyDeleteOrpahn": {
+		"UnpublishConnectionDetailsDeletionPolicyDeleteOrphan": {
 			reason: "Errors unpublishing connection details should trigger a requeue after a short wait.",
 			args: args{
 				m: &fake.Manager{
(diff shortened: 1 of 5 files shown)


.github/renovate.json5
```diff
@@ -14,7 +14,6 @@
   // PLEASE UPDATE THIS WHEN RELEASING.
   "baseBranches": [
     'main',
-    'release-1.18',
```
jbw976: this was already done in https://github.com/crossplane/crossplane-runtime/pull/865, so it's odd to see here again 🤔

pkg/reconciler/managed/reconciler_legacy_test.go
```diff
@@ -2459,6 +2569,33 @@ func TestLegacyShouldDelete(t *testing.T) {
 			},
 			want: want{delete: false},
 		},
+		"DeletionDeleteManagementActionOrphan": {
```
jbw976, line 2659: what do you think about another test case for [`TestShouldDelete`](https://github.com/crossplane/crossplane-runtime/blob/main/pkg/reconciler/managed/reconciler_modern_test.go#L2354) too? e.g. make sure deletion is not allowed when this new orphan policy is set?

jbw976 (approved):
This looks like a nice convenience policy that streamlines the experience for enabling the orphaning of resources. Thanks @bobh66! just a couple questions

lsviben:
IMO the idea of management policies is that we define primitives like Observe, Create, Update, Delete that clearly state what will be done with the resources. I intentionally didnt put LateInitialize here as its a bit of an outlier and in hindsight should have been a separate feature outside of management policies.

Adding more policies when we can do the same with the existing ones seems the move in the wrong direction. IMO it would just complicate things more. As its not a primitive, but a policy like `*` which contains more policies it would be another special case that cannot be combined with other policies.

If we really need a simplifier for management policies, maybe we should introduce a separate field:
`managementPolicyAlias` which would be translated into `managementPolicies`.

So you could just set `managementPolicyAlias: Orphan` which would fill in `managementPolicies: ['Observe', 'Create', 'Update', 'LateInitialize']`.

pkg/reconciler/managed/reconciler_legacy_test.go
bobh66: Done

PR #928, opened 2026-02-15
fix ssa patch by applying all managed fields at once

### LINKED ISSUE #694: Multiple Resolver resolution results in infinite patching
### What happened?
We found a reproducable inconsistancy when persisting resolved references. Basically whenever you have more then one reference and one of them is resolved in a later reconcilation loop then the first one patching this newly resolved value will purge previously resolved values at the same time, which will result in an endless loop of patching different values back and forth. 

Debugging this issue showed that the issue must lie in the way that the change is send as a server-sided apply in this line: 
https://github.com/crossplane/crossplane-runtime/blob/master/pkg/reconciler/managed/api.go#L193

Just FYI, doing the same patch from kubectl using a merge patch instead worked as expected. 

### How can we reproduce it?
A reproducable case can be easily reproduce in this example provider: https://github.com/mirzakopic/provider-xp-bug
You will find a `mytype` Kind having two references towards other kinds called `sometype` and `othertype`. 
The issue appears when you first apply the mytype with only one reference and reapply another version with both references being set. 

You can reproduce that with the provider using this flow:
1. (Setup the provider in a local environment)
2. `kubectl apply -f examples/sample/mytype.yaml`
3. `kubectl apply -f examples/sample/mytype_added_ref.yaml`

Screenflow of endless resolving:
refresolverpatchbugxp (1)

Crossplane version: 
1.16 and 1.15 (but should apply to older versions as well(

TehilaTheStudent:

When [APISimpleReferenceResolver](cci:2://file:///home/user/work/git-repo/crossplane-repos/crossplane-debug/crossplane-oss/fix/crossplane-runtime/pkg/reconciler/managed/api.go:204:0-206:1) resolves references and applies an SSA patch, it only included newly resolved fields. The API server interpreted missing previously-managed fields as intentional deletions, causing an infinite reconciliation loop.
 
This adds [mergeWithManagedFields](cci:1://file:///home/user/work/git-repo/crossplane-repos/crossplane-debug/crossplane-oss/fix/crossplane-runtime/pkg/reconciler/managed/api.go:244:0-275:1) which extracts fields previously managed by `managed.crossplane.io/api-simple-reference-resolver` and merges them into the patch before applying.
 

Fixes #694

I have: 

- [ X] Read and followed Crossplane's [contribution process].
- [ X] Run `./nix.sh flake check` to ensure this PR is ready for review.
- [ X] Added or updated unit tests.

### DIFF
--- a/pkg/reconciler/managed/api.go
+++ b/pkg/reconciler/managed/api.go
@@ -25,11 +25,14 @@ import (
 	"github.com/google/go-cmp/cmp/cmpopts"
 	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/managedfields"
 	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/structured-merge-diff/v6/typed"
 
 	"github.com/crossplane/crossplane-runtime/v2/pkg/errors"
 	"github.com/crossplane/crossplane-runtime/v2/pkg/meta"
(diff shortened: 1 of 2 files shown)


loafoe:
While creating a native (non upjet) provider and got bitten hard by this one, hope this patch makes it in a release soon.

PR #942, opened 2026-03-16
Draft: Add UpToDate condition for managed resources

### LINKED ISSUE #939: Expose the ResourceUpToDate status from Observe in the MR status/conditions
When resources are in Observe-only mode there is no way to know if the resource matches the remote resource, and if adding the Update policy would trigger any updates in the remote.

The `Observe` method returns a `ResourceUpToDate` flag that would determine whether the `Update` method gets called if it was allowed, so it would be very useful to expose the value of this flag on the Managed Resource status, possibly as a Condition.

bobh66:

Adds a new UpToDate condition (name is subject to better suggestions) that exposes the `ResourceUpToDate` status from the `Observe()` call. This allows users to know whether the ManagedResource desired state matches the external resource, and whether an update has been requested, failed, or restricted by management policies.

There are five basic (non-error condition) scenarios:

- Reconciliation is paused so the condition status is Unknown
- The resource is up to date with the remote resource so the condition is True
- The resource has differences from the remote resource and an Update() call has failed, the condition is False
- The resource has differences from the remote resource and an Update() call cannot be made due to restrictions from `managementPolicies`, the condition is False
- The resource has differences from the remote resource and Update() returned success - the condition is False because we can't know if the update is actually asynchronous and may yet fail, but if it succeeds the next reconciliation will show the changes from the update and the condition should become True.

The three False condition scenarios will include the contents of the `observation.Diff` if it is available.

This is important when importing existing resources to be able to be sure that Crossplane will not make any enexpected changes to the external resource.

Fixes #939 

I have: 

- [X] Run `./nix.sh flake check` to ensure this PR is ready for review.
~- [ ] Linked a PR or a [docs tracking issue] to [document this change].~

### DIFF
--- a/apis/common/condition.go
+++ b/apis/common/condition.go
@@ -45,6 +45,10 @@ const (
 	// as a system condition. See the tracking issue for more details
 	// https://github.com/crossplane/crossplane/issues/5643.
 	TypeHealthy ConditionType = "Healthy"
+
+	// TypeUpToDate resources are believed to accurately represent the remote resource state.
+	// The condition is controlled by the ResourceUpToDate return value from the Observe method.
+	TypeUpToDate ConditionType = "UpToDate"
 )
 
 // A ConditionReason represents the reason a resource is in a condition.
@@ -63,6 +67,14 @@ const (
 	ReasonReconcileSuccess ConditionReason = "ReconcileSuccess"
 	ReasonReconcileError   ConditionReason = "ReconcileError"
 	ReasonReconcilePaused  ConditionReason = "ReconcilePaused"
+)
+
+// Reasons a resource is or is not up to date.
+const (
+	ReasonUpdateRestricted ConditionReason = "UpdateRestricted"
+	ReasonObserveMatch     ConditionReason = "ObserveMatch"
+	ReasonUpdateFailed     ConditionReason = "UpdateFailed"
+	ReasonUpdateRequested  ConditionReason = "UpdateRequested"
 )
 
 // See https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties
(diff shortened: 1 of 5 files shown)


apis/common/condition.go
```diff
@@ -96,8 +108,7 @@ type Condition struct { //nolint:recvcheck // False positive - only has non-poin
 }
 
 // Equal returns true if the condition is identical to the supplied condition,
-// ignoring the LastTransitionTime.  If one or both conditions have not
-// provided the ObservedGeneration it is not considered in the comparison.
+// ignoring the LastTransitionTime.
```
jcogilvie: why the change?

pkg/reconciler/managed/reconciler.go
```diff
@@ -1107,11 +1107,9 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (resu
 			record.Event(managed, event.Warning(reasonCannotDisconnect, err))
 		}
 
-		if external != nil {
```
jcogilvie, line 1110: why the change?  is this safe?

pkg/reconciler/managed/reconciler_legacy_test.go
```diff
@@ -351,7 +351,7 @@ func TestReconciler(t *testing.T) {
 						MockGet: legacyManagedMockGetFn(nil, 42),
 						MockStatusUpdate: test.MockSubResourceUpdateFn(func(_ context.Context, obj client.Object, _ ...client.SubResourceUpdateOption) error {
 							want := newLegacyManaged(42)
-							want.SetConditions(xpv1.ReconcileSuccess().WithObservedGeneration(42))
+							want.SetConditions(xpv1.ReconcileSuccess().WithObservedGeneration(42), xpv1.ObserveMatch().WithObservedGeneration(42))
```
jcogilvie, line 354: It doesn't look like we ever assert on the diff message in any of these tests -- could/should we?

apis/common/condition.go
bobh66: Oops - left over from previous changes.  Must not have cleaned up properly.  I'll remove.

pkg/reconciler/managed/reconciler.go
bobh66: Ditto

pkg/reconciler/managed/reconciler_legacy_test.go
bobh66: Probably should for the three cases where it can happen.  I'll have to look into how we can inject specific diffs, since it isn't finding any now.

bobh66:
Superseded by #945

PR #945, opened 2026-04-10
Implement UpToDate condition in managed reconciler

### LINKED ISSUE #939: Expose the ResourceUpToDate status from Observe in the MR status/conditions

bobh66:

Adds a new UpToDate condition (name is subject to better suggestions) that exposes the ResourceUpToDate status from the Observe() call. This allows users to know whether the ManagedResource desired state matches the external resource, and whether an update has been requested, failed, or restricted by management policies.

There are five basic (non-error condition) scenarios:

    Reconciliation is paused so the condition status is Unknown
    The resource is up to date with the remote resource so the condition is True
    The resource has differences from the remote resource and an Update() call has failed, the condition is False
    The resource has differences from the remote resource and an Update() call cannot be made due to restrictions from managementPolicies, the condition is False
    The resource has differences from the remote resource and Update() returned success - the condition is False because we can't know if the update is actually asynchronous and may yet fail, but if it succeeds the next reconciliation will show the changes from the update and the condition should become True.

The three False condition scenarios will include the contents of the observation.Diff if it is available.

This is important when importing existing resources to be able to be sure that Crossplane will not make any enexpected changes to the external resource.

I'm open to renaming the condition if there is a better option.

This build will fail until crossplane/crossplane#7294 is merged, at which point go.mod needs to be updated to the new version.

Fixes #939 

I have: 

- [X] Run `./nix.sh flake check` to ensure this PR is ready for review.
- [X] Linked a PR or a [docs tracking issue] to [document this change].


pkg/reconciler/managed/reconciler.go
```diff
@@ -1456,7 +1456,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (resu
 	if !policy.ShouldUpdate() {
 		reconcileAfter := r.pollIntervalHook(managed, r.pollInterval)
 		log.Debug("Skipping update due to managementPolicies. Reconciliation succeeded", "requeue-after", time.Now().Add(reconcileAfter))
-		status.MarkConditions(xpv2.ReconcileSuccess())
+		status.MarkConditions(xpv2.ReconcileSuccess(), xpv2.UpdateRestricted().WithMessage(observation.Diff))
```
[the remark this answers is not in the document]
jcogilvie: I don't think so.  The point of this condition is to surface to the end user what elements of the resource are out of sync so they can operate on them.  End users don't necessarily have access to the debug logs they would otherwise need to figure out what fields are out of sync.

pkg/reconciler/managed/reconciler.go
jbw976: this path will not get the status condition `UpdateRequested()`, even though the update was successfully requested. 

what do you think about always calling `status.MarkConditions(xpv2.UpdateRequested().WithMessage(observation.Diff))` immediately after the update request is successful, e.g. around line 1482-1488? that way, we have the condition set no matter what path gets taken later on.

that would also mean to remove the setting of it on line 1509, because we've already done it.

it's my understanding that calling status.MarkCondtions() multiple times is safe as it only replaces matching condition types, it won't remove any conditions: https://github.com/crossplane/crossplane/blob/main/apis/core/v2/condition.go#L190-L193 

so this:
```
status.MarkConditions(xpv2.ReconcileSuccess(), xpv2.UpdateRequested().WithMessage(observation.Diff))
```

should be the same as:
```
status.MarkConditions(xpv2.UpdateRequested().WithMessage(observation.Diff))

# then later on...
status.MarkConditions(xpv2.ReconcileSuccess())
```

pkg/reconciler/managed/reconciler.go
jbw976, line 1519: what are we thinking for the size of these diffs and limits on k8s conditions fields? i think `metav1.Condition.Message` has a limit of 32KB: https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L1664

should we consider truncating this diff to be safe?

jbw976 (commented):
Thanks for driving this @bobh66. A couple meta questions in addition to the review comments:

* what does this look like in practice? e.g. do you have an example of a `kubectl describe` with these conditions added, especially with a diff?
* have we verified that providers are setting the diff field in practice and we'll typically have good data to show? e.g. a couple code paths in upjet that may be relevant that are not setting the ExternalObservation.Diff field:
  * https://github.com/crossplane/upjet/blob/main/pkg/controller/external_tfpluginfw.go#L706-L711
  * https://github.com/crossplane/upjet/blob/main/pkg/controller/external.go#L358-L362

bobh66:
> Thanks for driving this @bobh66. A couple meta questions in addition to the review comments:
> 
>     * what does this look like in practice? e.g. do you have an example of a `kubectl describe` with these conditions added, especially with a diff?

I have not been able to test this with a real provider yet.  The API refactoring that moved the common APIs to c/c caused a lot of downstream fallout that makes pulling these changes into a provider for testing a lot more work.  I'm hoping to get it working in provider-kubernetes but it's a lot more effort than I had expected.
 
>     * have we verified that providers are setting the diff field in practice and we'll typically have good data to show? e.g. a couple code paths in upjet that may be relevant that are not setting the ExternalObservation.Diff field:
>       
>       * https://github.com/crossplane/upjet/blob/main/pkg/controller/external_tfpluginfw.go#L706-L711
>       * https://github.com/crossplane/upjet/blob/main/pkg/controller/external.go#L358-L362

I have not validated this, and indeed even provider-kubernetes does not supply the diff info.  We could leave it out but I think it's the direction we want/need to go, so I'd prefer to support it and let the providers catch up.

pkg/reconciler/managed/reconciler.go
bobh66: Probably a good idea.

pkg/reconciler/managed/reconciler.go
bobh66: That works - I'll make the change.

PR #982, opened 2026-04-20
Add per-resource poll interval and reconcile-request annotations

yordis:
- Companion to crossplane/crossplane#7239 which implemented the XR side
- Operators managing stable, rarely-changing resources (databases, DNS zones) pay unnecessary API costs with the global poll interval, while dynamic resources need faster drift detection — there is no middle ground today
- Providers inherit both features upon upgrading their crossplane-runtime dependency with no code changes, because managed resources that embed `ObservedStatus` automatically satisfy the reconcile-request tracker interface


adamwg (approved):
lgtm, thanks for seeing this through!

PR #1031, opened 2026-06-13
fix(tests): update error messages for semver v3.5.0 constraint quotes

boxcee-interview:
semver v3.5.0 wraps invalid constraint values in double quotes in error messages.

Updated test expectations in:
- pkg/version/version_test.go: constraint strings now include quotes
- pkg/xpkg/lint_test.go: constraint error string format updated

Tests verified locally with go test ./...


bobh66 (approved):
LGTM - Thanks!

Issue #1010, opened 2026-05-29
[Bug]: Namespace resources block namespace deletion


sjiekak:
## What happened?                                                                                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                                                  
  When a namespace containing Crossplane managed resources (e.g. `roles.iam.aws.m.upbound.io`, `lbs.elbv2.aws.m.upbound.io`, `securitygroups.ec2.aws.m.upbound.io`, `vpcendpointservices.ec2.aws.m.upbound.io`) is deleted, the namespace gets stuck in `Terminating` indefinitely.

I initially taught it is an issue with upbound aws and filed https://github.com/crossplane-contrib/provider-upjet-aws/issues/2091 but I see many projects are using `Track`, so it might be related to the runtime behavior                                                                                              
                                                                                                                                                                                                                                                                                                                                                                                    
  `kubectl describe` on the namespace shows that the managed resources still carry `finalizer.managedresource.crossplane.io`, and content cannot be removed:                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                                                                                    
  ```                                                                                                                                                                                                                                                                                                                                                                               
  Status:       Terminating
  Conditions:
    NamespaceDeletionDiscoveryFailure            False   ResourcesDiscovered   All resources successfully discovered
    NamespaceDeletionGroupVersionParsingFailure  False   ParsedGroupVersions   All legacy kube types successfully parsed                                                                                                                                                                                                                                                            
    NamespaceDeletionContentFailure              False   ContentDeleted        All content successfully deleted, may be waiting on finalization                                                                                                                                                                                                                                     
    NamespaceContentRemaining                    True    SomeResourcesRemain   Some resources are remaining: securitygroups.ec2.aws.m.upbound.io has 1 resource instances, vpcendpointservices.ec2.aws.m.upbound.io has 1 resource instances                                                                                                                                        
    NamespaceFinalizersRemaining                 True    SomeFinalizersRemain  Some content in the namespace has finalizers remaining: finalizer.managedresource.crossplane.io in 2 resource instances                                                                                                                                                                              
  ```                                                                                                                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                                                                                    
  Looking at the resource conditions, the controller is trying to **create** a `providerconfigusages.aws.m.upbound.io` in the namespace that is being terminated, which the API server (correctly) rejects:                                                                                                                                                                               
[...]

PR #1048, opened 2026-06-28
fix(meta): RemoveFinalizer handles duplicate finalizers without panicking

arpitjain099:

`RemoveFinalizer` in pkg/meta deletes from the finalizer slice in place while ranging over it by index, which is the classic delete-while-ranging bug:

```go
for i, e := range f {
    if e == finalizer {
        f = append(f[:i], f[i+1:]...)
    }
}
```

`range` fixes the length at loop start but `append(f[:i], f[i+1:]...)` shrinks `f` underneath it, so after one removal the loop keeps indexing into a now-shorter slice. When the finalizer appears more than once this either panics (`["fin","fin"]`) or silently leaves a copy behind (`["fin","fin","fun"]` returns `["fin","fun"]`). It backs `APIFinalizer.RemoveFinalizer` on the managed-resource deletion path, so a duplicate finalizer (concurrent reconciles, patch races, version skew) can crash a controller during deletion or leave it stuck Terminating, which lines up with #1010. `AddFinalizer` already dedupes with `slices.Contains`; this side didn't.

Swapped the loop for `slices.DeleteFunc`, which removes all matches in one pass. `slices` is already imported here, so no new deps. Added two duplicate cases to `TestRemoveFinalizer`; they panic before the change and pass after.

`go test -race ./pkg/meta/...`, `go vet ./pkg/meta/...`, and `go build ./...` are all green.

I have:

- [ ] ~Run `./nix.sh flake check` to ensure this PR is ready for review.~ (ran `go test -race`, `go vet`, and `go build ./...` directly; happy to run flake check if needed)
- [ ] ~Linked a PR or a [docs tracking issue] to [document this change].~ (internal helper, no user-facing docs)
- [ ] ~Added `backport release-x.y` labels to auto-backport this PR.~ (deferring to maintainers on backport scope)

### DIFF
--- a/pkg/meta/meta.go
+++ b/pkg/meta/meta.go
@@ -190,11 +190,9 @@ func AddFinalizer(o metav1.Object, finalizer string) {
 // RemoveFinalizer from the supplied Kubernetes object's metadata.
 func RemoveFinalizer(o metav1.Object, finalizer string) {
 	f := o.GetFinalizers()
-	for i, e := range f {
-		if e == finalizer {
-			f = append(f[:i], f[i+1:]...)
-		}
-	}
+	f = slices.DeleteFunc(f, func(e string) bool {
+		return e == finalizer
+	})
 
 	o.SetFinalizers(f)
 }
--- a/pkg/meta/meta_test.go
+++ b/pkg/meta/meta_test.go
@@ -538,6 +538,28 @@ func TestRemoveFinalizer(t *testing.T) {
 			},
 			want: []string{funalizer},
 		},
+		"DuplicateFinalizersExist": {
+			args: args{
+				o: &corev1.Pod{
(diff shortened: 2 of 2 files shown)


adamwg (approved):
lgtm, thanks for the fix!

arpitjain099:
Following up on this one. @adamwg approved it on 4 August and it has been quiet since. If you would like anything adjusted before it lands I am glad to do it, otherwise I think it is ready.

adamwg:
Thanks for the poke, @arpitjain099. CI didn't run automatically on this PR because of our settings, but I also don't see the button that usually lets maintainers trigger it. I'm going to try closing and re-opening the PR to see if it shows up. Assuming CI is green I'm happy to merge.

PR #1058, opened 2026-07-20
feat: add TransformStripCRDSchema for reduced cache memory usage

### LINKED ISSUE #1056: Reduce Provider memory footprint by optimizing CRD and Secret caching
### **Background & Problem Statement**
When deploying a Crossplane provider (built from `provider-template` or official Upjet providers like `provider-upjet-aws` or `provider-vault`) on a large-scale Kubernetes cluster, the provider's pod memory usage can balloon rapidly, leading to `OOMKilled` crashes even under **near-zero** reconciler load.

On our largest production cluster (~900 CRDs, ~4,000 Secrets), a newly deployed provider managing a single resource type sat at **~1.5 GB of memory** shortly after creating the first managed resource. 

We traced this memory bloat back to the default `controller-runtime` manager cache configuration that the template and official providers inherit. Specifically, two types of resources are being cached in their entirety:
1. **Every CRD in the cluster:** The SafeStart gate implementation watches/reads CRDs. Because the cache has no filters or transforms, it caches the **full OpenAPI v3 schema** and `managedFields` of every single CRD on the cluster.
https://github.com/crossplane/crossplane-runtime/blob/2cf2ed593a36789b22f4e3aecd8611c1b982346c/pkg/reconciler/customresourcesgate/setup.go#L42-L45
2. **Every Secret in the cluster:** Because providers perform point-reads on Secrets (for credential resolution and connection-secret publishing), `controller-runtime`'s default client utilizes a GVK-level Informer. Upon the **very first point-read of any single Secret**, the client initializes an Informer for the Secret type, which immediately issues a cluster-wide `List` call and caches **every Secret in the cluster** in memory. ([`provider-vault` cache definition](https://github.com/upbound/provider-vault/blob/bf6d1e756742933272d5d003979183309cf90ef5/cmd/provider/main.go#L155-L157) & [example `provider-upjet-aws` cache definition](https://github.com/crossplane-contrib/provider-upjet-aws/blob/c52635ac7753ec54ca9ad187125ac0f57f9f1f79/cmd/provider/ec2/zz_main.go#L157-L159))
---

### **Proposed Optimizations**

We tested two highly effective optimizations in production that reduced our provider's memory footprint from **~1.5 GB to ~80 MB**, with absolutely no loss of functionality:

#### **1. Strip OpenAPI Schemas and Metadata from Cached CRDs**
The SafeStart gate only needs to check basic metadata (group, name, version, and `Established` status) to determine if a controller is safe to start. It never reads the OpenAPI schema. 

We can apply a cache `Transform` to drop these massive fields before they ever hit the memory cache:

```go
// 1) Clear schema and bloated metadata off CRDs entering the cache
cache.Options{
    ByObject: map[client.Object]cache.ByObject{
        &apiextensionsv1.CustomResourceDefinition{}: {Transform: stripCachedCRD},
    },
}

func stripCachedCRD(obj any) (any, error) {
    crd, ok := obj.(*apiextensionsv1.CustomResourceDefinition)
    if !ok {
        return obj, nil
    }
    for i := range crd.Spec.Versions {
        crd.Spec.Versions[i].Schema = nil
    }
    crd.ManagedFields = nil
    delete(crd.Annotations, "kubectl.kubernetes.io/last-applied-configuration")
    return crd, nil
}
```

#### **2. Disable Caching for Secrets**
Because providers point-read Secrets and do not actually watch them, we can configure the client to bypass the cache for Secrets entirely, reading them directly from the API server:

```go
// 2) Keep Secrets off the cache
client.Options{
    Cache: &client.CacheOptions{
        DisableFor: []client.Object{&corev1.Secret{}},
    },
}
```

---

### **Production Impact & Verification**
We have been running these changes in production with the following results:
* **Memory footprint:** Dropped from **~1.5 GB to ~80 MB** on our largest cluster.
* **Functionality:** 100% stable. The SafeStart gate still correctly detects CRDs and starts controllers, credentials still resolve, connection-secrets publish successfully, and reconciliation is completely unaffected.

---

### **Questions & Architectural Alignment**

To upstream this optimization safely, we would love the maintainers' feedback on the following:

1. **CRD Schema Dependencies:** Are there any other consumers or edge cases in `crossplane-runtime` (or standard provider patterns) that rely on reading `CRD.Spec.Versions[].Schema` from the cache? 
2. **Secret Watches:** Is there any current or planned logic in `crossplane` or `upjet` that watches Secrets and would therefore be broken by `DisableFor: []client.Object{&corev1.Secret{}}`?
3. **Rollout Strategy:** Since `crossplane-runtime` is the target repository, how should this behavior be exposed?
   * Should this be strictly **opt-in** (e.g., exposed via new options/flags that providers must explicitly enable)?
   * Or can we make this the **default** behavior in the runtime, given that standard providers don't seem to rely on these cached fields, perhaps offering an opt-out for custom implementations?

rafal-jan:

Fixes #1056

I have: 

- [x] Run `./nix.sh flake check` to ensure this PR is ready for review.

This PR introduces a new cache optimization function, `TransformStripCRDSchema`, to reduce the memory footprint of Crossplane providers that watch CustomResourceDefinition (CRD) objects.

By default, the controller-runtime cache stores the entire schema for every watched object. For clusters with a massive number of CRDs storing the OpenAPI v3 validation schemas in the informer cache leads to significant memory bloat (hundreds of megabytes per provider pod).

`TransformStripCRDSchema` is a `cache.TransformFunc` that safely strips the following fields from CRD objects before they enter the informer cache:

- `Spec.Versions[].Schema` (OpenAPI v3 validation schemas)
- `ObjectMeta.ManagedFields`
- The `kubectl.kubernetes.io/last-applied-configuration` annotation

Providers can leverage this by wiring it into their manager's `cache.Options.ByObject` configuration. Since most controllers (like the CRD gate) only require basic metadata, group/version/kind names, and status conditions to function correctly, stripping these heavy fields is completely safe.

## How has this been tested?

- Added comprehensive unit tests in `pkg/reconciler/customresourcesgate/cache_test.go` verifying that schemas, specific annotations, and managed fields are correctly stripped while leaving all other data and non-CRD objects intact.
- Validated via a local test bed using `provider-template`. Testing a provider against a cluster with 2,045 Upjet AWS CRDs showed controller memory usage dropping from ~630 MiB down to ~83 MiB after applying this cache transform.

### DIFF
--- a/pkg/reconciler/customresourcesgate/cache.go
+++ b/pkg/reconciler/customresourcesgate/cache.go
@@ -1,0 +1,58 @@
+/*
+Copyright 2025 The Crossplane Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package customresourcesgate
+
+import (
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+)
+
+// TransformStripCRDSchema is a cache.TransformFunc that removes heavy fields
+// from CustomResourceDefinition objects before they are stored in the informer
+// cache. It strips:
+//   - Spec.Versions[].Schema (OpenAPI v3 validation schemas)
+//   - ObjectMeta.ManagedFields
+//   - The "kubectl.kubernetes.io/last-applied-configuration" annotation
+//
+// This significantly reduces memory usage in clusters with many CRDs. The CRD
+// gate reconciler only needs basic metadata (group, kind, version names, served
+// status) and status conditions to function correctly.
+//
+// Usage:
+//
+//	cache.Options{
+//	    ByObject: map[client.Object]cache.ByObject{
+//	        &apiextensionsv1.CustomResourceDefinition{}: {
+//	            Transform: customresourcesgate.TransformStripCRDSchema,
+//	        },
+//	    },
+//	}
+func TransformStripCRDSchema(obj any) (any, error) {
+	crd, ok := obj.(*apiextensionsv1.CustomResourceDefinition)
+	if !ok {
+		return obj, nil
+	}
+
+	for i := range crd.Spec.Versions {
+		crd.Spec.Versions[i].Schema = nil
+	}
+
+	crd.ManagedFields = nil
+
+	delete(crd.Annotations, "kubectl.kubernetes.io/last-applied-configuration")
+
+	return crd, nil
+}
--- a/pkg/reconciler/customresourcesgate/cache_test.go
+++ b/pkg/reconciler/customresourcesgate/cache_test.go
@@ -20,6 +20,7 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -238,7 +239,7 @@ func TestTransformStripCRDSchema(t *testing.T) {
 		t.Run(name, func(t *testing.T) {
 			got, err := TransformStripCRDSchema(tc.args.obj)
 
-			if diff := cmp.Diff(tc.want.err, err); diff != "" {
+			if diff := cmp.Diff(tc.want.err, err, cmpopts.EquateErrors()); diff != "" {
 				t.Errorf("%s\nTransformStripCRDSchema(...): -want error, +got error:\n%s", tc.reason, diff)
 			}
 
--- a/pkg/reconciler/customresourcesgate/cache_test.go
+++ b/pkg/reconciler/customresourcesgate/cache_test.go
@@ -1,0 +1,250 @@
+/*
+Copyright 2025 The Crossplane Authors.
+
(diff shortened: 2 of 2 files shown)


bobh66 (approved):
This LGTM but I'm far from an expert in this area, so might want someone else to take a look.

jonasz-lasut:
Hi @rafal-jan thank you for contributing the PR!

I've done a short safety verification against consumers of this code - `crossplane-runtime` and `upjet` like we've discussed on slack.

- **Confirmed (code):** the gate reconciler
  (`crossplane-runtime/pkg/reconciler/customresourcesgate/reconciler.go:40-86`) reads
  only `Spec.Group`, `Spec.Versions[].Name`/`.Served`, `Spec.Names.Kind`,
  `Status.Conditions` (`Established`), and the deletion timestamp — all preserved by
  the transform.
- **Confirmed (code):** the gate is the *only* consumer of cached CRDs in
  crossplane-runtime. `pkg/xcrd` and `pkg/xpkg` generate/lint CRDs from package
  contents; neither reads the informer cache.
- **Confirmed (code):** upjet's CRD storage-version migrator is unaffected twice over:
  `cmd/crdmigrator/main.go:102` builds a **direct** client (`client.New`, no cache),
  and `UpdateCRDStorageVersion` (`upjet/pkg/config/crd_migrator.go:221-238`) re-`Get`s
  the CRD and patches status with `client.MergeFrom`, so only the `storedVersions`
  diff goes over the wire — it cannot write a stripped schema back even if handed a
  cached client.
- **Falsified:** the concern that something watches Secrets. No
  Secret watch, `Owns`, or informer registration exists in upjet or
  crossplane-runtime. All consumers are point reads/writes: sensitive-parameter
  resolution (`upjet/pkg/controller/api.go:57-63`), credential extraction
  (`crossplane-runtime/pkg/resource/providerconfig.go`), and connection-secret
  publishing (`crossplane-runtime/pkg/reconciler/managed/api.go`, Get + Apply). The
  generated upjet controller watches only the managed-resource kind
  (`pkg/pipeline/templates/controller.go.tmpl:153`). External secret stores — the
  historical Secret-heavy subsystem — were removed in runtime v2.3.
  
From my perspective it's ok to get this merged in the current state

jonasz-lasut (approved):

Issue #1086, opened 2026-07-29
flake.lock revisions are never refreshed, so our Go toolchain is frozen at 1.25.9


jbw976:
### What happened?

Our Go toolchain seems to be frozen at `1.25.9`, with no automation to keep it updated. Go 1.25.12 is the current 1.25 patch release.

We get `go` from the [`nixpkgs-unstable` input](https://github.com/crossplane/crossplane-runtime/blob/b864aea9c0f2deccb3ea4359088fc20af1c2db7b/flake.nix#L45-L46), which is a branch ref rather than a version. Renovate's `nix` manager tracks it, but there's no newer branch name to bump to, so it never reports an update. Refreshing the pinned revision behind a branch ref takes [`lockFileMaintenance`](https://docs.renovatebot.com/configuration-options/#lockfilemaintenance), and we don't have that enabled anywhere.

@phisco added the `nixpkgs-unstable` input in #961 to give us access to latest versions of go, but it looks like nothing is updating this over time.

Looks like we're already using `lockFileMaintenance` in core crossplane: https://github.com/crossplane/crossplane/blob/33126fd9eebe3028bc2a889885f8e85a1a70939e/.github/renovate-nix.json5#L10-L15 

### How can we reproduce it?

```
nix eval --raw --impure --expr 'let f = builtins.getFlake (toString ./.); in f.inputs.nixpkgs-unstable.legacyPackages.${builtins.currentSystem}.go_1_25.version'
1.25.9
```

### Proposed fix

Perhaps we can enable `lockFileMaintenance` for the `nix` manager in [renovate-nix.json5](https://github.com/crossplane/crossplane-runtime/blob/b864aea9c0f2deccb3ea4359088fc20af1c2db7b/.github/renovate-nix.json5#L11-L19)? That's the piece that runs `nix flake update` and moves the locked revisions, which is what we actually want here.

While we're at it, we should also move to a `weekly` rather than the `monthly` schedule that rule uses today. We still need upstream to update go versions in nixpkgs before we can pick them up, but trying weekly will help us pick it up faster after that.

PR #1087, opened 2026-07-29
chore(ci): enable weekly flake.lock maintenance for nix manager

pujitha24:
Motivation:

Our Go toolchain comes from the nixpkgs-unstable flake input, which is
pinned by a branch ref rather than a version. Renovate's nix manager
tracks this input, but a branch ref never shows a newer version to bump
to, so Renovate never proposes updating flake.lock. As a result the
locked revision - and therefore our Go version - never advances; we
have been stuck on Go 1.25.9 while 1.25.12 is current.

Approach:

Enable Renovate's lockFileMaintenance for the nix manager, which runs
`nix flake update` on a schedule regardless of whether Renovate detects
a version bump. This is the mechanism that actually moves branch-ref
inputs like nixpkgs-unstable forward. Also switch the cadence from
monthly to weekly, since we still depend on upstream nixpkgs picking up
new Go releases first and want to pick up those updates faster once it
does. This mirrors the lockFileMaintenance configuration already in use
in crossplane/crossplane's .github/renovate-nix.json5.

This is a Renovate/CI automation config change only; it does not alter
any runtime code or user-facing behavior. It changes how often and
under what conditions our automated dependency-update bot opens a PR
to bump flake.lock.

Validation:

Ran this repo's own CI validation checks for Renovate presets locally:

  npx --yes json5 .github/renovate-nix.json5
  npx --yes --package renovate -- renovate-config-validator

Both passed ("Config validated successfully").

Related to #1086

Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

### DIFF
--- a/.github/renovate-base.json5
+++ b/.github/renovate-base.json5
@@ -8,8 +8,9 @@
   // We only want renovate to rebase PRs when they have conflicts, default
   // "auto" mode is not required.
   rebaseWhen: 'conflicted',
-  // The maximum number of PRs to be created in parallel
-  prConcurrentLimit: 5,
+  // The maximum number of PRs to be created in parallel - branchConcurrentLimit
+  // inherits this value as well
+  prConcurrentLimit: 10,
   // The branches renovate should target
   // PLEASE UPDATE THIS WHEN RELEASING.
   baseBranches: [
--- a/.github/renovate-nix.json5
+++ b/.github/renovate-nix.json5
@@ -7,16 +7,18 @@
     enabled: true,
   },
 
+  // Refresh flake.lock weekly by running 'nix flake update', even when no
+  // flake input has a newer version to bump to (e.g. branch refs like
+  // nixpkgs-unstable). This is what actually moves the locked revision
+  // forward and picks up new package versions, such as Go.
+  lockFileMaintenance: {
+    enabled: true,
+    extends: [
+      'schedule:weekly',
+    ],
+  },
+
   packageRules: [
-    {
-      description: 'Update flake.lock monthly',
-      matchManagers: [
-        'nix',
-      ],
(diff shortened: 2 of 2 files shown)


jbw976 (commented):
awesome @pujitha24, thanks for taking this on so quickly! i was also working on a fix last night that updates a couple other config settings, so to stay fair and keep your credit for contribution here, i'll just push my additional settings as an extra commit to this branch/PR, unless you have major objections to that 😇

pujitha24:
Thanks @jbw976! No objections at all — happy for you to push your additional config settings as an extra commit on this branch. Let me know if there's anything else you'd like adjusted here.

jbw976:
Awesome, thanks for the collaboration @pujitha24! 🙌 

I pushed https://github.com/crossplane/crossplane-runtime/pull/1087/changes/39820284962e013822695d1b2dcf68c36e5e9054, so now this PR has both of our commits, to add a couple more settings:

* set the schedule to `on monday` so our renovate workflow cron config will be compatible with renovate's odd definition of "weekly" 😂 
* allow this lock file maintenance to happen on release branches as well, so we don't miss Go version bumps there
* bump up the concurrent PR limit to get more throughput again when there's a backlog

pujitha24:
Thanks for the fixes, @jbw976! The `on monday` schedule fix, release-branch coverage, and `prConcurrentLimit` bump all make sense to me — appreciate you catching the `schedule:weekly` reachability issue, I wouldn't have caught that. I re-ran the repo's renovate config validation locally with both commits in place and it still passes cleanly. Nothing further from my side — this looks ready whenever you are.

haarchri (approved):

PR #1113, opened 2026-08-10
Protect ProviderConfigs while managed resources terminate

### LINKED ISSUE #4661: ProviderConfigUsage does not work when deletion policy is set to Foreground
### What happened?

While testing my [function tee prototype](https://github.com/crossplane/crossplane/discussions/4639#discussioncomment-7079961) with a claim using `compositeDeletePolicy: Foreground`, I noticed that `ProviderConfigUsage` does not work as expected. 

So, my function outputs a provider kubernetes `ProviderConfig` resource (with InjectedIdentity) and an `Object` wrapping a ConfigMap and using that `ProviderConfig`. When I delete my claim, I noticed that the `Object` resource is orphaned since the `ProviderConfig` was deleted before the `Object` itself. I verified that the `ProviderConfig` resource had the `in-use.crossplane.io` finalizer and the corresponding `ProviderConfigUsage` resource existed.

After some investigation and thought, I believe this is a bug with the current implementation of `ProviderConfigUsage` when it is used with deletion policy Foreground. The issue is, with this mode of deletion propagation, resources are being deleted starting from the bottom of the tree and hence `ProviderConfigUsage` being owned by the `Object` is deleted before the actual `Object` is deleted. This removes the blockage on `ProviderConfig` and it got deleted without waiting the deletion of the `Object`. 

@bobh66 I am wondering if you had encountered something like this before? 

### How can we reproduce it?

```bash
kubectl apply -f setup.yaml
kubectl apply -f claim.yaml
# wait until claim is ready
kubectl delete -f claim.yaml

```
<details>
<summary>setup.yaml</summary>

```yaml
apiVersion: pkg.crossplane.io/v1
kind: Provider
metadata:
  name: provider-nop
spec:
  package: xpkg.upbound.io/crossplane-contrib/provider-nop:v0.2.0
  ignoreCrossplaneConstraints: true
---
apiVersion: pkg.crossplane.io/v1beta1
kind: Function
metadata:
  name: function-dummy
spec:
  # NOTE(negz): This is currently manually pushed. See README.md at
  # https://github.com/crossplane-contrib/function-dummy.
  package: xpkg.upbound.io/crossplane-contrib/function-dummy:v0.2.0
---
apiVersion: pkg.crossplane.io/v1beta1
kind: Function
metadata:
  name: function-tee
spec:
  package: turkenh/function-tee:v0.1.0
---
kind: CompositeResourceDefinition
metadata:
  name: xnopresources.nop.example.org
spec:
  group: nop.example.org
  names:
    kind: XNopResource
    plural: xnopresources
  claimNames:
    kind: NopResource
    plural: nopresources
  connectionSecretKeys:
  - test
  versions:
  - name: v1alpha1
    served: true
    referenceable: true
    schema:
     openAPIV3Schema:
       type: object
       properties:
        spec:
          type: object
          properties:
            coolField:
              type: string
          required:
          - coolField
        status:
          type: object
          properties:
            coolerField:
              type: string
---
kind: Composition
metadata:
  name: xnopresources.nop.example.org
spec:
  compositeTypeRef:
    apiVersion: nop.example.org/v1alpha1
    kind: XNopResource
  mode: Pipeline
  pipeline:
  - step: be-a-dummy
    functionRef:
      name: function-dummy
    input:
      apiVersion: dummy.fn.crossplane.io/v1beta1
      kind: Response
      # This is a YAML-serialized RunFunctionResponse. function-dummy will
      # overlay the desired state on any that was passed into it.
      response:
        desired:
          composite:
            resource:
              status:
                coolerField: "I'M COOLER!"
          resources:
            nop-resource-1:
              resource:
                apiVersion: nop.crossplane.io/v1alpha1
                kind: NopResource
                spec:
                  forProvider:
                    conditionAfter:
                    - conditionType: Ready
                      conditionStatus: "False"
                      time: 0s
                    - conditionType: Ready
                      conditionStatus: "True"
                      time: 10s
              ready: READY_TRUE
        results:
         - severity: SEVERITY_NORMAL
           message: "I am doing a compose!"
  - step: tee
    functionRef:
      name: function-tee
```

</details>

<details>
<summary>claim.yaml</summary>

```
apiVersion: nop.example.org/v1alpha1
kind: NopResource
metadata:
  namespace: default
  name: apiextensions-composition-functions
spec:
  coolField: "I'm cool!"
  # This is necessary to ensure the claim's MRs are actually gone before we
  # delete the Provider - https://github.com/crossplane/crossplane/issues/4251
  compositeDeletePolicy: Foreground
```

</details>

Crossplane version: v1.14.0-rc.0.369.g293439e0

ezgidemirel:

This PR prevents a `ProviderConfig` from being deleted while a managed resource
that uses it is still terminating.

It:

- Adds `finalizer.providerconfigusage.crossplane.io` to a `ProviderConfigUsage`
  and makes the managed reconciler the owner of both ends of that finalizer:
  `Protect` adds it right after a successful `Connect`, `Untrack` removes it
  after the managed resource has removed its own finalizer, on both the Delete
  and the Orphan/no-Delete paths. `Track` never touches the finalizer and
  preserves existing finalizers when it updates the usage.
- Exposes this through a new `ProviderConfigUsageCleaner` interface
  (`Protect`/`Untrack`), implemented by both existing trackers, and turns it
  on by default. The managed reconciler derives its cleaner from the managed
  resource's scope and the single `ProviderConfigUsage` kind of that scope
  registered with the manager's scheme, which is the kind the provider's
  connector records usages with. `managed.WithProviderConfigUsageCleaner`
  overrides the default, for schemes that register no or several usage kinds
  of a scope, and passing `NewNopProviderConfigUsageCleaner()` turns
  protection off. `NewReconciler`'s signature is unchanged.
- Skips adding the usage finalizer when the usage is already being deleted,
  since the API server rejects new finalizers on a terminating object. The
  usage is recreated and protected on the next reconcile once it is gone.
- Teaches the `ProviderConfig` reconciler to release a terminating, finalized
  usage whose owner is gone, was replaced, or has itself finished teardown
  (deleting, with no finalizers other than the garbage collector's), reading the
  owner through an uncached API reader. It also strips the finalizer before
  deleting a usage that has no owner reference.
- Changes the delayed-delete gate from #855 to count only non-GC finalizers
  (`meta.NonGCFinalizers`), so `foregroundDeletion` and `orphan` no longer
  hold back external deletion. Without this the usage finalizer deadlocks
  foreground deletion.
- Emits warning events when protecting or releasing a usage fails.

## Why is this needed?

During foreground deletion, a `ProviderConfigUsage` can be garbage-collected
before its managed resource finishes deleting its external resource. The
`ProviderConfig` controller then observes no remaining usages and removes its
`in-use.crossplane.io` finalizer. The managed resource's next reconciliation
fails because its `ProviderConfig` has already disappeared, leaving the resource
stuck in deletion.

The gate change is needed independently: any controller that adds its own
finalizer to a `ProviderConfigUsage` deadlocks foreground deletion on the
current gate.

## Provider integration

None required. A provider gets ProviderConfig deletion protection by upgrading
crossplane-runtime: the reconciler finds the provider's usage kind in the
manager's scheme, so no code change or option is needed, including for upjet
family providers that register both a cluster-scoped and a namespaced usage
kind. `managed.WithProviderConfigUsageCleaner(tracker)` is available to
override the choice, where `tracker` is a `ProviderConfigUsageTracker` or
`LegacyProviderConfigUsageTracker` built with the same client and usage type
the connector uses, and `WithProviderConfigUsageCleaner(NewNopProviderConfigUsageCleaner())`
turns protection off.

## Testing

Tested with locally built `provider-kubernetes` and `provider-aws-iam` (with
`provider-aws-config`) on kind: foreground and background cascades of
cluster-scoped and namespaced managed resources with the `ProviderConfig`
deleted concurrently, composed claims with both `compositeDeletePolicy` values,
a paused managed resource, migration of usages created without the finalizer,
and namespace deletion. Both providers were also built from their upstream
trees with only the runtime replaced, to confirm the default applies without
any provider change.

Design: crossplane/crossplane#7594
Fixes: https://github.com/crossplane/crossplane/issues/4661

I have:

- [x] Run `./nix.sh flake check` to ensure this PR is ready for review.

### DIFF
--- a/pkg/meta/meta.go
+++ b/pkg/meta/meta.go
@@ -205,6 +205,23 @@ func FinalizerExists(o metav1.Object, finalizer string) bool {
 	return slices.Contains(f, finalizer)
 }
 
+// FinalizersExcludingPropagation returns the supplied object's finalizers
+// without Kubernetes deletion propagation finalizers.
+func FinalizersExcludingPropagation(o metav1.Object) []string {
+	f := o.GetFinalizers()
+	out := make([]string, 0, len(f))
+
+	for _, e := range f {
+		if e == metav1.FinalizerDeleteDependents || e == metav1.FinalizerOrphanDependents {
+			continue
+		}
+
+		out = append(out, e)
+	}
+
+	return out
+}
+
 // AddLabels to the supplied object.
 func AddLabels(o metav1.Object, labels map[string]string) {
 	l := o.GetLabels()
--- a/pkg/meta/meta.go
+++ b/pkg/meta/meta.go
@@ -205,9 +205,11 @@ func FinalizerExists(o metav1.Object, finalizer string) bool {
 	return slices.Contains(f, finalizer)
 }
 
-// FinalizersExcludingPropagation returns the supplied object's finalizers
-// without Kubernetes deletion propagation finalizers.
-func FinalizersExcludingPropagation(o metav1.Object) []string {
+// NonGCFinalizers returns the supplied object's finalizers, excluding the two
+// the Kubernetes garbage collector uses to implement deletion propagation -
+// foregroundDeletion and orphan. Neither belongs to a controller, and neither
+// makes a claim on an external system.
+func NonGCFinalizers(o metav1.Object) []string {
 	f := o.GetFinalizers()
 	out := make([]string, 0, len(f))
 
--- a/pkg/reconciler/managed/reconciler.go
+++ b/pkg/reconciler/managed/reconciler.go
@@ -19,12 +19,14 @@ package managed
 import (
 	"context"
 	"math/rand"
+	"reflect"
 	"strings"
 	"time"
 
 	xpv2 "github.com/crossplane/crossplane/apis/v2/core/v2"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -622,19 +624,68 @@ type mrManaged struct {
 	resource.ProviderConfigUsageCleaner
 }
 
-func defaultMRManaged(m manager.Manager) mrManaged {
+func defaultMRManaged(m manager.Manager, mg resource.Managed) mrManaged {
 	return mrManaged{
-		CriticalAnnotationUpdater: NewRetryingCriticalAnnotationUpdater(m.GetClient()),
-		Finalizer:                 resource.NewAPIFinalizer(m.GetClient(), FinalizerName),
-		Initializer:               NewNameAsExternalName(m.GetClient()),
-		ReferenceResolver:         NewAPISimpleReferenceResolver(m.GetClient()),
-		ConnectionPublisher:       NewAPISecretPublisher(m.GetClient(), m.GetScheme()),
-		LocalConnectionPublisher:  NewAPILocalSecretPublisher(m.GetClient(), m.GetScheme()),
-
-		// Providers opt in to ProviderConfigUsage protection by supplying their
-		// tracker via WithProviderConfigUsageCleaner.
-		ProviderConfigUsageCleaner: resource.NewNopProviderConfigUsageCleaner(),
+		CriticalAnnotationUpdater:  NewRetryingCriticalAnnotationUpdater(m.GetClient()),
+		Finalizer:                  resource.NewAPIFinalizer(m.GetClient(), FinalizerName),
+		Initializer:                NewNameAsExternalName(m.GetClient()),
+		ReferenceResolver:          NewAPISimpleReferenceResolver(m.GetClient()),
+		ConnectionPublisher:        NewAPISecretPublisher(m.GetClient(), m.GetScheme()),
+		LocalConnectionPublisher:   NewAPILocalSecretPublisher(m.GetClient(), m.GetScheme()),
+		ProviderConfigUsageCleaner: defaultProviderConfigUsageCleaner(m, mg),
 	}
+}
+
+// defaultProviderConfigUsageCleaner returns the cleaner that protects and
+// releases the ProviderConfigUsages of the supplied managed resource. The
+// tracker family follows the managed resource's scope, and the usage kind is
+// the only ProviderConfigUsage of that scope registered with the manager's
+// scheme - the same kind the provider's connector records usages with. A
+// managed resource that is neither cluster scoped nor namespaced, or whose
+// scheme registers no or several usage kinds of its scope, gets a cleaner that
+// does nothing; WithProviderConfigUsageCleaner overrides the choice.
+func defaultProviderConfigUsageCleaner(m manager.Manager, mg resource.Managed) resource.ProviderConfigUsageCleaner {
+	switch mg.(type) {
+	case resource.LegacyManaged:
+		if u, ok := onlyUsageKind[resource.LegacyProviderConfigUsage](m.GetScheme()); ok {
+			return resource.NewLegacyProviderConfigUsageTracker(m.GetClient(), u)
+		}
+	case resource.ModernManaged:
+		if u, ok := onlyUsageKind[resource.TypedProviderConfigUsage](m.GetScheme()); ok {
+			return resource.NewProviderConfigUsageTracker(m.GetClient(), u)
+		}
+	}
+
+	return resource.NewNopProviderConfigUsageCleaner()
+}
+
+// onlyUsageKind returns the single ProviderConfigUsage kind implementing U that
+// is registered with the supplied scheme. It returns false if the scheme
+// registers no such kind, or more than one.
+func onlyUsageKind[U resource.ProviderConfigUsage](s *runtime.Scheme) (U, bool) {
+	var found []U
+
+	seen := map[reflect.Type]bool{}
+
+	for _, t := range s.AllKnownTypes() {
+		if seen[t] {
+			continue
+		}
+
+		seen[t] = true
+
+		if u, ok := reflect.New(t).Interface().(U); ok {
+			found = append(found, u)
+		}
+	}
+
+	if len(found) != 1 {
+		var zero U
+
+		return zero, false
+	}
+
+	return found[0], true
 }
 
 func (m mrManaged) PublishConnection(ctx context.Context, managed resource.Managed, c ConnectionDetails) (bool, error) {
@@ -813,14 +864,15 @@ func WithFinalizer(f resource.Finalizer) ReconcilerOption {
 }
 
 // WithProviderConfigUsageCleaner specifies how the Reconciler should protect
-// and release the managed resource's ProviderConfigUsage. Managed resources
-// that don't use a ProviderConfigUsage don't need this option; the Reconciler
-// neither protects nor releases a usage by default.
+// and release the managed resource's ProviderConfigUsage. The Reconciler adds
+// the usage's finalizer once it has connected, and removes it once the managed
+// resource no longer needs its ProviderConfig.
 //
-// Supplying a cleaner is what enables ProviderConfig deletion protection. The
-// Reconciler adds the usage's finalizer once it has connected, and removes it
-// once the managed resource no longer needs its ProviderConfig, so the
-// finalizer is only ever added where something is wired up to remove it.
+// By default the Reconciler derives a cleaner from the managed resource's scope
+// and the ProviderConfigUsage kind registered with the manager's scheme, so
+// most providers need not supply one. Supply one when the scheme registers no
+// or several usage kinds of the managed resource's scope, or supply
+// resource.NewNopProviderConfigUsageCleaner() to turn protection off.
 //
 // The cleaner need not be the tracker the provider records usages with. It's
 // only used to find the usage belonging to a managed resource, so a separately
@@ -926,7 +978,7 @@ func NewReconciler(m manager.Manager, of resource.ManagedKind, o ...ReconcilerOp
 		pollIntervalHook:            defaultPollIntervalHook,
 		creationGracePeriod:         defaultGracePeriod,
 		timeout:                     reconcileTimeout,
-		managed:                     defaultMRManaged(m),
+		managed:                     defaultMRManaged(m, nm()),
 		external:                    defaultMRExternal(),
 		supportedManagementPolicies: defaultSupportedManagementPolicies(),
 		log:                         logging.NewNopLogger(),
--- a/pkg/reconciler/managed/reconciler.go
+++ b/pkg/reconciler/managed/reconciler.go
@@ -67,6 +67,7 @@ const (
 	errReconcileUpdate          = "update failed"
 	errReconcileDelete          = "delete failed"
 	errRecordChangeLog          = "cannot record change log entry"
+	errUntrackProviderConfig    = "cannot release ProviderConfigUsage"
 
 	errExternalResourceNotExist = "external resource does not exist"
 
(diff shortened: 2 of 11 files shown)


rossigee:
Found one issue in pkg/event/event.go:103:

The Eventf call passes reason for both the reason and action parameters:
```go
r.kube.Eventf(obj, nil, string(e.Type), string(e.Reason), string(e.Reason), "%s", e.Message)
```

Per Kubernetes API docs, action should describe what the controller did (e.g., empty string, "Created", "Updated"), not why. Suggest changing the 5th parameter to an empty string or appropriate action.

pkg/meta/meta.go
```diff
@@ -205,6 +205,23 @@ func FinalizerExists(o metav1.Object, finalizer string) bool {
 	return slices.Contains(f, finalizer)
 }
 
+// FinalizersExcludingPropagation returns the supplied object's finalizers
+// without Kubernetes deletion propagation finalizers.
+func FinalizersExcludingPropagation(o metav1.Object) []string {
```
ulucinar: Naming tastes are subjective :), but how about something like the following?:
```suggestion
func NonGCFinalizers(o metav1.Object) []string {
```
My reasoning is those finalizers that we filter are used by the GC. "Propagation" already refers to that but I think "GC" is more direct.

pkg/meta/meta.go
```diff
@@ -205,6 +205,23 @@ func FinalizerExists(o metav1.Object, finalizer string) bool {
 	return slices.Contains(f, finalizer)
 }
 
+// FinalizersExcludingPropagation returns the supplied object's finalizers
+// without Kubernetes deletion propagation finalizers.
```
ulucinar: We may list these two finalizer names explicitly in the doc comment. I don't think there exists a third one but it's worth checking for correctness.

pkg/meta/meta.go
ulucinar: Naming is as usual subjective :), but how about the following?

```suggestion
func NonGCFinalizers(o metav1.Object) []string {
```
"Propagation" already implies that by referring to GC's propagation mechanisms but why not mention the Kubernetes GC directly in the name?

pkg/reconciler/managed/reconciler.go
```diff
@@ -876,12 +878,18 @@ func (r *Reconciler) effectivePollInterval(o metav1.Object) time.Duration {
...
-// capable of managing resources in a real system.
-func NewReconciler(m manager.Manager, of resource.ManagedKind, o ...ReconcilerOption) *Reconciler {
+// provider API. The usage cleaner should be the tracker used to track the
+// managed resource's ProviderConfigUsage. Managed resources that do not use a
+// ProviderConfigUsage should pass resource.NewNopProviderConfigUsageCleaner().
+// NewReconciler panics if the usage cleaner is nil or the managed resource kind
+// is not registered with the supplied manager's runtime.Scheme. The returned
+// Reconciler uses a no-op external system by default; callers should supply an
+// ExternalConnector that can manage resources in a real system.
+func NewReconciler(m manager.Manager, of resource.ManagedKind, usage resource.ProviderConfigUsageCleaner, o ...ReconcilerOption) *Reconciler {
```
ulucinar: How about making `usage` a `ReconcilerOption`? `managed.NewReconciler` then becomes a variadic function, providers not participating will not set the option (e.g., `WithProviderConfigUsageCleaner`) and then we can default to the new `NewNopProviderConfigUsageCleaner` in `defaultMRManaged` in case the provider does not specify the option.

We would then remove the `nil` check and the panic below.

pkg/reconciler/managed/reconciler.go
```diff
@@ -907,6 +915,7 @@ func NewReconciler(m manager.Manager, of resource.ManagedKind, o ...ReconcilerOp
 		change:                      newNopChangeLogger(),
 		conditions:                  new(conditions.ObservedGenerationPropagationManager),
 	}
+	r.managed.ProviderConfigUsageCleaner = usage
```
ulucinar: We may consider moving this to a `ReconcilerOption`, e.g., `WithProviderConfigUsageCleaner` (similar to, for example, `WithFinalizer`). Please see the comments above.

pkg/reconciler/managed/reconciler.go
```diff
@@ -1070,6 +1079,18 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (resu
 			return reconcile.Result{Requeue: true}, errors.Wrap(updateStatus(), errUpdateManagedStatus)
 		}
 
+		// Release the usage after removing the managed resource finalizer so the
+		// ProviderConfig remains available throughout deletion.
+		if err := r.managed.Untrack(ctx, managed); err != nil {
+			log.Debug("Cannot release ProviderConfigUsage", "error", err)
```
ulucinar, line 1156: nit: We may also consider emitting an event.

pkg/reconciler/managed/reconciler.go
```diff
@@ -1576,3 +1609,9 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (resu
 
 	return reconcile.Result{RequeueAfter: reconcileAfter}, errors.Wrap(updateStatus(), errUpdateManagedStatus)
 }
+
+// numControllerFinalizers returns the number of finalizers not managed by the
+// Kubernetes garbage collector.
+func numControllerFinalizers(mg resource.Managed) int {
```
ulucinar: nit: We may consider dropping this.

pkg/resource/providerconfig.go
```diff
@@ -195,36 +226,58 @@ func (u *ProviderConfigUsageTracker) Track(ctx context.Context, mg ModernManaged
 		Kind:       gvk.Kind,
 		Name:       mg.GetName(),
 	})
+	meta.AddFinalizer(pcu, ProviderConfigUsageFinalizer)
```
ulucinar: Looks like we are unconditionally adding the PCU finalizer here, regardless of whether the provider implementation has opted-in for implementing the new PCU protection. I think this violates what's discussed in the corresponding design. We had better make this conditional, or update the design to reflect what's implemented in this PR. We should also cover the `AllowUpdateIf` below, i.e., make that conditional also, if we are conditionally adding the finalizer here.

We also need to consider these for the `LegacyProviderConfigUsageTracker` implementation below.

The current reaper implementation seems to be helping us here (in this unconditional finalizer implementation), in a way not discussed in the design. If the provider supplies a `NopProviderConfigUsageCleaner`, which does not actually remove the PCU finalizer unconditionally added here, then the reaper effectively removes the finalizer in both background & foreground cascading deletes when the controller MR of the PCU has been removed/replaced or has already removed its MR finalizer.

However, please also see the comment on the reaper implementation for the edge case when PCU's `ref` is `nil`...

pkg/reconciler/providerconfig/reconciler.go
ulucinar: But now that we are (unconditionally) adding a PCU finalizer, if this boundary condition holds, merely deleting the PCU will not be sufficient: PCU will be stuck in deletion state. I think we need to remove the finalizer here first, if it exists, before making the delete call.

ulucinar (commented):
Thank you @ezgidemirel, left some comments for consideration.

pkg/meta/meta.go
ezgidemirel: Done

pkg/meta/meta.go
ezgidemirel: Took the suggestion :)

pkg/reconciler/managed/reconciler.go
ezgidemirel: Done. `NewReconciler` is back to (m, of, o ...ReconcilerOption), `defaultMRManaged` sets `NewNopProviderConfigUsageCleaner()`, and the nil check and panic are gone. This also removes the compile-time break for  every provider, which was the main reason I'd made it required

pkg/reconciler/managed/reconciler.go
ezgidemirel: Done, `WithProviderConfigUsageCleaner`. Its doc comment explains that the reconciler owns both ends of the finalizer and that the cleaner need not be the same tracker instance the connector uses, only one   
built with the same client and usage type

pkg/reconciler/managed/reconciler.go
ezgidemirel: Done

pkg/reconciler/managed/reconciler.go
ezgidemirel: Dropped

pkg/resource/providerconfig.go
ezgidemirel: You were right, and I went further than making it conditional: `Track` no longer touches the finalizer at all, on both trackers. The reconciler owns both ends through the cleaner: Protect runs right after a 
successful `Connect` and adds the finalizer if it's missing, reading from the cache and writing only when absent; `Untrack` removes it after the MR's own finalizer, on both the `Delete` and `Orphan` paths. A      
provider that doesn't pass the option gets the no-op cleaner and never creates a finalized usage, which is the rule the design asks for. `AllowUpdateIf` is back to the reference-change check, and `Track `     
applies with a preserveFinalizers hook so a ProviderConfig reference change doesn't strip a finalizer the reconciler put there.                                                                              

On the reaper: agreed it was doing more than the design said, and I've kept that behavior deliberately and documented it. It releases a terminating, finalized usage when the owner is gone, was replaced, or
is itself deleting with only GC finalizers left, so it backstops foreground as well as background. With the finalizer now opt-in it's a backstop for a wired provider's failed release, not a rescue for     
 unwired ones. The one-pager is updated to match.

pkg/reconciler/providerconfig/reconciler.go
ezgidemirel: Done. If the usage has no controller reference and carries our finalizer, the reconciler removes the finalizer and updates the usage before issuing the delete

pkg/resource/providerconfig.go
ulucinar: Hi @ezgidemirel,
Now that the managed reconciler is responsible for both adding the PCU finalizer and removing it once the corresponding MR no longer needs it (deleted/replaced or non-GC finalizer have been removed), would it be possible to use the actual (non-noop) `ProviderConfigUsageCleaner` implementation as the default one? This way we would have the fix automatically rolled out to the providers once they update their crossplane-runtime versions. We should be able to determine the appropriate `ProviderConfigUsageCleaner` implementation, `LegacyProviderConfigUsageTracker` or `ProviderConfigUsageTracker`, based on the GVK of the MR being reconciled as the GVK information is already available in `managed.NewReconciler`.

ezgidemirel: Hi @ulucinar, thanks for the idea! Now that the reconciler owns both ends of the finalizer it's safe. Done in 0d4f5d6.                                                                                                           
                                                                                                                                                                                                             
`defaultMRManaged` instantiates the MR kind to decide legacy vs. modern, then  
looks up the single `ProviderConfigUsage` kind of that scope registered with the manager's scheme, which is the same kind the provider's connector records usages with.                                                   

`WithProviderConfigUsageCleaner` stays as an override for a scheme that registers no or several usage kinds of a scope, and passing `NewNopProviderConfigUsageCleaner()` is how a provider turns protection off. 
                                                                                                                                                                                               
The same commit also makes `Protect` skip a usage that already has a deletion timestamp, since the API server rejects new finalizers there.

ulucinar (approved):
Thank you @ezgidemirel, lgtm.

phisco (approved):
Thanks @ezgidemirel!

jonasz-lasut:
Note for community provider maintainers:

If you pick up changes from this PR (ProviderConfig deletion protection, on `main` past v2.4.0), be aware that the managed reconciler only wires the new `ProviderConfigUsage` finalizer when your manager scheme registers exactly one `ProviderConfigUsage` Go type per scope, and silently falls back to a no-op cleaner otherwise. Providers that keep an older `ProviderConfigUsage` API version registered next to the storage version (provider-helm's cluster scope does) should pass their tracker via `managed.WithProviderConfigUsageCleaner(...)` or drop the old version from the scheme; see crossplane-contrib/provider-helm#399 for the one-line implementation.

